PR #113, opened 2025-02-17
Remove lapacke dependency.

cianciosa:
This removes the use of lapacke which is not standard across all LAPACK vendors. This enables building on macOS without home-brew.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Otherwise the Ubuntu package `python-is-python3` provides the `python` alias.
 1. Install dependencies via [Homebrew](https://brew.sh/):
 
 ```shell
-brew install python@3.10 gcc cmake ninja libomp netcdf-cxx eigen nlohmann-json protobuf lapack git open-mpi
+brew install python@3.10 gcc cmake ninja libomp netcdf-cxx eigen nlohmann-json protobuf git open-mpi
 ```
 
 2. Install VMEC++ as a Python package (possibly after creating a virtual environment):
--- a/src/vmecpp/cpp/vmecpp/free_boundary/laplace_solver/laplace_solver.cc
+++ b/src/vmecpp/cpp/vmecpp/free_boundary/laplace_solver/laplace_solver.cc
@@ -9,6 +9,24 @@
 #include "absl/algorithm/container.h"
 #include "absl/log/check.h"
 
+///  LU factorization of a general M-by-N matrix A
+extern "C" void dgetrf_(int *m,
+                        int *n,
+                        double *a,
+                        int *lda,
+                        int *ipiv,
+                        int *info);
+///  Solves a system of linear equations using the LU factorization.
+extern "C" void dgetrs_(char *transpose,
+                        int *num_rows,
+                        int *num_columns,
+                        double *matrix,
+                        int *leading_dim,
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -145,7 +145,7 @@ Otherwise the Ubuntu package `python-is-python3` provides the `python` alias.
 1. Install dependencies via [Homebrew](https://brew.sh/):
 
 ```shell
-brew install python@3.10 gcc cmake ninja libomp netcdf-cxx eigen nlohmann-json protobuf lapack git open-mpi
+brew install python@3.10 gcc cmake ninja libomp netcdf-cxx eigen nlohmann-json protobuf git open-mpi
```
jonathanschilling, line 148: (non-expert for compiling on Mac writing here) Could it ever occur that the Accelerate framework (which I assume would provide LAPACK?) is not installed on a MacOS system? If that can happen, maybe we need to add a word of warning that the homebrew `lapack` package would be needed in that case?

jonathanschilling (approved):
Thanks @cianciosa !
The PR looks fine to me - green CI, I can build and run the code on my local computer, and all the tests pass on my machine as well. Happy to merge when the comment above is resolved.
Congrats to being the first external contributor to VMEC++ :clap: :+1:

README.md
cianciosa: No it's a system library. That's also the reason why the previous cmake builds weren't finding lapacke from homebrew. FindLAPACK was most likely finding the Accelerate framework. To get the version in home-brew, you would have needed to set `CMAKE_PREFIX_PATH` to the install location for home-brew and specify the `BLA_VENDOR `.

jonathanschilling: OK, then lets leave it as-is. @eguiraud-pf OK for you as well?

src/vmecpp/cpp/vmecpp/free_boundary/laplace_solver/laplace_solver.cc
```diff
@@ -9,6 +9,24 @@
 #include "absl/algorithm/container.h"
 #include "absl/log/check.h"
 
+///  LU factorization of a general M-by-N matrix A
+extern "C" void dgetrf_(int *m,
```
eguiraud-pf, line 13: for my understanding, why do we need extern C forward declarations rather than a normal header inclusion?

jons-pf: AFAIK, LAPACK does not come with a C header file, since it is natively implemented in Fortran. Only LAPACKE does actually adds C header files, where (as @cianciosa  pointed out) no consistent de-facto standard exists though. Note that we are implicitly relying on the convention that the C signature of a Fortran method is the Fortran name with a trailing underscore (hence `dgetrf_` for `DGETRF` in Fortran) - but that actually is a de-facto standard (I think) and is apparently also what Apple uses in their Accelerate framework. Here's some more info on this that I found: https://stackoverflow.com/a/11321499

cianciosa:
I rebased to main to pick up recent changes.

jons-pf (approved):
Ok for me to merge - tested to work on multiple machines and CI is green.

PR #134, opened 2025-03-05
 Enable PyPI Support

ahnaf-tahmid-chowdhury:
## Description:
This PR introduces support for building and publishing VMECPP wheels to PyPI, along with improvements to the CMake configuration for better netCDF detection.  

#### GitHub Actions:
- Added `.github/workflows/build_wheels.yml` to automate wheel building for Linux and macOS using `cibuildwheel`.
- Added `.github/workflows/pypi_publish.yml` to publish wheels to PyPI on release, utilizing `twine`.  
- Configured dependency installation for both Linux (`dnf`) and macOS (`brew`).
- Improved MacOS build environment setup (`DYLD_LIBRARY_PATH`, `MACOSX_DEPLOYMENT_TARGET`, etc.).

#### CMake Improvements:
- Updated `CMakeLists.txt`:  
  - Increased `cmake_minimum_required` to `3.18` for better compatibility.  
  - Improved netCDF detection:
    - If `find_package(netCDF)` fails, fallback to `pkg_check_modules(netCDF REQUIRED)`.  
    - Added appropriate include directories and message logging.  
  - Fixed linking issues by explicitly using `${netCDF_LIBRARIES}` instead of `netCDF::netcdf`.

PR #135, opened 2025-03-05
Refactor PyPI Publishing Workflow

ahnaf-tahmid-chowdhury:
## Description:

This PR updates the GitHub Actions workflow for publishing vmecpp wheels, to support PR #134 

- Removed the build_wheels workflow as a single workflow can serve multiple purposes.

- Need to review environment variables. The default is set to vmecpp.

Actions: https://github.com/ahnaf-tahmid-chowdhury/vmecpp/actions


jurasic-pf (approved):
Looks good to me, its nice that the one workflow is now split in two stages with build and publish.

PR #138, opened 2025-03-06
Migrate ensure_vmec2000_input et al to _util

jons-pf:


jons-pf:
* **#138** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/138?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jons-pf:
This relieves us of the `import jax` for a stand-alone VMEC++ run, which was previously pulled in via Simsopt. The `import jax` takes a lot of time on startup, which I would like to avoid when just doing a stand-alone VMEC++ run.

jons-pf:
@jurasic-pf FYI

eguiraud-pf (commented):
^

jurasic-pf:
I agree that MPI init is the worst offender, and therefore suggest that we both
a) make MPI an optional dependency (just like in simsopt, already implemented in one of my PRs)
b) definitely prevent importing it when running VMEC standalone

But the jax import is also very slow, on some systems even worse than those 1.7s you measured... 
Can't we just have it in utils and import it into the `simsopt_compat` module, so it's still exposed?

eguiraud-pf:
big 👍 for a) and b).

I can't reproduce the problem with jax, `python -X importtime -c 'import vmecpp.simsopt_compat'` does not show jax as an offender for me, see picture below produced with https://github.com/kmichel/python-importtime-graph?tab=readme-ov-file .

jons-pf:
I would like VMEC++ to be usable without having to rely on `simsopt` in any way.

For now, with these changes, we still need to install the `simsopt` dependency, but we don't need to import it - which makes for a snappier user experience on startup. A possible next step to this would be to put the whole `simsopt_compat` layer into a separate Python module, so that one can get a clean VMEC++ without ever needing to think about `simsopt`.
Regarding the `_util` methods: If we want to expose them, what about making `_util` publicly usable then, or putting the Fortran-VMEC compatibility methods into a separate `vmec2000_compat` module?

eguiraud-pf:
> I would like VMEC++ to be usable without having to rely on simsopt in any way.

ok, but that feels like a separate discussion from slow import times. slow import times should be fixed by @jurasic-pf 's PR that makes the mpi initialization lazy.

> Regarding the _util methods: If we want to expose them

No, those are private. it's the simsopt_compat utils that are public, but you are moving them into a private module.

@jons-pf I'm getting a bit lost. Can you please check whether @jurasic-pf 's PR fixes the import time issues, which would make this PR obsolete? to be clear, the problem with this PR is that it makes currently public functionality, that was public on purpose, private.

jurasic-pf:
> big 👍 for a) and b).
> 
> I can't reproduce the problem with jax, `python -X importtime -c 'import vmecpp.simsopt_compat'` does not show jax as an offender for me, also see picture below produced with https://github.com/kmichel/python-importtime-graph?tab=readme-ov-file (EDIT: better viewed by opening it in its own tab).
> 
> 

## Plot without MPI:
importtime

@eguiraud-pf  Wow, what a useful tool for import times! 

Optional MPI does improve the situation a bit, but when MPI **is installed** its of course still bad - regardless of whether we use it or not. `simsopt` is the second worst offender, and still takes 1s just to dump the help message, on our laptops - which are quite fast imo.

About jax: On my work laptop it also isn't an issue, on my desktop at home and apparently Jons laptop jax imports take a lot longer.

eguiraud-pf:
> when MPI is installed its of course still bad - regardless of whether we use it or not

I think the expensive part is the creation of the `MpiPartition`, right? then we need to make that lazy but still cached at the level of the module, there are a bunch of ways to do that.

on top of that we can also make it so that `import vmecpp` does not transitively cause `import simsopt`, as @jons-pf suggests -- we just have to pick an appropriate place for `ensure_vmecpp_input` (if I'm not mistaken that's the only reason `src/vmecpp/__init__.py` imports `simsopt_compat`. in first approximation we could even just move `ensure_vmecpp_input` into `src/vmecpp/__init__.py` and call it a day (there are probably better ways).

jurasic-pf:
Ok, I went ahead and moved the public methods `ensure_vmec` and `is_vmec_input`
 to `__init__.py`, and all private methods associated with them into `_util`
 
Importtimes are now ~350ms, still noticable latency when running CLI, but not a level I would complain about anymore. 
importtime(3)

src/vmecpp/_util.py
```diff
@@ -146,3 +149,135 @@ def indata_to_json(
...
+    indata += fourier_coefficients_to_namelist("zbs", vmecpp_json)
+    indata += fourier_coefficients_to_namelist("rbs", vmecpp_json)
+    indata += fourier_coefficients_to_namelist("zbc", vmecpp_json)
+
+    indata += "\n/\n&END\n"
+
+    return indata
+
+
+def bool_to_namelist(varname: str, vmecpp_json: dict[str, Any]) -> str:
```
eguiraud-pf, line 232: this and the other helper functions can still all be private right? to reduce the surface of the public API exposed by `_util`

eguiraud-pf (approved):
LGTM, thanks @jurasic-pf !

src/vmecpp/_util.py
jurasic-pf: Oh, you mean, private from a "development perspective"!
I changed the _ prefix because anything in _util is private by definition, so they aren't part of public API anyway, but I guess there's still value in differentiating which are just helper functions internal to a file and which are "vmecpp internal", aka not user facing.

jurasic-pf:
### Merge activity

* **Mar 17, 12:48 PM EDT**: [Graphite](https://app.graphite.dev/github/pr/proximafusion/vmecpp/138) rebased this pull request as part of a merge.

PR #145, opened 2025-03-06
Fix ruff formatting complaints

jurasic-pf:


jurasic-pf:
* **#145** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/145?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/145?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jurasic-pf:
### Merge activity

* **Mar 6, 8:04 PM EST**: [Graphite](https://app.graphite.dev/github/pr/proximafusion/vmecpp/145) couldn't merge this PR because it was not satisfying all requirements (PR does not have required approvals).

eguiraud-pf (commented):
uhm, not sure about this one. what platform that we want to support does not have at least python 3.10? Ubuntu 22.04 has python 3.10.

I don't think we should force ourselves out of nice Python features just so we can support 3.9 for the sake of it.

jurasic-pf:
Fair point, I don't know any distro where 3.9 would be the crucial difference. Then 3.10 is a reasonable limit

src/vmecpp/simsopt_compat.py
```diff
@@ -62,29 +62,29 @@ class Vmec(Optimizable):
...
+    output_file: str | None
     # These are filled:
     # - by __init__ if Vmec is initialized with an output file
     # - by a call to run() and are None before otherwise
-    s_full_grid = Optional[Float[np.ndarray, " ns"]]
-    ds = Optional[Float]
-    s_half_grid = Optional[Float[np.ndarray, " nshalf"]]
+    s_full_grid: Float[np.ndarray, " ns"] | None
+    ds: Float | None
+    s_half_grid: Float[np.ndarray, " nshalf"] | None
```
jurasic-pf, line 76: These must have been a mistake, right? They were probably supposed to be type hints

jurasic-pf:
And since we've settled on 3.10, all type hints should follow the new syntax then

src/vmecpp/simsopt_compat.py
eguiraud-pf: `Float[np.ndarray, " ns"]` and similar is the correct type annotation using jaxtyping.

`ds: Float` is wrong, should be `ds: float`

src/vmecpp/simsopt_compat.py
```diff
@@ -62,29 +62,29 @@ class Vmec(Optimizable):
...
-    wout: Optional[Union[Struct, FortranWOutAdapter]]
+    wout: Struct | FortranWOutAdapter | None
     # Whether `run()` is available for this object:
     # depends on whether it has been initialized with an input configuration
     # or an output file.
     runnable: bool
     # False when the currently cached results are valid, True if we need to `run()`
     need_to_run_code: bool
-    mpi: Optional[MpiPartition]  # pyright: ignore
+    mpi: Optional[MpiPartition]  # pyright: ignore # noqa: UP007
```
eguiraud-pf, line 87: here and below, why not `MpiPartition | None`?

tests/test_simsopt_compat.py
```diff
@@ -205,7 +205,7 @@ def test_ensure_vmec2000_input_from_vmecpp_input():
             varname_vmec2000 = f"{varname[:-1]}c{varname[-1]}"
         vmec2000_var = getattr(vmec2000.indata, varname_vmec2000)
 
-        if isinstance(vmecpp_var, (str, int, bool)):
+        if isinstance(vmecpp_var, str | int | bool):
```
eguiraud-pf, line 208: the previous version was correct, we don't want to check whether `vmecpp_var` is a union, we want to check whether `vmecpp_var` is one of those types, which is done by passing a tuple of possible types.

src/vmecpp/simsopt_compat.py
```diff
@@ -62,29 +62,29 @@ class Vmec(Optimizable):
...
-    output_file: Optional[str]
+    output_file: str | None
     # These are filled:
     # - by __init__ if Vmec is initialized with an output file
     # - by a call to run() and are None before otherwise
-    s_full_grid = Optional[Float[np.ndarray, " ns"]]
-    ds = Optional[Float]
-    s_half_grid = Optional[Float[np.ndarray, " nshalf"]]
+    s_full_grid: Float[np.ndarray, " ns"] | None
+    ds: Float | None
```
eguiraud-pf, line 74:
```suggestion
    ds: float | None
```

eguiraud-pf (approved):
LGTM except for the comments below.

You can use `gt reorder` to move this PR to the bottom of the stack so the CI does not fail on ruff errors on the previous PRs.

tests/test_simsopt_compat.py
jurasic-pf: The | operator version does the same and is the >=3.10 recommended way of doing it (according to ruff)

src/vmecpp/simsopt_compat.py
jurasic-pf: I especially meant that it was an assignment = instead of an annotation : before.
Good point about `float`, ill change that

src/vmecpp/simsopt_compat.py
jurasic-pf: Because I learned the hard way, that functions can't be type annotated that way, and MpiPartition is overloaded with a call operator by @SimsoptRequires

Minimal example of what I mean:
```def a():
     pass
b: a|None = a
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for |: 'function' and 'NoneType'
```

jurasic-pf: I'll add a comment to clarify it was intentional

tests/test_simsopt_compat.py
eguiraud-pf: that's weird, "is this a union of types A, B and C?" is semantically different from "is this one of types A, B and C?"...

eguiraud-pf: or maybe not different after all when you are checking against a _value_, not a _type_

src/vmecpp/simsopt_compat.py
eguiraud-pf: ah, this is simsopt doing things wrong, `function`s should not be used as type annotations. ok I guess, we are just mimicking simsopt...

PR #162, opened 2025-03-13
MPI finite difference

jurasic-pf:


jons-pf (approved):
... subject to merging the changes suggested in the branch above this with this one :)

PR #181, opened 2025-03-15
without OpenMP max threads should always be one

jurasic-pf:
- Prevent segfault due to out of bounds array access when passing 0 or negative thread count
- Prevent segfault when passing >1 threads when compiled w/o openmp
- Avoid the transitive include of Abseil Check https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
+++ b/src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
@@ -3,6 +3,9 @@
 // SPDX-License-Identifier: MIT
 #include "vmecpp/common/flow_control/flow_control.h"
 
+#include "absl/log/log.h"
+#include "absl/log/check.h"
+
 #ifdef _OPENMP
 #include <omp.h>
 #endif  // _OPENMP
@@ -26,12 +29,11 @@ RestartReason RestartReasonFromInt(int restart_reason) {
 
 int get_max_threads(std::optional<int> max_threads) {
   if (max_threads == std::nullopt) {
-#ifdef _OPENMP
     return omp_get_max_threads();
-#else
-    return 1;
-#endif
   }
+  CHECK_GT(max_threads.value(), 0)
+    << "The number of threads must be >=1. "
+    "To automatically use all available threads, pass std::nullopt";
   return max_threads.value();
 }
 
--- a/src/vmecpp/simsopt_compat.py
+++ b/src/vmecpp/simsopt_compat.py
@@ -211,13 +211,18 @@ class Vmec(Optimizable):
     def recompute_bell(self, parent=None) -> None:  # noqa: ARG002
         self.need_to_run_code = True
 
-    def run(self, initial_state=None, max_threads=None) -> None:
+    def run(self, initial_state=None, max_threads=1) -> None:
         """Run VMEC if ``need_to_run_code`` is ``True``.
 
         The max_threads argument is not present in SIMSOPT's original implementation as
         it is specific to VMEC++, which will spawn the corresponding number of OpenMP
-        threads to parallelize execution. If max_threads is None (the default), VMEC++
-        runs on a single thread.
+        threads to parallelize execution. By default max_threads=1, so VMEC++ runs on a
+        single thread. To automatically enable all threads, pass max_threads=None
+
+        Most optimization frameworks use multi-process parallelization for finite
+        differencing, and we do not want to end up overcommitting the machine with NCPU
+        processes running NCPU threads each -- especially when OpenMP is involved, as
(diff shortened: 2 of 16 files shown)


src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
```diff
@@ -25,14 +28,16 @@ RestartReason RestartReasonFromInt(int restart_reason) {
 }
 
 int get_max_threads(std::optional<int> max_threads) {
+  #ifndef _OPENMP
+  return 1;
```
eguiraud-pf, line 32: if the user asked for `max_threads > 1`, it's surprising to silently set it to 1 instead if openmp support is missing: this should at least print a warning.

but actually, running VMEC++ without openmp is de-facto unsupported:
- I think we have zero test coverage for builds without openmp
- libopenmp-dev and equivalent are explicitly listed as VMEC++ requirements
- the build system does not have a flag to turn off openmp, it just so happens that gcc warns and keeps going if `-fopenmp` is passed but libopenmb-dev is missing

We should either add test coverage for a no-openmp configuration or declare openmp a hard requirement, e.g. checking for it at cmake configuration time (my not-so-strong preference would be for the latter because it would result in simplification of the code). @jons-pf is there a real-world argument for supporting installations where `_OPENMP` is not defined?

src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
```diff
@@ -25,14 +28,16 @@ RestartReason RestartReasonFromInt(int restart_reason) {
...
+  #else
   if (max_threads == std::nullopt) {
-#ifdef _OPENMP
     return omp_get_max_threads();
-#else
-    return 1;
-#endif
   }
+  CHECK_GT(max_threads.value(), 0)
+    << "The number of threads must be >=1. To automatically use all available threads, pass std::nullopt";
```
eguiraud-pf, line 38: I think this will interact pretty badly with Python:
- if I'm not mistaken, a failed `CHECK_GT` causes an abort with a stacktrace that looks quite scary from Python -- we prefer throwing exceptions that pybind11 transforms into python exceptions automatically
- Python users won't be able to pass `std::nullopt` anywhere

The simplest solution is probably to add a similar check at the level of the Python interfaces (`simsopt_compat.Vmec` and `vmecpp.run`).

src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
eguiraud-pf, line 1: please add at least a Python regression test

eguiraud-pf (changes requested):

src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
jurasic-pf: According to wikipedia OpenMP 3.1 (the featureset we use) has been supported by any relevant compiler since 2015... [Clang 3.7](https://en.wikipedia.org/wiki/OpenMP#cite_note-Clang_3.7_Release_Notes-31), [GCC 4.7](https://en.wikipedia.org/wiki/OpenMP#cite_note-openmp_%E2%80%93_GCC_Wiki-27)
I personally don't see any reason, if people don't want multithreading, just set `OMP_NUM_THREADS=1`

src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
jurasic-pf: I kept the check in CPP, so it still happens when using vmecpp as a bazel module. It is also checked in Python now

src/vmecpp/simsopt_compat.py
```diff
@@ -211,13 +211,18 @@ def __init__(
     def recompute_bell(self, parent=None) -> None:  # noqa: ARG002
         self.need_to_run_code = True
 
-    def run(self, initial_state=None, max_threads=None) -> None:
+    def run(self, initial_state=None, max_threads=1) -> None:
```
jurasic-pf, line 214: This default was inconsistent with vmecpp.run, I believe the behavior we want should be explicit.

eguiraud-pf: good for me!

eguiraud-pf (approved):

jurasic-pf:
### Merge activity

* **Mar 26, 7:14 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/proximafusion/vmecpp/181).

PR #197, opened 2025-03-27
Save method for VmecInput

jurasic-pf:
- Save VmecInput json
- Add test coverage

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -97,6 +97,22 @@ def test_vmecwout_load_from_fortran():
     assert loaded_wout is not None
 
 
+def test_vmecinput_io():
+    vmec_input = vmecpp.VmecInput.from_file(TEST_DATA_DIR / "cma.json")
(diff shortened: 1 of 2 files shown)


tests/test_init.py
```diff
@@ -97,6 +97,22 @@ def test_vmecwout_load_from_fortran():
     assert loaded_wout is not None
 
 
+def test_vmecinput_io():
+    vmec_input = vmecpp.VmecInput.from_file(TEST_DATA_DIR / "cma.json")
+    with tempfile.TemporaryDirectory() as tmp_dir:
```
eguiraud-pf, line 102: optional: as a nice simplification, I recently learned from @bonauer-pf about the built-in `tmp_path` fixture: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture -- you can use that and save one level of indentation

eguiraud-pf (approved):

PR #212, opened 2025-04-03
Fully support and test  editable installs

jurasic-pf:
### TL;DR

Fix package resource discovery to support editable installs and add editable install testing to CI.

### What changed?

- Modified the GitHub Actions workflow to test both regular and editable installs by adding a conditional pip install command
- Improved `indata_to_json()` to find the executable using `importlib.metadata.files()` rather than assuming a fixed location relative to the source files, because repo_root broke on editable.

> [!NOTE]
> Required checks are not finishing because the CI test names include editable=true/false due to this change, so they don't match the required check names. Will adjust the required checks once approved.

### DIFF
--- a/src/vmecpp/_util.py
+++ b/src/vmecpp/_util.py
@@ -1,6 +1,7 @@
 """The usual hodgepodge of helper utilities without a better home."""
 
 import contextlib
+import importlib.metadata
 import logging
 import os
 import shutil
@@ -20,6 +21,7 @@ def package_root() -> Path:
     Useful e.g. to point tests to test files using paths relative to the repo root
     rather than paths relative to the test module.
     """
+
     return Path(__file__).parent
 
 
@@ -86,9 +88,16 @@ def indata_to_json(
         msg = f"{filename} does not exist."
         raise FileNotFoundError(msg)
 
-    indata_to_json_exe = Path(
-        package_root(), "cpp", "third_party", "indata2json", "indata2json"
+    # Get the path to the packaged indata2json executable.
+    # For virtual environments, package_root() points to the source files,
+    # but this method will return the correct install location e.g. in your virtual environment.
+    indata_package_path = next(
+        filter(
+            lambda path: path.stem == "indata2json", importlib.metadata.files("vmecpp")
+        )
     )
+    indata_to_json_exe = indata_package_path.locate()
+
     if not indata_to_json_exe.is_file():
         msg = f"{indata_to_json_exe} is not a file."
         raise FileNotFoundError(msg)
--- a/src/vmecpp/_util.py
+++ b/src/vmecpp/_util.py
@@ -21,8 +21,19 @@ def package_root() -> Path:
     Useful e.g. to point tests to test files using paths relative to the repo root
     rather than paths relative to the test module.
     """
-
     return Path(__file__).parent
+
+
+def distribution_root() -> Path:
+    """The path to the install location of this package. This may be the same as
+    package_root, but doesn't have to be.
+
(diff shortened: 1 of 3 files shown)


jurasic-pf:
* **#212** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/212?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/212?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/vmecpp/_util.py
```diff
@@ -86,9 +88,16 @@ def indata_to_json(
         msg = f"{filename} does not exist."
         raise FileNotFoundError(msg)
 
-    indata_to_json_exe = Path(
-        package_root(), "cpp", "third_party", "indata2json", "indata2json"
+    # Get the path to the packaged indata2json executable.
+    # For virtual environments, package_root() points to the source files,
+    # but this method will return the correct install location e.g. in your virtual environment.
```
eguiraud-pf, line 93: I don't understand the comment, it seems to imply that package_root does not work with virtual envs, but it does..? Should this say:

```suggestion
    # For editable installs, package_root() points to the source files,
    # while importlib.metadata.files will return the correct install location in all cases.
```

src/vmecpp/_util.py
```diff
@@ -86,9 +88,16 @@ def indata_to_json(
...
+    indata_package_path = next(
+        filter(
+            lambda path: path.stem == "indata2json", importlib.metadata.files("vmecpp")
```
eguiraud-pf, line 96: optional: should we rather implement `package_root` in terms of `importlib`?

eguiraud-pf (approved):

src/vmecpp/_util.py
jurasic-pf: Yes, this is something I wanted to briefly discuss with you. 
To me it seems like this will generally give us the correct path, but I'm not sure how to best handle it:

__file__ is the wrong path:
```py
In [92]: vmecpp.__file__
Out[92]: '/home/jurasic/vmecpp/src/vmecpp/__init__.py'
```

resources also point to the wrong path:
```py
In [100]: importlib.resources.path("vmecpp", "_util.py")
Out[100]: PosixPath('/home/jurasic/vmecpp/src/vmecpp/_util.py')
```

Finally this contains the correct paths, but we would have to navigate to package root, relative to some file, I haven't found a way to get the root directly...
Also depending on how you install (from wheel, source, editable) not all files will be present (e.g. wheel instead of the source files)  
```py
importlib.metadata.files("vmecpp")
```

jurasic-pf: Yes, this is something I wanted to briefly discuss with you. 
To me it seems like this will generally give us the correct path, but I'm not sure how to best handle it:

__file__ is the wrong path:
```py
In [92]: vmecpp.__file__
Out[92]: '/home/jurasic/vmecpp/src/vmecpp/__init__.py'
```

resources also point to the wrong path:
```py
In [100]: importlib.resources.path("vmecpp", "_util.py")
Out[100]: PosixPath('/home/jurasic/vmecpp/src/vmecpp/_util.py')
```

Finally this contains the correct paths, but 
```py
importlib.metadata.files("vmecpp")
```

jurasic-pf:
### Merge activity

* **Apr 3, 1:03 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/proximafusion/vmecpp/212).

PR #198, opened 2025-03-29
Numpydantic to allow model_dump_json()

jurasic-pf:

PR #160, opened 2025-03-12
Remove duplicate deserialization logic in simsopt_compat.Vmec

jurasic-pf:
There were subtle differences in deserialization logic, due to the diverging implementation. This reconciles the two.

PR #216, opened 2025-04-04
Make the stellarator asymetric fields optional in the datamodel

jurasic-pf:
Made lasym specific terms std::optional and Optional in Python
### As discussed in 
https://github.com/proximafusion/vmecpp/pull/198#discussion_r2024553225

> it's a bit weird to say `if not self.lasym`, we don't care about "rbs" and "zbc".
> should we rather say `if self.lasym: # check shapes`, else `# check rbs and zbc are None`?

> At the moment the fields are populated by an empty list when stellarator symmetric! 
> I agree `| None` would be a clearer API, but constitutes a potentially breaking change.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -181,10 +181,10 @@ class VmecInput(pydantic.BaseModel):
     zaxis_s: npyd.NDArray[npyd.Shape["* ntor_plus_1"], float]
     """Magnetic axis coefficients for Z ~ sin(n*v); stellarator-symmetric."""
 
-    raxis_s: npyd.NDArray[npyd.Shape["* ntor_plus_1"], float]
+    raxis_s: npyd.NDArray[npyd.Shape["* ntor_plus_1"], float] | None = None
     """Magnetic axis coefficients for R ~ sin(n*v); non-stellarator-symmetric."""
 
-    zaxis_c: npyd.NDArray[npyd.Shape["* ntor_plus_1"], float]
+    zaxis_c: npyd.NDArray[npyd.Shape["* ntor_plus_1"], float] | None = None
     """Magnetic axis coefficients for Z ~ cos(n*v); non-stellarator-symmetric."""
 
     rbc: SerializableSparseCoefficientArray  # [mpol, 2 * ntor + 1]
@@ -193,10 +193,10 @@ class VmecInput(pydantic.BaseModel):
     zbs: SerializableSparseCoefficientArray  # [mpol, 2 * ntor + 1]
     """Boundary coefficients for Z ~ sin(m*u - n*v); stellarator-symmetric"""
 
-    rbs: SerializableSparseCoefficientArray  # [mpol, 2 * ntor + 1]
+    rbs: SerializableSparseCoefficientArray | None = None  # [mpol, 2 * ntor + 1]
     """Boundary coefficients for R ~ sin(m*u - n*v); non-stellarator-symmetric"""
 
-    zbc: SerializableSparseCoefficientArray  # [mpol, 2 * ntor + 1]
+    zbc: SerializableSparseCoefficientArray | None = None  # [mpol, 2 * ntor + 1]
     """Boundary coefficients for Z ~ cos(m*u - n*v); non-stellarator-symmetric"""
 
     @pydantic.model_validator(mode="after")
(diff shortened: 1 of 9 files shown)


jurasic-pf:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/216?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#216** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/216?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#198** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/198?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#160](https://github.com/proximafusion/vmecpp/pull/160) <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/160?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#212** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/vmecpp/__init__.py
```diff
@@ -211,13 +211,28 @@ def _validate_fourier_coefficients_shapes(self) -> typing.Self:
         for field in mpol_two_ntor_plus_one_fields:
             shape = np.shape(getattr(self, field))
             if shape != expected_shape:
-                msg = (
-                    f"{field} has shape {shape} instead of the expected {expected_shape}. "
-                    f"Does your coefficient array exceed mpol={self.mpol} or ntor={self.ntor}?"
-                )
+                msg = f"{field} has shape {shape} instead of the expected {expected_shape}\n"
+                f"Does your coefficient array exceed mpol={self.mpol} or ntor={self.ntor}?"
                 raise ValueError(msg)
```
eguiraud-pf, line 216: now that I think about it, isn't it common practice in Fortran VMEC to have many coefficients and then truncate as needed by reducing mpol and ntor? should this maybe be a warning instead of a hard error?

src/vmecpp/__init__.py
```diff
@@ -211,13 +211,28 @@ def _validate_fourier_coefficients_shapes(self) -> typing.Self:
...
         return self
 
+    @pydantic.model_validator(mode="after")
+    def _validate_stellarator_stellarator_asymmetric_fields(self) -> typing.Self:
```
eguiraud-pf, line 220:
```suggestion
    def _validate_stellarator_asymmetric_fields(self) -> typing.Self:
```

src/vmecpp/__init__.py
```diff
@@ -211,13 +211,28 @@ def _validate_fourier_coefficients_shapes(self) -> typing.Self:
...
+        """Check if all fields that break stellarator symmetry match the lasym flag."""
+        ASYMMETRIC_FIELDS = ["rbs", "zbc", "zaxis_c", "raxis_s"]
+        # If stellarator-symmetric
+        if not self.lasym:
```
eguiraud-pf, line 224: optional: why use a comment when we can use code 😬 
```suggestion
        is_stellarator_symmetric = not self.lasym
        if is_stellarator_symmetric:
```

src/vmecpp/__init__.py
```diff
@@ -211,13 +211,28 @@ def _validate_fourier_coefficients_shapes(self) -> typing.Self:
...
+            for key in ASYMMETRIC_FIELDS:
+                value = getattr(self, key)
+                # Then all asymmetric fields should be None
+                if value is not None:
+                    msg = (
+                        "The input is for a stellarator symmetric configuration (lasym=False), "
+                        f"but the symmetry breaking field '{key}' is populated with \n{value}"
```
eguiraud-pf, line 231:
```suggestion
                        f"but the symmetry-breaking field '{key}' is populated with \n{value}"
```

src/vmecpp/__init__.py
```diff
@@ -282,7 +296,12 @@ def _to_cpp_vmecindatapywrapper(self) -> _vmecpp.VmecINDATAPyWrapper:
         cpp_indata._set_mpol_ntor(self.mpol, self.ntor)
         for attr in readonly_attrs - {"mpol", "ntor"}:
             # now we can set the elements of the readonly_attrs
-            getattr(cpp_indata, attr)[:] = getattr(self, attr)
+            value = getattr(self, attr)
+
+            # Asymmetric fields are only populated when lasym==True
+            # so we need to skip them for itemwise assignment
+            if value is not None:
+                getattr(cpp_indata, attr)[:] = value
```
eguiraud-pf, line 304: this can end up silently doing weird things in case of bugs elsewhere in the code. let's assert that the preconditions we expect are satisfied:

```suggestion
            if value is None:  # must be an asymmetric field and lasym == True
                assert attr in {"rbs", "zbc", "zaxis_c", "raxis_s"} and not cpp_indata.lasym
            else:
                getattr(cpp_indata, attr)[:] = value
```

src/vmecpp/cpp/vmecpp/vmec/pybind11/vmec_indata_pywrapper_test.cc
```diff
@@ -61,17 +61,35 @@ void CheckEquality(const vmecpp::VmecINDATAPyWrapper &wrapper,
   EXPECT_EQ(wrapper.lforbal, indata.lforbal);
   EXPECT_THAT(wrapper.raxis_c, ElementsAreArray(indata.raxis_c));
   EXPECT_THAT(wrapper.zaxis_s, ElementsAreArray(indata.zaxis_s));
-  EXPECT_THAT(wrapper.raxis_s, ElementsAreArray(indata.raxis_s));
-  EXPECT_THAT(wrapper.zaxis_c, ElementsAreArray(indata.zaxis_c));
-
+  if(indata.lasym){
+    EXPECT_THAT(wrapper.raxis_s.value(), ElementsAreArray(indata.raxis_s));
+    EXPECT_THAT(wrapper.zaxis_c.value(), ElementsAreArray(indata.zaxis_c));
+  }else{
```
eguiraud-pf, line 67: not for this PR, but looking at the formatting here, maybe we should add a clang-format check to the CI and in pre-commit 😬

eguiraud-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/pybind11/vmec_indata_pywrapper_test.cc
jurasic-pf: Yeah, I briefly added it and noticed it made large changes in every file I touched - but going forward we should just format once and add it to CI.

src/vmecpp/__init__.py
jurasic-pf, line 304:
```suggestion
            if value is None:  # must be a symmetric field and lasym == False
                assert attr in {"rbs", "zbc", "zaxis_c", "raxis_s"} and not cpp_indata.lasym
            else:
                getattr(cpp_indata, attr)[:] = value
```

src/vmecpp/__init__.py
jurasic-pf: It is a workflow that exists, but I would argue that we don't really want to support it (also dicussed this with @jons-pf today). This isn't the first precedent of making the "contracts" stricter in VMECPP, e.g. we don't allow populated rbs, zbc fields when lasym==False.

To be clear: This doesn't break backwards compatibility with namelists which use that convention! Input files which have those fields populated will be accepted and handled, but by the time they reach the Python layer, all the fields should have consistent dimensions, symmetry properties etc.

There are helper methods for resizing coefficient arrays if needed and the error messages guide the users to them now.

PR #244, opened 2025-04-14
Vmec constructor -> Factory for error handling during initialization.

jurasic-pf:

PR #245, opened 2025-04-14
mgrid file loading with absl::Status

jurasic-pf:
## Issue
The Vmec constructor had absl::CHECK calls in the constructor, which terminated not just the vmecpp execution but also the python wrapper process when mgrid file loading failed, without an Exception or error message that the user could handle. 

### TLDR 
Init method for mgrid loading

### Long version
The PR refactors the `Vmec` constructor to do pure, exception free initialization (Ignoring the possibility of a memory allocation error from std::vector) and moves all sources of exceptions to places where we can report an `absl::Status`. It removes the magnetic response table constructor of Vmec, and all mgrid related initialization is instead handled in `InitMgrid`, a new method that returns the completion status.

**The downside** is, that `Vmec(indata)` does no longer produce a valid object for free-boundary input files. Instead, the mgrid load is finalized during the first call to `run()` if any file IO is required.

**Positive** is that the error handling is now correct for missing, invalid mgrid files, wrong resolutions etc. and returns error codes/Python exceptions instead of terminating.

A long term better API would be: https://github.com/proximafusion/vmecpp/pull/244

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
@@ -105,7 +105,7 @@ absl::Status CheckInitialState(const vmecpp::HotRestartState& initial_state,
 absl::StatusOr<vmecpp::OutputQuantities> vmecpp::run(
     const VmecINDATA& indata, std::optional<HotRestartState> initial_state,
     std::optional<int> max_threads, bool verbose) {
-  Vmec v(indata, nullptr, max_threads, verbose);
+  Vmec v = Vmec(indata, max_threads, verbose);
 
   // the values of the first three arguments should just be VMEC's defaults
   absl::StatusOr<bool> s =
@@ -123,7 +123,11 @@ absl::StatusOr<vmecpp::OutputQuantities> vmecpp::run(
     const makegrid::MagneticFieldResponseTable& magnetic_response_table,
     std::optional<HotRestartState> initial_state,
     std::optional<int> max_threads, bool verbose) {
-  Vmec v(indata, &magnetic_response_table, max_threads, verbose);
+  Vmec v(indata, max_threads, verbose);
+  absl::Status mgrid_status = v.LoadMGrid(magnetic_response_table);
+  if (!mgrid_status.ok()) {
+    return mgrid_status;
+  }
 
   // the values of the first three arguments should just be VMEC's defaults
   absl::StatusOr<bool> s =
@@ -140,12 +144,6 @@ namespace vmecpp {
 
 Vmec::Vmec(const VmecINDATA& indata, std::optional<int> max_threads,
            bool verbose)
-    : Vmec(indata, nullptr, max_threads, verbose) {}
-
-// initialize based on input file contents
-Vmec::Vmec(const VmecINDATA& indata,
-           const makegrid::MagneticFieldResponseTable* magnetic_response_table,
-           std::optional<int> max_threads, bool verbose)
     : indata_(indata),
       s_(indata_),
       t_(&s_),
@@ -166,16 +164,6 @@ Vmec::Vmec(const VmecINDATA& indata,
   fc_.haveToFlipTheta = b_.setupFromIndata(indata_, verbose_);
 
   if (fc_.lfreeb) {
-    if (magnetic_response_table == nullptr) {
-      absl::Status s = mgrid_.LoadFile(indata_.mgrid_file, indata_.extcur);
-      CHECK_OK(s) << "Could not load mgrid file '" << indata_.mgrid_file << "'";
-    } else {
-      absl::Status s =
-          mgrid_.LoadFields(magnetic_response_table->parameters,
-                            *magnetic_response_table, indata_.extcur);
-      CHECK_OK(s);
-    }
-
     // tangential Fourier resolution
     // 0 : ntor
     int nf = s_.ntor;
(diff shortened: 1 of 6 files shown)


jurasic-pf:
* **#245** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/245?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/245?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/proximafusion?ref=merge-queue-instructions-in-stack-comment&prId=245).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
```diff
@@ -164,35 +162,40 @@ Vmec::Vmec(const VmecINDATA& indata,
...
-    if (magnetic_response_table == nullptr) {
-      absl::Status s = mgrid_.LoadFile(indata_.mgrid_file, indata_.extcur);
-      CHECK_OK(s) << "Could not load mgrid file '" << indata_.mgrid_file << "'";
-    } else {
-      absl::Status s =
-          mgrid_.LoadFields(magnetic_response_table->parameters,
-                            *magnetic_response_table, indata_.extcur);
-      CHECK_OK(s);
-    }
+absl::Status Vmec::InitMgrid(
```
jons-pf, line 167: nit: what about `LoadMGrid` to be consistent with the `mgrid_` member methods?

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
```diff
@@ -164,35 +162,40 @@ Vmec::Vmec(const VmecINDATA& indata,
...
-    bvecShare.resize(mnpd, 0.0);
-
-    // not so nice to do this here, but meh...
-    h_.vacuum_magnetic_pressure.resize(s_.nZnT, 0.0);
-
-    h_.vacuum_b_r.resize(s_.nZnT);
-    h_.vacuum_b_phi.resize(s_.nZnT);
-    h_.vacuum_b_z.resize(s_.nZnT);
+  absl::Status s{};
+  if (magnetic_response_table == nullptr) {
```
jons-pf, line 175: Hmm, I am not too sure if it is obvious to users that passing a `nullptr` is expected-working default behavior... What about passing a `std::optional<const makegrid::MagneticFieldResponseTable&> = std::nullopt` ?

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.h
```diff
@@ -73,13 +73,17 @@ class Vmec {
...
                 std::optional<int> max_threads = std::nullopt,
                 bool verbose = true);
 
-  Vmec(const VmecINDATA& indata,
-       const makegrid::MagneticFieldResponseTable* magnetic_response_table,
-       std::optional<int> max_threads = std::nullopt, bool verbose = true);
+  // Mgrid loading for free-boundary VMEC++ done outside of the Vmec constructor
+  // for improved exception handling
+  absl::Status InitMgrid(const makegrid::MagneticFieldResponseTable*
+                             magnetic_response_table = nullptr);
```
jons-pf, line 86: see above - `std::nullopt` if we make this `std::optional`

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
jurasic-pf: I believe the reasoning for the pointer type is due to the size of the response table - we don't want to unnecessarily copy `O(10MB)` of data around for nothing (although we already copy it).

I instead propose we simply forward the two `_mgrid.Load` methods, with two method signatures.

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.h
jurasic-pf: See above

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
jons-pf: Sounds good, let's do that :)

jons-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.h
```diff
@@ -73,13 +73,20 @@ class Vmec {
...
   explicit Vmec(const VmecINDATA& indata,
                 std::optional<int> max_threads = std::nullopt,
                 bool verbose = true);
 
-  Vmec(const VmecINDATA& indata,
-       const makegrid::MagneticFieldResponseTable* magnetic_response_table,
-       std::optional<int> max_threads = std::nullopt, bool verbose = true);
+  // Mgrid loading for free-boundary VMEC++ from a precomputed response-table is
+  // done outside of the Vmec constructor for improved exception handling
+  absl::Status LoadMGrid(
```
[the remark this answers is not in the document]
jurasic-pf: It would kind of behave like an optional, in the sense that it would pick the correct method depending on the arguments, but it would be different in the sense that you couldn't pass `None`. 
Either way, we are not exposing this to Python, only vmecpp::run is a valid entrypoint there, and that has `optional` for the fields.

PR #253, opened 2025-04-16
Raise error if mirror != NIL

jurasic-pf:
Looking at a few mgrid files, it seems that NUL and NIL are valid values here, could also just check if 'N' is the first character, but this is stricter.

- Check for NIL
- **Report the status if parsing failed and don't silently return an invalid configuration!!!**
- Report status codes in test cases on failiure

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/magnetic_configuration_lib/magnetic_configuration_lib.cc
+++ b/src/vmecpp/cpp/vmecpp/common/magnetic_configuration_lib/magnetic_configuration_lib.cc
@@ -48,9 +48,14 @@ absl::Status ParseCurrentCarriers(
 
     if (absl::StartsWith(stripped_line, "mirror")) {
       // coils can be mirrored within MAKEGRID --> ignore this line for now
-      // TODO(jons): check if remainder after "mirror" is anything else than
-      // "NIL"
-      //             --> would be an error to ignore this then!
+      if (!(absl::EndsWithIgnoreCase(stripped_line, "NIL") ||
+            absl::EndsWithIgnoreCase(stripped_line, "NUL"))) {
+        return absl::InvalidArgumentError(absl::StrCat(
+            "The magnetic_configuration_lib only supports coilsets for which "
+            "the mirror option is deactivated. Please check your "
+            "coils file, should be 'mirror NUL' but found ",
+            stripped_line));
+      }
       continue;
     } else if (absl::StartsWith(stripped_line, "end")) {
       // current carrier geometry ended on this line
@@ -187,7 +192,7 @@ absl::StatusOr<MagneticConfiguration> ImportMagneticConfigurationFromMakegrid(
         // If something did not work out during parsing, return an empty
         // MagneticConfiguration, since the data is likely messed up anyway.
         magnetic_configuration.Clear();
-        break;
+        return status;
       }
     }
   }
--- a/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
+++ b/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
@@ -17,6 +17,7 @@
 #include "util/netcdf_io/netcdf_io.h"
 #include "util/testing/numerical_comparison_lib.h"
 #include "vmecpp/common/composed_types_lib/composed_types_lib.h"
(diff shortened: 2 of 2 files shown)


jurasic-pf:
* **#253** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/253?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/253?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/proximafusion?ref=merge-queue-instructions-in-stack-comment&prId=253).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

jons-pf:
FYI:
This is where the coils file gets parsed in `MAKEGRID`: https://github.com/ORNL-Fusion/LIBSTELL/blob/master/Sources/Modules/biotsavart.f#L89
Apparently, the `mirror` parameter is then somewhat skipped though? https://github.com/ORNL-Fusion/LIBSTELL/blob/master/Sources/Modules/biotsavart.f#L156
Here, `mirror = NUL` seems to be writted by the Fortran implementation: https://github.com/ORNL-Fusion/LIBSTELL/blob/master/Sources/Modules/biotsavart.f#L562

src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
```diff
@@ -17,6 +17,7 @@
 #include "util/netcdf_io/netcdf_io.h"
 #include "util/testing/numerical_comparison_lib.h"
 #include "vmecpp/common/composed_types_lib/composed_types_lib.h"
+#define ASSERT_OK(quantity) ASSERT_TRUE((quantity).ok()) << (quantity).status();
```
jons-pf, line 20: I would bet that Google internally has something like that available :D

jurasic-pf: One would think so, right... If so, I couldn't find it 😓 
Alternatively we can use a `.value()` accessor, which will throw an exception and also prints the error message nicely:
```log

[ RUN      ] TestMakegridLib.CheckMakeCylindricalGrid
unknown file: Failure
C++ exception with description "Bad StatusOr access: INVALID_ARGUMENT: Not ok" thrown in the test body.

[  FAILED  ] TestMakegridLib.CheckMakeCylindricalGrid (0 ms)
[ RUN      ] TestMakegridLib.CheckComputeMagneticFieldResponseTable
```

I just want to see what happened when a test fails, otherwise all our helpful status messages are hidden.

jurasic-pf:
Thanks!

src/vmecpp/cpp/vmecpp/common/magnetic_configuration_lib/magnetic_configuration_lib.cc
```diff
@@ -48,9 +48,14 @@ absl::Status ParseCurrentCarriers(
...
       // coils can be mirrored within MAKEGRID --> ignore this line for now
-      // TODO(jons): check if remainder after "mirror" is anything else than
-      // "NIL"
-      //             --> would be an error to ignore this then!
+      if (!(absl::EndsWithIgnoreCase(stripped_line, "NIL") ||
+            absl::EndsWithIgnoreCase(stripped_line, "NUL"))) {
+        return absl::InvalidArgumentError(absl::StrCat(
+            "The magnetic_configuration_lib only supports coilsets for which "
+            "the mirror option is deactivated. Please check your "
+            "coils file, should be 'mirror NIL' but found ",
```
jons-pf, line 56: To be consistent with Fortran MAKEGRI, I would suggest to propose `NUL` to the user here in order to be consistent with the files that `biotsavart.f` would produce.

jons-pf (approved):

src/vmecpp/cpp/vmecpp/common/magnetic_configuration_lib/magnetic_configuration_lib.cc
jurasic-pf:
```suggestion
            "coils file, should be 'mirror NUL' but found ",
```

PR #260, opened 2025-04-20
Default initializer for VmecInput (reads values from VmecINDATAPyWrapper)

jurasic-pf:
### What changed?

- Added a custom `__init__` method to `VmecInput` class that initializes default values from the C++ `VmecINDATAPyWrapper` class when no arguments are provided

### Why make this change?
**Pro:** 
- Enables a "pure Python workflow", where the user doesn't have to load any input file. `VmecInput` can be comfortably constructed in a script, without providing explicit defaults for **all quantities**. 

**Con:** 
- Defaults aren't clearly visible for Python users without constructing and inspecting the object.
- only passing a few arguments and defaulting the rest isn't supported, which may be desirable too. 

Alternative: Explicitly default all fields to the same as VmecINDATAPyWrapper

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -338,6 +338,12 @@ class VmecInput(pydantic.BaseModel):
             }
         )
 
+    @staticmethod
+    def default():
+        """Construct a VmecInput with the same default settings as VMEC2000."""
+        cpp_defaults = _vmecpp.VmecINDATAPyWrapper()
+        return VmecInput._from_cpp_vmecindatapywrapper(cpp_defaults)
+
     def _to_cpp_vmecindatapywrapper(self) -> _vmecpp.VmecINDATAPyWrapper:
         cpp_indata = _vmecpp.VmecINDATAPyWrapper()
 
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -380,3 +380,13 @@ def test_aux_arrays_from_cpp_wout():
         assert_aux_defaults(wout)
     np.testing.assert_almost_equal(wout.am_aux_s[:2], np.array([2.0, 3.0]))
     np.testing.assert_almost_equal(wout.am_aux_f[:2], np.array([2.0, 3.0]))
+
+
+def test_default_preset():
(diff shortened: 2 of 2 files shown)


jurasic-pf:
* **#260** <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/260?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/proximafusion/vmecpp/260?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.dev/invite/github/proximafusion?ref=merge-queue-instructions-in-stack-comment&prId=260).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/vmecpp/__init__.py
```diff
@@ -202,6 +202,19 @@ class VmecInput(pydantic.BaseModel):
     zbc: SerializableSparseCoefficientArray | None = None  # [mpol, 2 * ntor + 1]
     """Boundary coefficients for Z ~ cos(m*u - n*v); non-stellarator-symmetric"""
 
+    def __init__(self, *args, **kwargs):
```
eguiraud-pf, line 205: this signature is opaque to users and language servers and breaks autocompletion. this won't work anymore:



Also unclear what happens when passing `*args`.

Overall I'm not sure this is better than before.

Maybe consider using instead a factory method, e.g. `VmecInput.default()` with a docstring that explains what kind of values you get (high-fidelity? low-fidelity? same as VMEC2000?)

eguiraud-pf (approved):

src/vmecpp/__init__.py
jurasic-pf: *args are not supported by pydantic classes, they immediately throw an exception.

I like your factory suggestion, then we can provide multiple presets. VMEC2000 should at least be an option, because those defaults are what users have come to expect.

PR #274, opened 2025-04-25
[flow control] while->for loop

jurasic-pf:
### TL;DR

Turned a while loop into a for loop, whats the big deal?

### Why?
In very rare cases,  optimizations didn't terminate, possibly because VMEC would continuously reset without increasing the iter2 counter. 

-> Refactored to a simpler for loop, ensuring that the iteration count is always updated when entering, to avoid an infinite loop.

>[!NOTE]
This subtley changes the meaning of niter compared to the original implementation. **Now it counts all evaluations of the forward model, not just the successful ones.** 

----------------
**Now the entire control flow consists of two nested for loops with a max iteration count, will definitely terminate.**


jons-pf:
> In very rare cases, optimizations didn't terminate, possibly because VMEC would continuously reset without increasing the iter2 counter.
Do we have an exemplary input file that would trigger this, or can we somehow test that this does not happen again?

jons-pf (approved):

PR #275, opened 2025-04-25
Add ubsan to the CI

jurasic-pf:
### TL;DR

Added Undefined Behavior Sanitizer (UBSan) to the CI workflow alongside the existing Address Sanitizer (ASan).
https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

### Why make this change?

I observed rare instances of nan values in output quantities, and wondered if there are remaining undefined behavior/uninitialized memory issues that ASan doesn't detect:
- Integer overflow
- Null pointer dereference
- Using misaligned pointers
- Signed integer overflow
- Array bounds violations

Two minor UB cases were caught (downstack PRs), but mainly this will ensure we don't introduce UB with upstack refactorings.

### DIFF
--- a/.github/workflows/asan.yaml
+++ b/.github/workflows/asan.yaml
@@ -13,9 +13,12 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  asan:
-    name: Address sanitizer
+  sanitizer:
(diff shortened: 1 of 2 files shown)


eguiraud-pf (approved):
Nice!

PR #278, opened 2025-04-25
Uninitialized bool was triggering ubsan

jurasic-pf:


src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.h
```diff
@@ -1003,13 +1003,13 @@ struct WOutFileContents {
   int nfp;
   int mpol;
   int ntor;
-  bool lasym;
+  bool lasym{};
```
eguiraud-pf, line 1006: if we have to initialize it let's give it a sensible default explicitly:


```suggestion
  bool lasym = false;
```

src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.h
```diff
@@ -1003,13 +1003,13 @@ struct WOutFileContents {
...
 
   int ns;
   double ftolv;
   int maximum_iterations;
 
-  bool lfreeb;
+  bool lfreeb{};
```
eguiraud-pf, line 1012: if we have to initialize it let's give it a sensible default explicitly:

```suggestion
  bool lfreeb = false;
```

eguiraud-pf (approved):
bit of a pity that we have to silence ubsan like this: it's actually really useful to leave the attributes uninitialized for objects that are typically constructed by factory functions, so sanitizers can tell you if the factory function is forgetting something!

if there is a way to instead make sure that we don't return partially uninitialized output quantities from the factory function that might be preferrable.

src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.h
eguiraud-pf: I think it sends a different message: "I want this to default to false, because lasym/lfreeb is more typically false" vs "I want this to default to whatever the type's default is" (which happens to be `false`, sure :D )

eguiraud-pf:
both can be done at the same time in principle, but I have not worked through the logic hence the "if there is a way":
- a checkpoint mechanism that stops the vmec iteration mid-way to ensure until then things went as expected
- all initialization done in the factory function so that ubsan can actually warn us if we forgot something

PR #308, opened 2025-05-05
Generalized wout handling, supports arbitrary extra fields now

jurasic-pf:
## Design Logic
Pydantic provides all kinds of hooks to modify data before serialization and the ability to set a `field_alias`. We combine these features to dump wout files with arbitrary fields, and handle subtle differences betwee `VmecWOut` Python type and the NetCDF type. 

- Special handling (e.g. allow for padding, interpret as int but serialize as float) is done in the pre/post `BeforeValidator` and `PlainSerializer` hooks
- The dimension annotations of the fields == wout dimensions

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -27,16 +27,38 @@ from vmecpp.cpp import _vmecpp  # type: ignore # bindings to the C++ core
 
 logger = logging.getLogger(__name__)
 
-SerializableSparseCoefficientArray = typing.Annotated[
-    jt.Float[np.ndarray, "mpol two_ntor_plus_one"],
+
+_ArrayType = typing.TypeVar("_ArrayType")
+
+
+SerializableSparseCoefficientArray: typing.TypeAlias = typing.Annotated[
+    _ArrayType,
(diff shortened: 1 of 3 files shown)


src/vmecpp/__init__.py
```diff
@@ -27,17 +27,39 @@
...
+
+
+SerializableSparseCoefficientArray: typing.TypeAlias = typing.Annotated[
+    _ArrayType,
     pydantic.PlainSerializer(
         _util.dense_to_sparse_coefficients, when_used="unless-none"
     ),
     pydantic.BeforeValidator(_util.sparse_to_dense_coefficients_implicit),
 ]
+SerializeAsFloat: typing.TypeAlias = typing.Annotated[
```
jons-pf, line 41: maybe call this `SerializeInt32AsFloat`, to mark what is returned when reading?

jons-pf (approved):

jons-pf:
I am not entirely sure about so much custom logic in here in the long term... Probably we should adjust the internal code as much as possible to be a 1:1 match with what is in the wout file, e.g., `volume_p` instead of `volume` etc.

PR #309, opened 2025-05-05
Optional parameter to format VmecInput json for human readability

jurasic-pf:


src/vmecpp/__init__.py
```diff
@@ -375,13 +375,25 @@ def _to_cpp_vmecindatapywrapper(self) -> _vmecpp.VmecINDATAPyWrapper:
...
+        """Serialize the object to JSON.
+
+        Args:
+            improved_readability: Enable to dump non-default values with indentation.
+            **kwargs: Additional keyword arguments to forward to the model_dump_json method.
+        """
+
+        if improved_readability:
+            kwargs.setdefault("indent", 2)
+            kwargs.setdefault("exclude_defaults", True)
```
jons-pf, line 390: I don't see yet how excluding defaults really helps with readability. Putting myself into the situation of someone using this, who is not aware of the default values (as most users probably will be): they dumped the contents with the default `improved_readability=False`, then they want to improve readability, but suddenly a lot of the quantites are gone from the output...
I feel quite strongly for always dumping complete objects, and rather improving/compromising on the visual appearance, but never on completeness and context-free-usability.

jons-pf (approved):

PR #317, opened 2025-05-06
Renamed ivac counter to vacuum_state

jurasic-pf:


eguiraud-pf (approved):
no strong opinion on `vacuum_state` vs `vacuum_pressure_state` on my side, any of the proposed options is much more telling than `ivac`.

maybe not `nestor_state` as VMEC++ tries to be agnostic to the exact solver (not 100% succeeding, but nevertheless)?

PR #319, opened 2025-05-06
Fix normalize by coil currents

jurasic-pf:
## Why?
`normalize_by_coil_currents` had no effect on the resulting response table. I believe there was a mixup of what was meant with `num_windings` and `current`. 

Loading a polygon filament with 4 fields per row, e.g. 
```
x y z w
1 2 3 1e6
```
resulted in `num_windings=1e6` and `current=1`. When normalizing, only current was taken into account and not `num_windings`.


src/vmecpp/cpp/vmecpp/common/magnetic_configuration_lib/magnetic_configuration_lib.cc
```diff
@@ -196,6 +196,11 @@ absl::StatusOr<MagneticConfiguration> ImportMagneticConfigurationFromMakegrid(
       }
     }
   }
+  absl::Status status = NumWindingsToCircuitCurrents(magnetic_configuration);
```
jons-pf, line 199: I don't think this is the right place to do this. The import should store the 4th column in `num_windings`. The migration into the circuit currents, for the normalization to work in the response table computation, should be done only there and not here, I think.

PR #330, opened 2025-05-11
Skip gradual vacuum pressure activation on hot-restart

jurasic-pf:
Usually VMEC turns on the pressure gradually over the first few iterations, to improve robustness. When hot-restarting, we don't want to do that, because the initial guess should already be close to the final solution. Therefore we initialize the vacuum state to `kInitialized` from the beginning, which transitions to `kActive` after the first iteration.

One additional change to the iteration logic is required for everything to work:

The problem is that only the pre-conditioned force-balance includes the vacuum contribution, but the termination criterion (`fsqr<ftol && fsqz<ftol && fsql<ftol`) depends only on the regular, invariant forces. This means a hot-restarted, free-boundary iteration would immediately, if the inner surfaces are in force-balance, eventhough the vacuum force-balance at the LCFS may not be fullfilled at all! 
Additional logic is required to handle that case, and prevent termination at 0 iterations.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
@@ -862,7 +862,9 @@ absl::StatusOr<bool> IdealMhdModel::update(
   // end of bcovar
 
   // back in funct3d, free-boundary force contribution active?
-  if (m_fc_.lfreeb && iter2 > 1) {
+  // This can even happen in the first iteration when hot-restarted.
+  if (m_fc_.lfreeb &&
+      (iter2 > 1 || m_vacuum_pressure_state_ != VacuumPressureState::kOff)) {
     ivacskip = (iter2 - iter1) % nvacskip;
     // when R+Z force residuals are <1e-3, enable vacuum contribution
     if (m_vacuum_pressure_state_ != VacuumPressureState::kActive &&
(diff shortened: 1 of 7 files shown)


src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -887,7 +889,7 @@ absl::StatusOr<bool> IdealMhdModel::update(
         m_last_full_update_nestor = iter2;
       }
     }
-
+#pragma omp barrier
```
jons-pf, line 892:
```suggest
// protects read of `m_vacuum_pressure_state_` below from writing it in the code above
#pragma omp barrier
```

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
```diff
@@ -242,10 +242,12 @@ absl::StatusOr<bool> Vmec::run(const VmecCheckpoint& checkpoint,
     // before the user-provided ns values from ns_array are processed
     // in the multi-grid run
 
-    if (fc_.lfreeb && jacob_off_ == 1) {
-      // jacob_off=1 indicates that in the previous iteration, the Jacobian was
-      // bad
-      // --> also need to restart vacuum calculations
+    // Free-boundary hot-restart attempt. Assuming that the initial_guess is
+    // very good, we can immediately take the vacuum contribution into account,
+    // instead of slowly activating it (as we would in a regular fb solve).
```
jons-pf, line 247:
```suggestion
    // instead of slowly activating it (as we would in a regular free-bdy solve).
```

jons-pf (approved):

PR #334, opened 2025-05-12
WOut quantity docstrings

jurasic-pf:


src/vmecpp/__init__.py
```diff
@@ -205,7 +198,7 @@
     curtor: float
     """Toroidal current in A.
```
jons-pf, line 199:
```suggestion
"""Net toroidal current in A.
```

src/vmecpp/__init__.py
```diff
@@ -691,74 +861,106 @@
...
+    """Total plasma beta. 
+
+    The ratio of the total thermal energy of the plasma,
+    to the total magnetic energy.
+    `beta=W_th/W_B=\int p dV/(\int B/(2 mu_0) dV)`
+    """
 
     # In the C++ WOutFileContents this is called raxis_c.
     raxis_cc: jt.Float[np.ndarray, "ntor_plus_1"]
+    """Fourier coefficients of `R(zeta)` of the magnetic axis geometry."""
```
jons-pf, line 886: let's aim to only have `phi` as toroidal angle anywhere in the code and (over time) fade out `zeta` -> this means that we might need to write `nfp` out explicitly in some places, but `phi` is the real angle, which I think is easier to comprehend than `zeta`

jons-pf (approved):
Fantastic, thanks a lot!

src/vmecpp/__init__.py
jons-pf: I see - when I started writing TNOV, I did not have that as clearly figured out myself :/ Thanks for fixing this now :)

PR #338, opened 2025-06-10
Add AGENTS.md for working with AI

jons-pf:


jons-pf:
Good point, adjusted :)

Issue #376, opened 2025-07-30
Convergence Problem, VMEC++ hangs


missing-user:
The following input hangs vmecpp

```log
 NS = 71   NO. FOURIER MODES = 98   FTOLV = 5.000e-13   NITER = 10000

 ITER |    FSQR     FSQZ     FSQL    |    fsqr     fsqz      fsql   |   DELT   |  RAX(v=0) |    W_MHD   |   <BETA>   |  <M>  
------+------------------------------+------------------------------+----------+-----------+------------+------------+-------
    2 | 4.28e-02  7.33e-03  3.60e-05 | 1.39e-09  1.20e-08  1.75e-07 | 5.00e-01 | 4.911e+00 | 2.1811e+00 | 0.0000e+00 | 1.686
HAVING A CONVERGENCE PROBLEM: RESETTING DELT TO    0.480.  If this does NOT resolve the problem, try changing (decrease OR increase) the value of DELT
HAVING A CONVERGENCE PROBLEM: RESETTING DELT TO    0.480.  If this does NOT resolve the problem, try changing (decrease OR increase) the value of DELT
```

```
RuntimeError: FATAL ERROR in SolveEquilibrium.
Vmec status code: JACOBIAN_75_TIMES_BAD
VmecINDATA had these contents:
[...]

PR #379, opened 2025-08-10
Fix a race condition in ijacobian

jurasic-pf:


src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
```diff
@@ -855,7 +860,9 @@ absl::StatusOr<Vmec::SolveEqLoopStatus> Vmec::SolveEquilibriumLoop(
 #pragma omp single
 #endif  // _OPENMP
       {
-        const double scale = fc_.ijacob == 25 ? 0.98 : 0.96;
+        // fc_.ijacob is incremented in RestartIteration
+        const double scale = fc_.ijacob < 50 ? 0.98 : 0.96;
+
```
jurasic-pf, line 865: This was a bug, ijacob is either 26 or 51 at this point (never 25). 
This becomes apparent when inspecting the logs, e.g. https://github.com/proximafusion/vmecpp/issues/376

```log
HAVING A CONVERGENCE PROBLEM: RESETTING DELT TO    0.480.  If this does NOT resolve the problem, try changing (decrease OR increase) the value of DELT
HAVING A CONVERGENCE PROBLEM: RESETTING DELT TO    0.480.  If this does NOT resolve the problem, try changing (decrease OR increase) the value of DELT
```
has the same delt both times (because we always selected the 0.96 scaling by accident)

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
```diff
@@ -836,6 +836,11 @@ absl::StatusOr<Vmec::SolveEqLoopStatus> Vmec::SolveEquilibriumLoop(
 
     // ADDITIONAL STOPPING CRITERION (set m_liter_flag to FALSE)
 
+#ifdef _OPENMP
+// Protect the reads of fc_.ijacob from writes in the single region below.
+#pragma omp barrier
+#endif  // _OPENMP
+
```
jurasic-pf, line 843: Prevent race condition that caused the infinite hang.

jons-pf: Wow, nice find! I absolutely think that we need to get the TSan/Archer setup working again for VMEC++ now.

jons-pf (approved):

PR #394, opened 2025-11-30
Refactor HandoverStorage to use flat contiguous buffers

jungdaesuh:
## Summary

- Replace nested `std::vector<std::vector<double>>` with flat `std::vector<double>` arrays in HandoverStorage for improved cache locality and reduced pointer chasing in the radial preconditioner / tri-diagonal solver
- Add inline index helpers (`IdxFourier()`, `IdxTridiag()`, `IdxTridiagRhs()`) for flat array access
- Update `TridiagonalSolveSerial` signature to use `std::span` and flat RHS pointer with stride
- Fix potential integer overflow in span accessor pointer arithmetic using `static_cast<std::size_t>`
- Only allocate `lthreed`-dependent arrays when `lthreed=true`, saving memory for axisymmetric (2D) cases
- Add benchmark script for measuring cold start and hot restart performance

## Test plan

- [x] All 14 C++ Bazel tests pass
- [ ] Run benchmark to verify performance improvement
- [ ] Test with 2D (axisymmetric) case to verify memory savings


examples/benchmark_vmec.py
jurasic-pf, line 1: I dont think this should become a general example for vmec++ users. 
We have some benchmarks here, if you feel it should be added, then please open a separate pr and we'll discuss https://github.com/proximafusion/vmecpp-benchmarks

src/vmecpp/cpp/vmecpp/vmec/handover_storage/handover_storage.h
```diff
@@ -94,57 +95,135 @@ class HandoverStorage {
...
-  std::vector<std::vector<std::vector<double>>> all_cr;
-  std::vector<std::vector<std::vector<double>>> all_cz;
-
-  // radial preconditioner; parallel tri-diagonal solver
-  std::vector<std::vector<double>> handover_cR;
-  std::vector<double> handover_aR;
-  std::vector<std::vector<double>> handover_cZ;
+  int ns_ = 0;  // number of radial surfaces (set in allocate)
+
+  std::vector<double> all_ar;  // [mnsize * ns]
```
jurasic-pf, line 153: We are aiming to refactor the use of 2d vectors to Eigen matrices, which are also backed by flat storage.

This refactoring will become a bit harder by this change, because it's no longer obvious from the type that its 2D. 
Could you please use a 2D type like`RowMatrixXd` or `MatrixXd` here, depending on which storage layout is preferred for performance?

jurasic-pf: Then we don't need custom index accessors for everything

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -3423,19 +3423,21 @@ absl::Status IdealMhdModel::applyRZPreconditioner(
     jMax = m_fc_.ns;
   }
 
-  // gather everything into HandoverStorage
+  // gather everything into HandoverStorage (flat layout)
   for (int jF = r_.nsMinF; jF < r_.nsMaxF; ++jF) {
     for (int mn = 0; mn < s_.mnsize; ++mn) {
       int idx_mn = (jF - r_.nsMinF) * s_.mnsize + mn;
-      m_h_.all_ar[mn][jF] = ar[idx_mn];
```
jurasic-pf, line 3430: same as above

input.test
jurasic-pf, line 1: please remove test file

BENCHMARK_RESULTS.txt
jurasic-pf, line 1: Cool results, please attach them to the PR description, but not the commit.

jurasic-pf (changes requested):

PR #396, opened 2025-12-01
Refactor HandoverStorage to use Eigen::RowMatrixXd for improved performance

jungdaesuh:
## Summary

- Replace nested `std::vector<std::vector<double>>` with `Eigen::RowMatrixXd` for improved cache locality and reduced pointer chasing in the radial preconditioner / tri-diagonal solver  
- Only allocate lthreed-dependent arrays when `lthreed=true`, saving memory for axisymmetric (2D) cases  
- Use `static_cast<std::size_t>` in accessor pointer arithmetic to prevent potential integer overflow for large `mn * ns` products  

This PR addresses all feedback from #394:  
- Uses Eigen::RowMatrixXd instead of flat vectors with custom indexers  
- No benchmark files or test artifacts  
- Clean commit history from upstream main  

## Test plan

- [x] All C++ Bazel tests pass  
- [x] All Python tests pass  
- [x] Code builds cleanly with CMake

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
@@ -3462,14 +3462,18 @@ absl::Status IdealMhdModel::applyRZPreconditioner(
 
   // call serial Thomas solver for every mode number individually
   // Uses span accessors to pass contiguous radial slices to the solver
-  const int ns = m_h_.GetNs();
+  const int ns = m_h_.all_ar.cols();
   for (int mn = mnmin; mn < mnmax; ++mn) {
-    TridiagonalSolveSerial(m_h_.TridiagAr(mn), m_h_.TridiagDr(mn),
-                           m_h_.TridiagBr(mn), m_h_.TridiagCrData(mn), ns,
-                           jMin[mn], jMax, s_.num_basis);
-    TridiagonalSolveSerial(m_h_.TridiagAz(mn), m_h_.TridiagDz(mn),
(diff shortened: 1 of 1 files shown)


src/vmecpp/cpp/vmecpp/vmec/handover_storage/handover_storage.cc
```diff
@@ -10,6 +10,7 @@ namespace vmecpp {
 
 HandoverStorage::HandoverStorage(const Sizes* s) : s_(*s) {
   plasmaVolume = 0.0;
+  voli = 0.0;
```
jurasic-pf: ?

jungdaesuh: I was going to tag you after one more review, but you were faster! That did catch my eye, so I was going to take a look at it. Addressed it. Thank you for the review.

https://github.com/proximafusion/vmecpp/pull/396/commits/1a3e44bf13cb573c247211b3beba520246521114

jungdaesuh (commented):
.

jonathanschilling:
Great to see more people working on VMEC++ and thanks in advance for tackling this! Would you mind also testing these changes against the `vmecpp_large_cpp_tests`?

jungdaesuh:
> Great to see more people working on VMEC++ and thanks in advance for tackling this! Would you mind also testing these changes against the `vmecpp_large_cpp_tests`?

- Are you referring to these tests? "https://github.com/proximafusion/vmecpp/actions/runs/19817604086/job/56773928299""

- I use macOS, so the tests failed on my machine. But on HuggingFace, the tests passed. 
Screenshot 2025-12-04 at 1 38 32 AM
Is this right?

PR #400, opened 2026-01-10
Fix #164: rebind spans after copy/move/assignment

### LINKED ISSUE #164: Copy and move semantics of FourierCoeffs and deriving classes are broken
(Issue migrated from old repo, opened by @eguiraud-pf ) 
~~Move-construction and move-assignment of `FourierCoeffs` is disabled by the presence of const data members (e.g. `const Sizes& s_`).~~

~~Code such as~~

```
FourierCoeffs newCoeffs(std::move(oldCoeffs));
```

~~actually compiles but implicitly falls back to the copy constructor.~~
ERRATA: Only move-assignment is implicitly disabled. The move constructor is present and actually works as it should, except for the possibly confusing fact that a moved-from span still points to the original memory.

What's worse: the copy-constructor of deriving classes such as `FourierVelocity` and `FourierGeometry` has broken semantics: `FourierGeometry::rmncc` and the other data members that are just referencing data members of the base class, when a copy happens, will still be referencing the data of the copied-from instance.

P.S.
at the moment this is not causing any problems because each VMEC++ instance instantiates these types once and that's it. We don't actually ever copy or move them around.

jungdaesuh:
## Summary
Fixes Issue #164: FourierGeometry/FourierVelocity/FourierForces store std::span views into FourierCoeffs-owned vectors. The compiler-generated copy/move semantics shallow-copy those spans, so after a copy the spans can still point at the source object's buffers.

This PR implements copy/move constructors and copy/move assignment operators for:
- FourierGeometry
- FourierVelocity
- FourierForces

These operations rebind the span members to the destination object's own storage after copy/move/assignment.

It also adds copy/move assignment operators to FourierCoeffs (const config members delete the implicit operator=).

## Testing
- bazel test --test_output=errors --nocache_test_results //vmecpp_large_cpp_tests/vmec/fourier_geometry:fourier_geometry_test
- bazel test --test_output=errors --nocache_test_results //vmecpp/...

## Limitations / follow-up
- FourierCoeffs keeps config as const refs/ints (Sizes/RadialPartitioning/ns*), so FourierCoeffs::operator= copies/moves only coefficient vectors and does not update config. Assignment is intended only between same-config objects; cross-config assignment is unsupported/undefined.
- Follow-up idea (API-breaking for C++): rule-of-zero value semantics by storing needed config scalars by value and replacing stored span members with span-returning accessors.

### DIFF
--- a/src/vmecpp/simsopt_compat.py
+++ b/src/vmecpp/simsopt_compat.py
@@ -259,8 +259,7 @@ class Vmec(Optimizable):
                 verbose=self.verbose,
                 restart_from=restart_from,
             )
-            wout = self.output_quantities.wout
-            self.wout = wout
+            self.wout = self.output_quantities.wout
         except RuntimeError as e:
             msg = f"Error while running VMEC++: {e}"
             raise ObjectiveFailure(msg) from e
@@ -268,7 +267,7 @@ class Vmec(Optimizable):
         if self._should_save_outputs:
             assert self.input_file is not None
             wout_fname = _make_wout_filename(self.input_file)
-            wout.save(Path(wout_fname))
+            self.wout.save(Path(wout_fname))
             self.output_file = str(wout_fname)
 
         logger.debug("VMEC++ run complete. Now loading output.")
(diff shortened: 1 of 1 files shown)


src/vmecpp/__init__.py
jurasic-pf, line 1: These changes seem unrelated to the PR, could you please cherry pick them and submit a separate PR if needed?

jungdaesuh:
@jurasic-pf Thanks for the review. I will reply to specific comments once I sorted things out. It's like solving one issue keeps leading to another issue.

PR #411, opened 2026-02-11
add free-boundary method that only uses magnetic field from external coils

jons-pf:
…coils for pure vacuum case


src/vmecpp/__init__.py
```diff
@@ -262,6 +262,11 @@ class VmecInput(BaseModelWithNumpy):
     nvacskip: int = 1
     """Number of iterations between full vacuum calculations."""
 
+    free_boundary_method: typing.Annotated[
+        _vmecpp.FreeBoundaryMethod, pydantic.Field()
+    ] = _vmecpp.FreeBoundaryMethod.NESTOR
+    """Method for handling free-boundary conditions."""
+
```
jurasic-pf, line 299: In terms of user experience this should be a string enum I believe

Issue #374, opened 2025-07-25
Allow cancelling a vmecpp run with `Ctrl+C`


jurasic-pf:
Allow cancelling a vmecpp run with `Ctrl+C` by following these instructions:
https://pybind11.readthedocs.io/en/stable/faq.html#how-can-i-properly-handle-ctrl-c-in-long-running-functions

this should probably be called in the optimization loop.
```cpp
if (PyErr_CheckSignals() != 0)
    throw py::error_already_set();
```

Issue #173, opened 2025-03-14
In Jupyiter notebooks no progress output until execution terminates


jurasic-pf:
Running vmecpp in a Jupyter environment does not display any progress in the commandline/Jupyter output, until the program terminates.
It seems like ostream is buffered, and not flushed correctly. There is some info on cout for pybind11 modules here, that might help
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/utilities.html

eguiraud-pf:
Good news is I remembered wrong and it seems the problem is indeed only with concurrent writes, so we could work around it by only printing things from thread 0. We do have some debug printouts from all threads atm.

PR #415, opened 2026-02-11
GIL handling to allow Ctrl+C and progressive logging in Jupyter

jurasic-pf:
Addresses https://github.com/proximafusion/vmecpp/issues/173 and https://github.com/proximafusion/vmecpp/issues/374

We can finally cancel a `vmecpp` run from Python. It had no measurable performance impact in my benchmarks

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/pybind11/pybind_vmec.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/pybind11/pybind_vmec.cc
@@ -649,8 +649,27 @@ PYBIND11_MODULE(_vmecpp, m) {
          std::optional<vmecpp::HotRestartState> initial_state,
          std::optional<int> max_threads,
          bool verbose = true) -> vmecpp::OutputQuantities {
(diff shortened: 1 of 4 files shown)


src/vmecpp/cpp/vmecpp/vmec/pybind11/pybind_vmec.cc
```diff
@@ -649,8 +649,25 @@ PYBIND11_MODULE(_vmecpp, m) {
          std::optional<vmecpp::HotRestartState> initial_state,
          std::optional<int> max_threads,
          bool verbose = true) -> vmecpp::OutputQuantities {
-        auto ret =
-            vmecpp::run(indata, std::move(initial_state), max_threads, verbose);
+        bool was_interrupted = false;
+        auto interrupt_check = [&was_interrupted]() -> bool {
+          if (was_interrupted) return true;
```
jons-pf: nit: I would prefer to always have `{` and `}`, even for trivial checks

PR #525, opened 2026-05-18
README: fix missing word + 'suggest to give' phrasing

CharlesCNorton:
## Summary
- L74: rephrase "suggest to give ... a try" -> "suggest giving ... a try"
- L217: add missing "on" in "different path newer OS-X versions"


jonathanschilling (approved):
Thanks!

jons-pf (approved):

PR #529, opened 2026-05-26
output_quantities: zero-initialize jPS2 axis/edge entries

CharlesCNorton:
## Problem

`ComputeIntermediateThreed1GeometricMagneticQuantities` resizes `jPS2` to `ns`, but the loop that fills it only covers the interior surfaces `jF = 1 .. ns-2`:

```cpp
intermediate.jPS2.resize(fc.ns);
...
for (int jF = 1; jF < fc.ns - 1; ++jF) {
  intermediate.jPS2[jF] = jxbout.jpar2[jF] - ...;
}
```

The axis (`jF = 0`) and edge (`jF = ns-1`) entries are never assigned, yet the whole array is serialized via `Threed1GeometricAndMagneticQuantitiesIntermediate::WriteTo`. Those two entries are read from uninitialized memory, so they differ from run to run. Across repeated runs of the same input I observed `jPS2[0]` come out as `6.4e-310`, `6.6e-310`, and `0.0`, and `jPS2[ns-1]` as `-0.394`, `1.5e-6`, and `0.0`, while the interior is `O(1e10)`.

No physics is affected: the only consumer, `loc_jparPS_perp`, is guarded to `jF = 1 .. ns-2`, so the change is limited to the serialized diagnostic.

## Fix

Zero-initialize the array after the resize, so the two unassigned slots are deterministic.

## Verification

Built before and after, and ran a fixed-boundary (`cth_like_fixed_bdy`) and a free-boundary (`cth_like_free_bdy`) case single-threaded. Every other dataset in the output is bit-for-bit identical between the two builds (433 of 434 datasets equal); the only change is `jPS2`, whose axis and edge entries become `0.0`.


jons-pf (approved):
Thanks!

CharlesCNorton:
looking at ci

CharlesCNorton:
> Thanks!

Np! Sorry, was going to check out the failure.

jurasic-pf:
CI is red due to a billing issue with GitHub, and is unrealt3d to the PR. Thanks! We'll merge when CI is green again

PR #530, opened 2026-05-26
Populate full-grid bsubsmns_full in wout

CharlesCNorton:
## Problem

`WOutFileContents::bsubsmns_full` (full-grid covariant `B_s` Fourier coefficients, a VMEC++ addition over Fortran VMEC) is declared, serialized via `WRITEMEMBER`, and read back via `READMEMBER`, but it is never assigned in `ComputeWOutFileContents`. It is therefore left as a default-constructed Eigen matrix and written to every wout file as an empty `(0, 0)` dataset, on fixed- and free-boundary, 2D and 3D runs alike. The half-grid `bsubsmns` and the realspace full-grid `bsubs_full` (`PutBSubSOnFullGrid` / `ExtrapolateBSubS`) are both computed; only the forward transform into `bsubsmns_full` was missing.

The pybind binding exposes `bsubsmns_full` read/write, so Python consumers currently receive an empty array.

## Fix

Pass `bsubs_full` (`BSubSFull`) into `ComputeWOutFileContents` and forward sine-transform it onto the Nyquist mode set, mirroring the existing half-grid `bsubsmns` loop.

## Verification

Built and ran single-threaded on `solovev` (2D) and `w7x` (3D). `bsubsmns_full` now has shape `(mnmax_nyq, ns)` instead of `(0, 0)`. Because the radial half-to-full interpolation is linear and commutes with the angular DFT, on interior full-grid surfaces `bsubsmns_full[:, jF]` must equal `0.5 * (bsubsmns[:, jF+1] + bsubsmns[:, jF])`; observed max relabs `6.8e-18` (solovev) and `1.1e-15` (w7x). Axis and edge are populated from the realspace extrapolation.

## Test suite

No existing C++ test breaks. `CheckWOutEquality`, the only exact-equality helper that compares `bsubsmns_full`, is used solely in the Save/Load round-trip tests in `output_quantities_io_test.cc`, which compare an object against its own serialized-and-reloaded copy and so stay self-consistent whether the field is empty or populated. `CompareWOut`, used for Fortran-reference V&V, does not compare `bsubsmns_full`. Any externally stored VMEC++-format HDF5 that captured the previously-empty field would now read it populated.


jonathanschilling:
Nice! May I suggest to put this reasoning into a proper unit tests, if it works so nicely?
> Because the radial half-to-full interpolation is linear and commutes with the angular DFT, on interior full-grid surfaces bsubsmns_full[:, jF] must equal 0.5 * (bsubsmns[:, jF+1] + bsubsmns[:, jF]); observed max relabs 6.8e-18 (solovev) and 1.1e-15 (w7x). Axis and edge are populated from the realspace extrapolation.

CharlesCNorton:
> Nice! May I suggest to put this reasoning into a proper unit tests, if it works so nicely?
> 
> > Because the radial half-to-full interpolation is linear and commutes with the angular DFT, on interior full-grid surfaces bsubsmns_full[:, jF] must equal 0.5 * (bsubsmns[:, jF+1] + bsubsmns[:, jF]); observed max relabs 6.8e-18 (solovev) and 1.1e-15 (w7x). Axis and edge are populated from the realspace extrapolation.

👍

jurasic-pf:
CI is red due to a billing issue with GitHub, and is unrelated to the PR. Thanks! We'll merge when CI is green again

jurasic-pf (approved):
Thanks!

PR #531, opened 2026-05-27
radial_profiles: implement spline and analytic profile evaluators

CharlesCNorton:
## Problem

`RadialProfiles::evalProfileFunction` dispatches to per-parameterization evaluators, but nine of them returned `0.0` as placeholders: `gauss_trunc`, `two_power_gs`, `akima_spline`/`akima_spline_i`/`akima_spline_ip`, `cubic_spline`/`cubic_spline_i`/`cubic_spline_ip`, `pedestal`, `rational`, and `nice_quadratic`. Any input selecting one of these for its pressure, current, or iota profile silently produced a zero profile rather than the requested parameterization.

## Fix

Port each evaluator from educational_VMEC:
- `gauss_trunc`, `pedestal`, and `rational` from the `pcurr`/`piota`/`pmass` cases in `profile_functions.f`;
- `two_power_gs` from `functions.f`;
- the four splines from `spline_akima.f`, `spline_akima_int.f`, `spline_cubic.f`, and `spline_cubic_int.f`.

The truncated-Gaussian and two-power-with-Gaussian current profiles integrate `I-prime(s)` with the same 10-point Gauss-Legendre rule `pcurr` uses. `gauss_trunc`, `two_power_gs`, and `pedestal` serve both the pressure and current profiles, so they take `shouldIntegrate` like the existing `evalTwoPower`, and the dispatcher threads it through. Spline coefficient construction is shared between the direct and integrated variants via file-local helpers (`BuildAkimaCoeffs`, `BuildCubicSecondDerivatives`), reproduced from the Fortran including the `spline_akima.f` edge convention where both right-edge phantom points reuse the left curvature.

## Verification

Numerical agreement with compiled educational_VMEC: a driver links the real `spline_*.f`, `functions.f`, and `profile_functions.f` and prints their output at a fixed set of inputs; `MatchesFortranReference` asserts the in-repo evaluators reproduce those values at 37 points to `1e-11`. Inverting the `spline_akima.f` left/right edge-curvature convention turns that test red while every invariant-only test stays green, confirming it discriminates.

Full equilibrium: `cth_like_fixed_bdy` with its `two_power` pressure `(1 - s^5)^10` re-expressed as a 201-knot `cubic_spline` (reproducing the analytic profile to `5e-9`) runs to convergence, and the converged `wout` matches the educational_VMEC golden `wout_cth_like_fixed_bdy.nc` within the case's `1e-6` tolerance. The worst deviation normalized by field peak is `3.6e-9`, on the spline-driven pressure.

## Test suite

New `//vmecpp/vmec/radial_profiles:radial_profiles_test` with 16 cases: the spline and analytic invariants, the educational_VMEC reference comparison, and two cases that drive `setupInputProfiles` so spline and analytic profiles reach the evaluators with the aux arrays wired through. `//vmecpp/vmec/output_quantities:output_quantities_test` gains the spline-equilibrium case and runs 7/7, with the six analytic V&V cases unchanged. No existing test or example case uses any of the nine parameterizations, and the already-live `power_series` and `two_power` dispatch and bodies are unchanged, so no existing case is affected.


jurasic-pf:
CI is red due to a billing issue with GitHub, and is unrelated to the PR. Thanks! We'll merge when CI is green again

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
```diff
@@ -19,6 +22,175 @@
...
+    0.01304673574141414, 0.06746831665550774, 0.1602952158504878,
+    0.2833023029353764,  0.4255628305091844,  0.5744371694908156,
+    0.7166976970646236,  0.8397047841495122,  0.9325316833444923,
+    0.9869532642585859};
+constexpr std::array<double, 10> kGaussLegendreWeights01 = {
+    0.03333567215434407, 0.0747256745752903, 0.1095431812579910,
+    0.1346333596549982,  0.1477621123573764, 0.1477621123573764,
+    0.1346333596549982,  0.1095431812579910, 0.0747256745752903,
+    0.03333567215434407};
+
```
jurasic-pf, line 38: Can we static_assert that these coefficients are self consistent GL nodes + weights?

Testing on e.g. a constant and linear profile would be enough imo.
something like
```cpp
template <typename Profile>
constexpr double integrate(Profile profile) {
  double result = 0.0;

  for (std::size_t i = 0; i < kGaussLegendreNodes01.size(); ++i) {
    result += kGaussLegendreWeights01[i] * profile(kGaussLegendreNodes01[i]);
  }

  return result;
}

static_assert(integrate_gl_01([](double) { return 2.0; }) == 2.0);
static_assert(integrate_gl_01([](double x) { return x; }) == 0.5);
```

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
```diff
@@ -19,6 +22,175 @@
...
+  double c =
+      ((ya[2] - ya[0]) / (xa[2] - xa[0]) - (ya[1] - ya[0]) / (xa[1] - xa[0])) /
+      (xa[2] - xa[1]);
+  const double yp1 = (ya[1] - ya[0]) / (xa[1] - xa[0]) - c * (xa[1] - xa[0]);
+  c = ((ya[n - 3] - ya[n - 1]) / (xa[n - 3] - xa[n - 1]) -
+       (ya[n - 2] - ya[n - 1]) / (xa[n - 2] - xa[n - 1])) /
+      (xa[n - 3] - xa[n - 2]);
+  const double ypn = (ya[n - 2] - ya[n - 1]) / (xa[n - 2] - xa[n - 1]) -
+                     c * (xa[n - 2] - xa[n - 1]);
+  std::vector<double> y2(n, 0.0), u(n, 0.0);
```
jurasic-pf: Why the mixing of `std::vector` and `Eigen::VectorXd` here? We tend to prefer `Eigen::VectorXd` where possible nowadays.

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
```diff
@@ -19,6 +22,175 @@
...
+  const double cr = (M(iv - 2) - M(iv - 1)) / (X(iv) - X(iv - 2));
+  const double br = M(iv - 2) - cr * (X(iv - 1) - X(iv - 2));
+  Y(0) = Y(1) + bl * (X(0) - X(1)) + cl * std::pow(X(0) - X(1), 2);
+  Y(-1) = Y(1) + bl * (X(-1) - X(1)) + cl * std::pow(X(-1) - X(1), 2);
+  // NOTE: the right-edge extrapolation reuses cl (the left curvature), not cr.
+  // This is reproduced verbatim from Fortran VMEC for bit-for-bit agreement.
+  Y(iv + 1) =
+      Y(iv) + br * (X(iv + 1) - X(iv)) + cl * std::pow(X(iv + 1) - X(iv), 2);
+  Y(iv + 2) =
+      Y(iv) + br * (X(iv + 2) - X(iv)) + cl * std::pow(X(iv + 2) - X(iv), 2);
```
jurasic-pf: This seems like a bug - while we kept many of the historical quirks of VMEC for compatibility reasons, this is an instance where fixing the incorrect behavior is warranted imo.

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
```diff
@@ -575,66 +768,266 @@ double RadialProfiles::evalTwoPower(const Eigen::VectorXd& coeffs, double x,
...
+    return 0.0;
+  }
+  const AkimaCoeffs coeffs = BuildAkimaCoeffs(splineKnots, splineValues);
+  const int iv = coeffs.iv;
+  auto X = [&](int f) { return coeffs.xloc[f + 1]; };
+  auto A = [&](int f) { return coeffs.a[f + 1]; };
+  auto B = [&](int f) { return coeffs.b[f + 1]; };
+  auto C = [&](int f) { return coeffs.c[f + 1]; };
+  auto D = [&](int f) { return coeffs.d[f + 1]; };
+  constexpr double kThird = 1.0 / 3.0;
```
jurasic-pf, line 875: Nit: Can we inline these +1 index offsets instead of using accessor functions, so the integral evaluation is more comparable to the regular `evalAkima` above?

jurasic-pf (commented):
I left a few minor comments, other than that it looks great! Thanks for the contribution.

jurasic-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
CharlesCNorton: Done in 3e28aa9. Added a constexpr `GaussLegendreIntegral01` plus static_asserts for a constant and a linear integrand. The tabulated weights sum to 1 and integrate x to 0.5 bit-exactly, so the exact `== 2.0` / `== 0.5` form you suggested compiles as written.

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
CharlesCNorton: Done in 3e28aa9. Switched the `y2`/`u` work arrays and the return type of `BuildCubicSecondDerivatives` to `Eigen::VectorXd`.

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
CharlesCNorton: Fixed in 3e28aa9: the right-edge phantom points now use the right-edge curvature `cr` instead of `cl`. One subtlety worth flagging: the `cr` that was already computed was sign-flipped relative to the true right-edge curvature (it was dead code, never used), so I also corrected the `cr`/`br` formulas to mirror the left edge. Since the corrected edge now diverges from spline_akima.f by design, I dropped the Akima entries from the educational_VMEC golden comparison (reproducing them would re-import the bug) and added `AkimaIsReflectionSymmetric`, which fails on the old `cl` behavior and passes on the fix.

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
CharlesCNorton: Done in 3e28aa9. Inlined the `coeffs.a[i + 1]` offsets and removed the `A`/`B`/`C`/`D` accessor lambdas so the integral evaluation reads like `evalAkima` above.

CharlesCNorton:
> I left a few minor comments, other than that it looks great! Thanks for the contribution.

All addressed @jurasic-pf! Thank you for your review and patience.

jurasic-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
jonathanschilling: bonus points: propose a fix against edu_VMEC (and corresponding unit test, which fails in current state and succeeds with this fix), then ensure 1:1 compatibility against VMEC++ again :)

jonathanschilling:
Thanks a lot @CharlesCNorton !

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
CharlesCNorton: ok give me a few mins

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles_test.cc
```diff
@@ -0,0 +1,451 @@
...
+// real educational_VMEC routines (spline_cubic.f, spline_cubic_int.f,
+// functions.f, profile_functions.f) and prints their output at exactly the
+// inputs used here. See that driver (plus fref/stubs.f90, the minimal module
+// shim) to regenerate. The pure-arithmetic cubic spline agrees to ~1e-13; the
+// closed-form profiles use exp/pow/tanh whose last ulp differs between the
+// Windows libm (golden) and the Linux libm (this test), so the tolerance is
+// held at 1e-11, still many orders of magnitude tighter than any indexing or
+// formula error would produce.
+//
+// The Akima spline is intentionally absent here: its right-edge curvature is
```
jonathanschilling: How nice would it be if that was fixed in `educational_VMEC`? 😍

jonathanschilling (approved):
Very nice indeed!

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
CharlesCNorton: Done in jonathanschilling/educational_VMEC#3. `spline_akima.f` and `spline_akima_int.f` now form `cr`/`br` as the mirror of the left edge and use the right-edge curvature `cr` (not `cl`) for the upper phantom points.

That PR adds `test/unit/test_spline_akima.f90`, wired via CTest as `spline_akima_reflection_symmetry`: it interpolates a dataset and its mirror image about the knot span and checks that they agree at mirrored query points. The maximum reflection asymmetry is `1.5e-1` against the previous right edge and `2.2e-16` with the fix, so it fails before the change and passes after.

With that fix in place, `evalAkima` and `evalAkimaIntegrated` reproduce the Fortran reference again, so I restored the `akima` and `akima_int` rows in `MatchesFortranReference` here, with golden values regenerated from the corrected routines and agreement at `1e-11`. `AkimaIsReflectionSymmetric` stays as an additional invariant on the corrected edge.

PR #533, opened 2026-05-27
free_boundary: implement non-stellarator-symmetric surface geometry

CharlesCNorton:
## Problem

The free-boundary `SurfaceGeometry::inverseDFT` receives the full set of boundary Fourier coefficients, including the non-stellarator-symmetric (`lasym`) antisymmetric arrays `rSC`, `rCS`, `zCC`, `zSS`, but discarded them: the four spans were consumed by `(void)` casts and never entered the transform.

```cpp
// TODO(jons): implement lasym-related code
(void)rSC;
(void)rCS;
(void)zCC;
(void)zSS;
```

The surface was evaluated only on the first poloidal half `[0, pi]` and, in the stellarator-symmetric case, mirrored into `]pi, 2pi[`. With `lasym = true` the antisymmetric content was dropped and the second poloidal half was never populated from it; the assembly site was a stub:

```cpp
if (s_.lasym) {
  // mirror quantities into respective
  // non-symmetric other half of poloidal interval ]pi,2pi[
  // TODO(jons)
}
```

As a result, non-stellarator-symmetric boundaries (up-down-asymmetric tokamaks, non-symmetric stellarators) could not be represented in the free-boundary surface geometry.

## Change

Port the antisymmetric surface evaluation and the full-range assembly from educational_VMEC (`surface.f90`, `symrzl.f90`, `totzspa`):

- `inverseDFT` accumulates the antisymmetric parities alongside the symmetric ones, guarded by `s_.lasym`. `R` gains `rSC sin(mu) cos(nv) + rCS cos(mu) sin(nv)` and `Z` gains `zCC cos(mu) cos(nv) + zSS sin(mu) sin(nv)`, with the matching first- and second-order poloidal and toroidal derivatives.
- The full poloidal range is assembled from the symmetric and antisymmetric pieces following `symrzl`: the second half `]pi, 2pi[` is the parity-signed mirror of `(symmetric - antisymmetric)` at the reflected point `(theta -> 2pi - theta, zeta -> 2pi - zeta)`, and the first half `[0, pi]` is their sum. `R`, `Ruu`, `Ruv`, `Rvv`, `Zu`, `Zv` are even under the reflection (`+sym - asym`); `Ru`, `Rv`, `Z`, `Zuu`, `Zuv`, `Zvv` are odd (`-sym + asym`).
- For `lasym`, each thread evaluates the surface derivatives over the full reduced poloidal range so it can perform that combination itself; the stellarator-symmetric case keeps the thread-local slice. The two are selected by an index offset (`derivOffset`), leaving the stellarator-symmetric path byte-unchanged.

## Verification

`//vmecpp/free_boundary/surface_geometry:surface_geometry_test` is present in the tree but had no Bazel target and was not built; this change wires it in and adds the asymmetric cases.

Stellarator-symmetric: `CheckConstants`, `CheckInvDFT`, and `CheckDerivedQuantities` (the W7-X LCFS mockup) pass at `1e-12`, unchanged by the refactor.

Non-stellarator-symmetric: two new cases drive a 7-mode, `nfp = 3` asymmetric boundary with nonzero `rbs` and `zbc`. `CheckInvDFTAsymmetric` compares `R`, `Z`, and their first- and second-order tangential derivatives over the full poloidal range against an independent analytic inverse DFT that includes the antisymmetric terms (`1e-12`), and against a central finite difference of `R` and `Z` (`1e-6`). `CheckDerivedQuantitiesAsymmetric` compares the surface normal `(snr, snv, snz)`, the metric elements `guu`, `guv`, `gvv`, the second-fundamental-form combinations `auu`, `auv`, `avv`, and `drv`, `rzb2`, `rcosuv`, `rsinuv` against vector reference constructions (`1e-12`). All five cases pass.

## Scope

This covers the free-boundary surface geometry. End-to-end non-stellarator-symmetric equilibrium validation against an educational_VMEC golden is a separate step, since the repository carries no asymmetric reference case yet and that golden has to be generated first. The `lasym` branch of the covariant-`B` output split in `output_quantities` is unchanged here.


CharlesCNorton:
checking

src/vmecpp/cpp/vmecpp/free_boundary/surface_geometry/surface_geometry.h
```diff
@@ -149,6 +149,25 @@ class SurfaceGeometry {
...
+  std::vector<double> rub_asym;
+  std::vector<double> rvb_asym;
+  std::vector<double> zub_asym;
+  std::vector<double> zvb_asym;
+  std::vector<double> ruu_asym;
+  std::vector<double> ruv_asym;
+  std::vector<double> rvv_asym;
+  std::vector<double> zuu_asym;
+  std::vector<double> zuv_asym;
+  std::vector<double> zvv_asym;
```
jurasic-pf: Why std vector instead of eigen like the rest of this file?

CharlesCNorton: Done in 83ff29e. Switched the twelve `_asym` arrays to `Eigen::VectorXd` and the fills to `.setZero()`, matching the rest of the file and the #543 migration; the now-unused `absl/algorithm/container.h` include is dropped. The arrays are sized `nZnT == nThetaEven*nZeta`, so behavior is unchanged.

jurasic-pf (approved):

jurasic-pf:
Amazing, thank you!

PR #534, opened 2026-05-27
tests: cross-check vmecpp/test_data and vmecpp_large_cpp_tests inputs via indata2json

CharlesCNorton:
## Problem

The fixed- and free-boundary test cases are defined in two places: the public-facing input files in `vmecpp/test_data` and the copies in `vmecpp_large_cpp_tests/test_data` that drive the integration tests. The latter carry an extra block of `dump_*` flags (and `iter2_to_dump`) so that Fortran VMEC writes out the per-step reference data the large C++ tests compare against; those flags are intentionally kept out of the more public-facing inputs. Because the two copies differ textually by that block, a plain file diff cannot confirm that they still describe the same equilibrium, so the physics inputs can drift apart unnoticed.

## Change

Add `tests/test_input_data_consistency.py`. For every case whose `input.<name>` exists in both `vmecpp/test_data` and `vmecpp_large_cpp_tests/test_data`, it converts both with `indata2json` and asserts the JSON representations are equal. `indata2json` does not emit the `dump_*` instrumentation, so the comparison sees only the physics fields; the `mgrid_file` entry is an environment-specific path and is ignored. Each case is additionally compared against the committed `<name>.json` that the C++ tests load, pinning the public input, its Fortran twin, and the instrumented copy to the same equilibrium.

The shared cases are discovered at runtime. The test skips when the `vmecpp_large_cpp_tests` checkout is absent; it is located where the integration CI checks it out (`src/vmecpp/cpp/vmecpp_large_cpp_tests`), or via the `VMECPP_LARGE_CPP_TESTS_TEST_DATA` environment variable. The Python test workflow now checks out `vmecpp_large_cpp_tests` next to the sources so the comparison runs on every pull request.

## Verification

With `vmecpp_large_cpp_tests` checked out, the test collects the seven shared cases (`cma`, `cth_like_fixed_bdy`, `cth_like_fixed_bdy_nzeta_37`, `cth_like_free_bdy`, `solovev`, `solovev_analytical`, `solovev_no_axis`) plus the discovery guard, and passes 8/8. All seven are identical after conversion; `mgrid_file` is the only differing key and is ignored. The non-overlapping cases are intentional and left unchecked: `li383_low_res` and `up_down_asymmetric_tokamak` exist only in `vmecpp/test_data`, and `solovev_free_bdy` only in `vmecpp_large_cpp_tests/test_data`. No existing test is affected.


tests/test_input_data_consistency.py
```diff
@@ -0,0 +1,110 @@
...
+import pytest
+
+from vmecpp import _util
+
+# tests/ live in the sources while the vmecpp module is installed separately,
+# so locate the data via the path to this file (see tests/test_util.py).
+REPO_ROOT = Path(__file__).parent.parent
+MAIN_TEST_DATA = REPO_ROOT / "src" / "vmecpp" / "cpp" / "vmecpp" / "test_data"
+
+_LARGE_ENV = os.environ.get("VMECPP_LARGE_CPP_TESTS_TEST_DATA")
```
jurasic-pf, line 34: Is the env var intended for local testing if users have the large cpp tests checked out to another location?

jurasic-pf (approved):
This is a nice solution to the consistency check, thanks!

tests/test_input_data_consistency.py
CharlesCNorton: Yes, exactly. In the integration CI the large-tests repo is checked out at `src/vmecpp/cpp/vmecpp_large_cpp_tests` (the same path `test_bazel.yaml` already uses), which is the default this check looks for. `VMECPP_LARGE_CPP_TESTS_TEST_DATA` is for local runs where it is cloned elsewhere: it points the check at that `test_data` directory instead. If neither is present the test skips. Thanks for the review!

PR #535, opened 2026-05-27
free_boundary: multi-grid test case and opt-in to persist the Nestor solver across grid steps

CharlesCNorton:
## Problem
VMEC++ rebuilt the Nestor free-boundary solver from scratch at every multi-grid step, re-initialising its vacuum-response matrix and right-hand side. Fortran VMEC keeps that state in memory across grid steps. The solver's buffers (the response matrix, the RHS, the pivots, and the vacuum field and pressure spans) live in ns-independent Fourier space, so rebuilding the solver per step discarded reusable state without changing the converged result. Separately, every free-boundary test case was single-grid, so the multi-grid free-boundary path had no dedicated case.

## Change
- Keep the free-boundary (Nestor) solver and its tangential partitioning across multi-grid steps instead of rebuilding them per step, reproducing Fortran VMEC's persistent vacuum state. The solver is constructed on the first grid step (when none exists yet) and reused on subsequent steps; single-grid runs build it once as before.
- Add the `cth_like_free_bdy_multigrid` test case: the existing `cth_like_free_bdy` boundary, coils and mgrid with `ns_array = [15, 25]` (one additional multi-grid step), wired into a `test_data` filegroup and the `vmec_test` data set.

## Verification
- New `vmec_test` case `MultiGridFreeBoundary` runs the two-grid free-boundary equilibrium to convergence (ns 15 -> 25, force residual ~1e-10); the ns=25 step reuses the solver built at ns=15.
- The converged equilibrium reproduces an educational_VMEC golden generated from the same input to a worst-case relative difference of 7.3e-6 (ctor to machine precision, b0 4.4e-6, volume 4e-6).
- `vmec_test`, `vmec_in_memory_mgrid_test`, `output_quantities_test` and `vmec_indata_test` pass; the symmetric and single-grid cases are unchanged.


src/vmecpp/__init__.py
```diff
@@ -353,6 +353,9 @@ class VmecInput(BaseModelWithNumpy):
     lforbal: bool = False
     """Hack: directly compute innermost flux surface geometry from radial force balance"""
 
+    nestor_persistent_rhs: bool = False
+    """Persist the Nestor (vacuum) solver state across multi-grid steps to reproduce Fortran VMEC's behaviour 1:1. Default false keeps the historical VMEC++ behaviour of re-initialising the solver per multi-grid step."""
+
```
jurasic-pf: I don't think this is necessary, the new behaviour should be strictly better

CharlesCNorton: Good point, done. The Nestor solver and its vacuum-response matrix and RHS live in ns-independent Fourier space, so reusing them across multi-grid steps is always valid and the converged result is unchanged. I removed the input field and made the persistence unconditional: the solver is built on the first grid step and reused afterwards, with single-grid runs building it once as before. The multi-grid test case and the full regression are green.

jurasic-pf (approved):

PR #536, opened 2026-05-28
free_boundary: implement axisymmetric (nzeta=1) free-boundary support

CharlesCNorton:
## Problem
VMEC++ did not support axisymmetric (`ntor = 0`, `nzeta = 1`) free-boundary equilibria. Running such a case (a tokamak with external coils) crashed in the Nestor free-boundary chain, because several routines assumed a non-trivial toroidal direction:

- `HandOverBoundaryGeometry` read the `sin(n v)` boundary Fourier coefficients (`rmnss`, `zmncs`, and the lasym `rmncs`, `zmnss`). These are only allocated when `lthreed`, so for `ntor = 0` it indexed empty arrays and segfaulted.
- The net-toroidal-current axis filament (`external_magnetic_field`) and the Green's-function toroidal integral (`regularized_integrals`) were built over the `nfp` field periods. For a tokamak (`nfp = 1`, `nzeta = 1`) that collapses the toroidal direction to a single point: the axis filament degenerated to a zero-length loop, so the plasma current produced no boundary poloidal field and never entered the vacuum solution (failing the VMEC/vacuum toroidal-current consistency check), and the Green's function was evaluated from a single toroidal sample.

## Change
- `HandOverBoundaryGeometry`: guard the `sin(n v)` source reads with `lthreed` (the corresponding `*_LCFS` targets stay zero, which is correct for `ntor = 0`), mirroring the `FourierGeometry` allocation. This also covers the axisymmetric `lasym` case.
- `external_magnetic_field`: replicate the axis-current filament over `nvper` equally-spaced toroidal angles (`64` for an axisymmetric plasma, the number of field periods otherwise), matching educational_VMEC's `nvper`, so the net toroidal current produces the correct boundary poloidal field.
- `regularized_integrals`: add an axisymmetric path that performs the Green's-function toroidal integral by summing over `nvper` toroidal images of the evaluation point, subtracting the analytic approximation (with the toroidal-angle factor of each image) at every image. The closed-form analytic integral added back in `SingularIntegrals` is grid-independent and unchanged.
- Add the `solovev_free_bdy` test case (the educational_VMEC free-boundary Solov'ev tokamak: 13 PF/CS circuits, `ns_array = [16, 32]`) and register it in `vmec_in_memory_mgrid_test`, driving it with an in-memory mgrid generated from `coils.solovev`.

## Verification
- `solovev_free_bdy` converges to `FTOL = 1e-14` (`ns = 32`, force residual ~1e-14) and reproduces an educational_VMEC free-boundary reference generated from the same coils and boundary: the Fortran iteration trajectory agrees step for step (`R_axis` 4.095, `W_MHD` 2.4596, vacuum/plasma pressure mismatch `DEL-BSQ` 6.64e-5 to four figures), and the converged `wout` scalars (`wb`, `betatotal`, `aspect`, `b0`, `volume`, `Aminor_p`, `Rmajor_p`, `ctor`, `volavgB`) agree to < 1e-4.
- Full regression passes (`vmec_test`, `vmec_in_memory_mgrid_test`, `output_quantities_test`, `vmec_indata_test`, `laplace_solver_test`, `singular_integrals_test`). The stellarator free-boundary path (`cth_like_free_bdy`) is unchanged, as the new behaviour is guarded by `nZeta == 1` / `lthreed`.


jurasic-pf (commented):
The changes look good, but could we add a test of some sort? solovev free boundary?

jurasic-pf:
i think @jons-pf started working on that in 
https://github.com/proximafusion/vmecpp/pull/462

CharlesCNorton:
Thanks! Added a solovev free-boundary regression against an educational_VMEC golden in ea6e494. It came together separately from #462, but happy to reconcile with that if useful.

jurasic-pf (approved):
Cool!

.github/workflows/benchmarks.yaml
jurasic-pf, line 1: please remove ci changes from this pr

CharlesCNorton: Sorry about that. Reverted in 5d7b7f4; benchmarks.yaml matches main again, so the PR no longer touches any workflow.

PR #537, opened 2026-05-28
output_quantities: implement lbsubs full-grid B_s via radial force balance

CharlesCNorton:
## Problem

`lbsubs` was listed as unimplemented in the README, and `ComputeOutputQuantities` left the full-grid covariant `B_s` as the linear half-to-full interpolation of the metric-element `B_s`, with a standing `TODO` at the call site (`re-compute B_s to solve radial force balance (lbsubs flag in Fortran VMEC)`). The `jxbout` diagnostic that consumes the `B_s` tangential derivatives — `itheta`/`izeta` (the `sqrt(g)` `K^{u,v}` currents) and `bsubs3` — therefore omits the current-sheet correction that the `lbsubs` path captures.

## Change

Add a `VmecINDATA::lbsubs` flag (parsed from JSON, persisted to/from HDF5, round-tripped through `ToJson`; default `false`) and implement `RecomputeBSubSFromRadialForceBalance`, a port of the `IF (lbsubs)` branch of `jxbforce.f90` together with the collocation solve in `getbsubs.f90`. On each interior full-grid surface it solves

```
bsupu * d(B_s)/du + bsupv * d(B_s)/dv = brho
```

in spectral space, where `brho` is the `sqrt(g)`-weighted radial-force residual with the flux-surface-averaged force balance subtracted (the local equation is otherwise unsolvable, since that surface integral vanishes identically). The recovered coefficients overwrite the interior `bsubs_full` and its tangential derivatives `bsubsu`/`bsubsv`; axis and edge are filled by the existing `ExtrapolateBSubS`. The solve is guarded: a singular matrix or a residual above `1e-8 * |rhs|` logs a warning and leaves that surface at its half-to-full interpolation. Both the stellarator-symmetric and the non-stellarator-symmetric (cos-basis) coefficient layouts follow `getbsubs.f90`. With `lbsubs == false` the output path is unchanged.

## Verification

New `lbsubs_test` (gtest), solovev fixture:
- `lbsubs == false` leaves `bsubs_full` equal to `0.5 * (bsubs_half[:, jF] + bsubs_half[:, jF-1])` on every interior surface (max abs diff `< 1e-12`).
- `lbsubs == true` changes `bsubs_full` measurably against the `false` baseline (guards against a silent no-op).
- the recovered `B_s` satisfies `bsupu * dB_s/du + bsupv * dB_s/dv = brho` at every interior collocation point, residual `< 1e-7 * |rhs|`, with the RHS re-derived independently of the implementation's scratch buffers.
- the `lbsubs` JSON flag round-trips.

Local run: 4 passed, 1 skipped. The skipped case parses `up_down_asym.json` (`lasym = true`) and validates input handling but does not run the solve, since the equilibrium iteration does not yet support `lasym = true`; the non-stellarator-symmetric end-to-end path is deferred. README updated to drop `lbsubs` from the unimplemented list.


CharlesCNorton:
checking

src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.h
```diff
@@ -207,6 +207,15 @@ class VmecINDATA {
   // balance
   bool lforbal;
 
+  // If true, re-compute the full-grid covariant B_s by solving the radial
+  // force balance equation
+  //   bsupu * d(B_s)/du + bsupv * d(B_s)/dv = brho
+  // in spectral space at every interior surface (lbsubs flag in Fortran VMEC).
+  // The default (false) keeps B_s as the linear half->full interpolation of
+  // the metric-element B_s plus the axis/edge extrapolation. Only affects the
+  // jxbout diagnostic; the equilibrium iteration is unchanged.
```
jons-pf: jxbout diagnostic and the bsubsmn{s,c} coefficients in the wout file. ...

jurasic-pf (approved):

jurasic-pf:
Can we check the results against PARVMEC with lbsubs true and resolve CI? Other than that it's ready to merge

CharlesCNorton:
👍 good to go! Checked against PARVMEC with `lbsubs = T` on solovev and cth_like — B_s matches to 1e-5 or better.

CharlesCNorton:
Wider PARVMEC sweep, in case it's useful. Interior full-grid `B_s`, as max abs difference between the two codes' `lbsubs` corrections relative to the correction magnitude:

| case | ns | nzeta | agreement |
| --- | --- | --- | --- |
| solovev | 55 | 1 | 2.3e-7 |
| solovev_no_axis | 11 | 1 | 5.3e-10 |
| solovev_analytical | 31 | 1 | 3.6e-8 |
| cth_like_fixed_bdy | 25 | 36 | 1.4e-4 |
| cma | 51 | 16 | 1.4e-1 |
| cth_like_fixed_bdy_nzeta_37 | 25 | 37 | no correction in either code |

cma looks like a difference between the Fortran generations rather than a VMEC++ one: VMEC++ reproduces the 8.52 `jxbout_cma.nc` reference to 3.5e-12, while the PARVMEC 10.0 run differs from that same reference by 3.1e-3, and the collocation solve amplifies it.

cth_like_fixed_bdy_nzeta_37 has odd `nzeta`, so the collocation row count cannot match `itotal` (299 expected, 297 built) and every interior surface falls back to the half-to-full interpolation. PARVMEC reports the same two counts and does the same thing, so `lbsubs` is a no-op for odd `nzeta` in both codes.

jurasic-pf:
I think we still need to sanity check this with some plots on our side. Code-wise, the PR looks good.

PR #538, opened 2026-05-28
free_boundary: non-stellarator-symmetric (lasym) free-boundary support

CharlesCNorton:
## Summary

Adds the non-stellarator-symmetric (`lasym = true`) path through the ideal-MHD geometry/force transforms and the free-boundary vacuum solver, so configurations without up-down symmetry run both fixed- and free-boundary.

## Implementation

- Geometry/force transforms (`ideal_mhd_model`, `dft_toroidal`): antisymmetric-parity inverse and forward DFTs, and the full-poloidal-range recombination of the symmetric and antisymmetric parts (`symrzl` on the inverse transform, `symforce` on the forces).
- Vacuum solver (`laplace_solver`, `nestor`, `vmec`): a four-block (sin/cos) vacuum response matrix with a doubled right-hand side; the potential inverse-DFT and the LU factorization/solve are sized for `2*mnpd`. The poloidal basis is tabulated on the reduced theta range, so for `l >= nThetaReduced` the lasym path reflects via stellarator symmetry (cosine even, sine odd) in the singular-integral source, the cos-basis matrix projection, and the real-space vacuum field.
- Surface geometry (`surface_geometry`): non-stellarator-symmetric surface and derivative quantities over the full poloidal interval.
- Outputs (`output_quantities`): the antisymmetric Fourier coefficient sets in the `wout` computation.

## Verification

- Fixed-boundary lasym: a zero-asymmetry degenerate case reduces to the stellarator-symmetric result, and an up-down-asymmetric tokamak matches a VMEC 8.52 (educational_VMEC) reference.
- Free-boundary lasym: the zero-asymmetry degenerate case reproduces the symmetric free-boundary equilibrium (`wout` scalars and Fourier geometry agree to the vacuum-solve round-off floor). A genuinely up-down-asymmetric free-boundary case matches an educational_VMEC reference for the same perturbed mgrid (`aspect`, `volume`, `Rmajor_p`, `Aminor_p`, `betatotal`, `rbtor`, `volavgB` within `1e-4`), with clearly non-zero antisymmetric Fourier content; its force residual settles near the `1e-8` tolerance rather than strictly below it, which does not move the converged equilibrium.
- New `vmec_test` cases cover the degenerate reductions (axisymmetric, 3D, and 3D free-boundary) and both educational_VMEC matches; `surface_geometry_test` adds the asymmetric inverse-DFT and derived-quantity checks. The opt, asan, and ubsan suites pass.

## Note

The non-stellarator-symmetric surface-geometry inverse DFT is also proposed on its own in #533; this branch contains the full lasym path, including that piece.


src/vmecpp/cpp/vmecpp/vmec/vmec/vmec_test.cc
```diff
@@ -171,6 +171,289 @@ TEST(TestVmec, CheckInMemoryMgrid) {
...
+  const auto asymmetric_output = vmecpp::run(asym_indata);
+  ASSERT_TRUE(asymmetric_output.ok());
+
+  // CompareWOut cannot be used directly because it requires identical lasym
+  // flags, so compare the converged physics: the key scalar quantities and the
+  // symmetric Fourier coefficients must match, and the antisymmetric
+  // coefficients of the lasym run must stay at zero.
+  const auto& sym = symmetric_output->wout;
+  const auto& asym = asymmetric_output->wout;
+  const double tol = 1.0e-9;
```
jurasic-pf: For a stellarator symmetric equilibrium i would expect numerically exactly the same results, to machine precision. Do we understand where non zero asym terms enter the picture/where numerical truncation changes the result?

CharlesCNorton: It reduces to machine precision; the gap was output normalization, not the equilibrium. The lasym `tmult *= 2.0` in `ComputeWOutFileContents` (the TODO citing PARVMEC #21) double-counts, since the Nyquist transform sums the reduced poloidal range for both parities, so `bmnc`/`gmnc`/`bsub*`/`bsup*`/currents came out 2x while geometry and scalars were fine. Dropping it (9c49fbc) makes the zero-asymmetry solovev match the symmetric run to ~1e-14 everywhere; I also tightened this test to assert the scalars and the symmetric and antisymmetric Nyquist arrays.

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec_test.cc
```diff
@@ -171,6 +171,322 @@ TEST(TestVmec, CheckInMemoryMgrid) {
...
+  const std::string filename = "vmecpp/test_data/up_down_asym.json";
+  absl::StatusOr<std::string> indata_json = ReadFile(filename);
+  ASSERT_TRUE(indata_json.ok());
+
+  absl::StatusOr<VmecINDATA> indata = VmecINDATA::FromJson(*indata_json);
+  ASSERT_TRUE(indata.ok());
+  ASSERT_TRUE(indata->lasym);
+
+  // VMEC++ finds the correct asymmetric equilibrium (its MHD energy and axis
+  // match VMEC 8.52), but the asymmetric force residual limit-cycles near
```
jurasic-pf: @jons-pf Could you look over this test?
I'm a little bit concerned that it might be masking a geniuent implementation error behind the loose tolerances.

@CharlesCNorton possibly we need to dump some more intermediate quantities along the convergence path from an educational vmec run to assess correctness

CharlesCNorton: I built educational_VMEC, ran `up_down_asym` through it, and dumped vmecpp's converged equilibrium against the reference quantity by quantity (not just the seven scalars). Bottom line: the equilibrium is correct, the loose scalar tolerance is not hiding a near-miss, and the one large discrepancy is on the reference side.

**Convergence.** The current branch drives this case to ftol `1e-11` (fsqr 9.5e-12 / fsqz 5.6e-12 / fsql 3.0e-13, niter 594), bit-identical across 1/2/4/8/32 threads (same niter, same volume, same b0 to all printed digits). The `return_outputs_even_if_not_converged = true` and the "limit-cycles near ~1e-9" comment are stale against this head; I will drop both.

**Scalars vs educational_VMEC** (the seven the test checks land far inside the 1e-4 tolerance):

| quantity | vmecpp | educational | rel |
|---|---:|---:|---:|
| aspect | 10.1 | 10.1 | 1e-15 |
| volume | 43.0630579 | 43.0630579 | 5e-16 |
| Rmajor_p | 6.06 | 6.06 | 1e-15 |
| Aminor_p | 0.60 | 0.60 | 2e-16 |
| rbtor | 31.958414 | 31.958251 | 5e-6 |
| rbtor0 | 32.071230 | 32.071235 | 1e-7 |
| volavgB | 5.2965403 | 5.2965399 | 8e-8 |

**Asymmetric geometry** (what the scalars are blind to): symmetric and antisymmetric Fourier geometry both agree to the convergence-ball floor (~1e-3 absolute on the R scale; educational takes 1215 iters to our 594, so the two land at slightly different points inside the 1e-11 ball). The asymmetry is genuinely present and matches: the magnetic axis sits off the midplane at Zaxis = 0.0790 (educational 0.0783), and rmns/zmnc are nonzero with the same mode structure.

**The one large diff is a clean factor of 2 on the field/current output spectra** (b0, bmnc, gmnc, bsub\*/bsup\*, buco, bvco, jcuru, jcurv). This is the lasym Nyquist output normalization: VMEC 8.52 emits the doubled spectra and this branch corrects them (the `tmult` factor is 0.5 as of 9c49fbc, and the degenerate test confirms that reproduces the symmetric-run spectra to 1e-10). Not an error in this branch. The physics settles which side is right:

- vmecpp b0 = 5.3092 = rbtor0 / Raxis = 32.0712 / 6.0407, the on-axis field. educational b0 = 10.62, which is 2x its *own* rbtor0 / Raxis (5.310), so the reference's b0 is internally inconsistent by exactly that factor.
- wb agrees to 1.6e-7, so the physical field energy is identical between the two runs; the 2x lives only in the output Nyquist transform, not in the equilibrium.

So there is no masked physics error, but the test is weaker than it should be. I will tighten it: drop the accept-unconverged flag (it converges to ftol now), pull the scalar tolerances down to where they actually land, and add direct assertions on the antisymmetric content (nonzero rmns/zmnc, off-midplane axis) so the asymmetric physics is checked head-on rather than inferred from symmetric-dominated scalars. The Nyquist field arrays I will assert against the half-corrected educational values, since the raw reference is doubled.

@jons-pf flagging you on the b0 / Nyquist normalization point specifically, since you know the VMEC output conventions here.

tests/test_lasym.py
```diff
@@ -0,0 +1,114 @@
...
+        "zaxis_c": zeros1d.copy(),
+    }
+    update.update(overrides)
+    return vmec_input.model_copy(update=update)
+
+
+def _assert_same_physics(ref, test, vol_rtol, beta_atol, iota_atol):
+    assert abs(test.volume_p - ref.volume_p) <= vol_rtol * abs(ref.volume_p)
+    assert abs(test.betatotal - ref.betatotal) <= beta_atol
+    assert np.max(np.abs(np.asarray(test.iotaf) - np.asarray(ref.iotaf))) <= iota_atol
```
jurasic-pf, line 63: This test is insufficient to confirm physics are unchanged.
We should really rotate all Fourier output quantities into the new frame and compare the rotated coefficients, to get full confidence.

jurasic-pf (commented):
The changes look correct, but I don't think it's safe to merge yet, until it's exhaustively validated.
I know Jonathan has implemented such a helper to rotate VMEC like fourier coefficients in the past for a different project, I'll try to find and port it over, so we can have full confidence in the lasym logic.

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec_test.cc
jurasic-pf: This is very good verification, Thanks!

jurasic-pf (approved):

jurasic-pf (approved):

PR #540, opened 2026-05-29
ideal_mhd_model: implement LFORBAL (m=1,n=0 force-balance modification)

CharlesCNorton:
## What `lforbal` does

With `lforbal = true`, the flux-averaged radial force balance evolves the m=1, n=0 R and Z components in place of their variational MHD forces. This is a non-variational modification of those two force components that drives the flux-surface-averaged residual `<EQUIF>` toward zero, so it changes the converged equilibrium (unlike an output-only diagnostic). `lforbal = false` leaves the iteration bit-for-bit unchanged.

## Implementation (`ideal_mhd_model`)

Ported from VMEC 8.52 / educational_VMEC, gated on `lforbal`:

- `computePreconditioningMatrix` accumulates the flux-averaged weight `temp` from the m=1 trig weights `cos01`/`sin01` and forms the scale factor `eqfactor = axd(m=1) * hs^2 / temp` (`precondn`).
- `updateRadialPreconditioner` builds the m=1,n=0 factors `rzu_fac`, `rru_fac` (each scaled by `sqrt(s)`, then halved) with `frcc_fac = 1/rzu_fac` and `fzsc_fac = -1/rru_fac` (`bcovar`).
- `forcesToFourier`, after the forward DFT, replaces the m=1,n=0 forces with `frcc = rzu_fac*(c*<EQUIF> + d)` and `fzsc = rru_fac*(c*<EQUIF> - d)`, where `d = frcc_fac*frcc + fzsc_fac*fzsc` and `c = nscale(0)` (`tomnsps`).

The default `lforbal = false` path is untouched.

## Validation

educational_VMEC had `lforbal` stripped from its compute path, so I reinstated it (`precondn` / `bcovar` / `tomnsp`), confirmed `lforbal = false` reproduces the solovev free-boundary baseline, and generated a converged `lforbal = true` golden. The new `SolovevFreeBoundaryLforbal.MatchesEducationalVmecGolden` test runs the axisymmetric solovev free-boundary case to convergence and matches the golden: flux-surface geometry, magnetic field, and integrated scalars to `5e-5`; the current-density profiles (the most edge-sensitive quantity) to `2e-3`. The worst normalized deviation anywhere is `7e-6`, in `jcurv`. The existing `lforbal = false` free-boundary regression continues to pass, so the gating is bit-clean.

## Base and lasym

This builds on the axisymmetric (nzeta=1) free-boundary support, which the solovev free-boundary case requires.

The non-stellarator-symmetric (`lasym = true`) case uses the same symmetric m=1,n=0 force balance, with the m=1 trig weights computed on the full theta grid. It is validated on the `implement-lforbal-lasym` branch (stacked on the lasym free-boundary work in #538): the asymmetric cth-like free-boundary case converges with lforbal on and off, lforbal drives the converged equilibrium to a different state, and `lforbal = false` is left bit-identical. That path folds into this change once the axisymmetric and lasym free-boundary branches are integrated; the lforbal code is identical on both.


jurasic-pf (approved):

jurasic-pf:
It's correct in principle but I'm not aware of any usage of this flag nowadays. The default in VMEC was changed many years ago, because empirically lforbal=false led to better convergence afaik.

I'd keep it open for now, if there's any user who would like that feature we can merge.

CharlesCNorton:
@jurasic-pf on the convergence point — I ran 40 configurations (13 cases plus sweeps in beta, radial resolution, tolerance, mpol, time step, iota and multigrid depth, fixed and free boundary). Zero lforbal-only failures; 30 converged in fewer iterations with it on, 5 in more, median −1.6%. The one real cost is solovev free-boundary multigrid at +34%, and that's the coarse-to-fine restart — each grid alone is faster.

Against exact on-axis iota from 7 published near-axis configs (pyQSC, nfp 2–4), lforbal is closer at 14 of 19 points, 6 of 7 at the smallest radius. Larger for strongly shaped cases: −18.6% on the `near_axis_iota_nfp4` iota profile, −33% on axis.

jurasic-pf (approved):
Thanks! That info does change the picture a bit, then it provides value as is

PR #541, opened 2026-05-29
Implement LFORBAL for the lasym free-boundary path

CharlesCNorton:
## lforbal for the non-stellarator-symmetric (lasym) free-boundary path

The lasym counterpart to #540, which implements and golden-validates lforbal for the axisymmetric (`lasym = false`) free-boundary case. This applies the same lforbal m=1, n=0 flux-averaged force balance through the lasym free-boundary code path, building on the lasym free-boundary support in #538.

The two free-boundary capabilities currently live on separate branches (nzeta=1 axisymmetric vs lasym), so lforbal is validated on each: the solovev nzeta=1 case in #540, and the asymmetric cth-like case here. The lforbal code is identical; the branches converge once the free-boundary work is integrated.

## Implementation

Same as #540 (`computePreconditioningMatrix` eqfactor, `updateRadialPreconditioner` factors, `forcesToFourier` m=1,n=0 force replacement), gated on `lforbal`. educational_VMEC applies lforbal only to the symmetric (`frcc`/`fzsc`) m=1 components, so the only lasym-specific addition is computing the m=1 trig weights `cos01`/`sin01` on the full theta grid (`nThetaEff == nThetaEven`) rather than the reduced symmetric grid.

## Validation

- `lforbal = false` leaves every existing case bit-identical (the full `output_quantities_test` suite passes).
- `CthLikeFreeBoundaryAsymLforbal` solves the asymmetric cth-like free-boundary case with lforbal off and on: both converge (FSQ < 1e-8), and lforbal drives the converged equilibrium to a different state (max |R,Z| change 1.1e-4), exercising the full lasym path (full-grid `cos01`/`sin01` + asymmetric DFT).
- The numerical force-balance physics is golden-validated against educational_VMEC in #540 (`lasym = false`, 7e-6 worst normalized deviation).

Depends on #538.


CharlesCNorton:
Consolidating into #540, which carries the golden validation against educational_VMEC (solovev free-boundary, lforbal=true reference matched to 7e-6). The lforbal code is identical here; this PR existed only because the lasym free-boundary support (#538) and the nzeta=1 axisymmetric support (#536) live on separate branches, so neither base runs both a symmetric nzeta=1 case and an asymmetric one. The lasym path (full-grid m=1 trig weights) and its regression test remain on the implement-lforbal-lasym branch and fold into #540 once the free-boundary branches are integrated.

PR #544, opened 2026-05-30
Python-side resolution continuation

CharlesCNorton:
## Summary

Adds Python-side resolution continuation.

`vmecpp.run_continuation` solves a schedule of increasing `(ns, mpol, ntor)`
resolutions, hot-restarting each step from the previous one. This covers the
classic radial `ns_array` multi-grid and adds Fourier continuation in `mpol` /
`ntor`, which had no first-class API before.

## Python API

- `vmecpp.interpolate_solution(source, target_input)` interpolates a converged
  `VmecOutput` onto the resolution of `target_input`: radial interpolation in
  `sqrt(s)` (with the odd-`m` axis handling the solver uses internally -- divide
  by `sqrt(s)`, extrapolate to the axis, zero the axis afterwards) and Fourier
  zero-padding or truncation. The result is a `restart_from` guess for the next
  step.
- `vmecpp.run_continuation(input, ns_array=..., mpol_array=..., ntor_array=...)`
  runs a whole schedule in one call. With only `ns_array` it performs the radial
  continuation; `mpol_array` / `ntor_array` add Fourier continuation.

This is a pure Python layer on top of `vmecpp.run`: each step solves a single
resolution and hot-restarts from the previous step's solution. The C++ solver is
unchanged.

## Validation

`tests/test_continuation.py` (8 tests):

- The generated state mode table matches the `xm` / `xn` VMEC++ writes out.
- Interpolation: identity is a no-op; radial up-sampling preserves the boundary
  surface; Fourier padding adds zero modes and truncation drops them.
- A single-resolution continuation is bit-for-bit identical to a direct
  single-grid `run`.
- A multi-grid continuation reaches the same equilibrium as the C++ multi-grid:
  bit-identical plasma volume, converged force balance, and geometry agreeing at
  the convergence level, with the agreement tightening as `ftol` tightens.
- Fourier continuation (ramping `mpol`) converges to the same equilibrium as a
  direct full-`mpol` solve.

## Example and docs

`examples/fourier_resolution_increase.py` is rewritten on the new API, replacing
~250 lines of hand-rolled mode-matching and interpolation. On
`cth_like_fixed_bdy`, ramping `mpol` 5 -> 9 -> 13 reaches the `mpol=13`
equilibrium in about 78% of the iterations of a direct `mpol=13` solve (same
equilibrium to ~3e-6); the advantage grows with the target resolution. This
resolves the `FIXME` / `TODO` in that example, which noted that the wout-based
interpolation did not match the native multi-grid and suggested turning that
comparison into a test -- it is now `test_continuation.py`. A "Resolution
continuation" section is added to the examples documentation.


jurasic-pf (commented):
Do you have any measurement on the performance impact of this change (if we were to default to it)? I'm guessing it's quite low, and we could replace the C++ interpolation logic with this entirely

CharlesCNorton:
> Do you have any measurement on the performance impact...

On `cth_like_fixed_bdy` with a radial multi-grid `ns_array=[15, 31, 51]` (mpol=5, ntor=4, ftol 1e-20), median of 5 runs on an i9-12900H:

- C++ native multi-grid: 6.224 s
- `run_continuation` (interpolation in Python): 6.457 s
- overhead: +0.23 s, about +3.7%, with bit-identical plasma volume

`interpolate_solution` itself is ~4.5 ms (ns 15 to 51), so the 0.23 s is the per-step Python/C++ round-trip (three `run()` calls instead of one solve), not the interpolation. The round-trip is a fixed per-step cost, so it's a smaller fraction on harder or longer solves.

Thanks for taking a look.

jurasic-pf (approved):

docs/examples_overview.md
```diff
@@ -14,3 +14,37 @@ The `examples/` directory contains short scripts demonstrating typical tasks:
...
+    mpol_array=[5, 9, 13],
+    ntor_array=[4, 4, 4],
+)
+```
+
+[`fourier_resolution_increase.py`](../examples/fourier_resolution_increase.py)
+ramps `mpol` from 5 to 13 on the `cth_like_fixed_bdy` case. It reaches the same
+equilibrium as a direct `mpol=13` solve (geometry agreeing to about 3e-6) in
+roughly 78% of the iterations, and the advantage grows with the target
+resolution.
```
jurasic-pf:
```suggestion
```

jurasic-pf: I don't think the test fits in the overview documentation.

CharlesCNorton:
will look at the ci failures and fix on Monday

CharlesCNorton:
@jurasic-pf done

src/vmecpp/_continuation.py
```diff
@@ -0,0 +1,359 @@
...
+    return out
+
+
+def _remap_axis(values: np.ndarray, ntor_new: int) -> np.ndarray:
+    """Truncate or zero-pad an axis Fourier array ``[ntor + 1]`` to ``ntor_new``."""
+    values = np.asarray(values, dtype=float)
+    out = np.zeros(ntor_new + 1, dtype=float)
+    keep = min(values.shape[0], ntor_new + 1)
+    out[:keep] = values[:keep]
+    return out
```
jurasic-pf: i believe this is functionally the same as `_util.right_pad()`

tests/test_continuation.py
```diff
@@ -0,0 +1,236 @@
...
+
+    assert int(continued.wout.mpol) == mpol_final
+    assert _final_force_residual(continued) < 1e-5
+    assert continued.wout.volume == pytest.approx(cma_direct.wout.volume, rel=1e-4)
+    np.testing.assert_allclose(
+        np.asarray(continued.wout.rmnc),
+        np.asarray(cma_direct.wout.rmnc),
+        atol=4e-3,
+        rtol=1e-2,
+    )
```
jurasic-pf, line 238: surprisingly loose tolerance given the solutions should be the same (just took a different path to get there)

CharlesCNorton: At this ftol the `rmnc` modes differ by up to 2.9e-3 absolute, so `atol=rtol=1e-3` fails 79/3009 elements. The agreement tightens with ftol (`test_continuation_agreement_tightens_with_ftol`).

PR #545, opened 2026-05-30
Alternative initial-guess methods for flux-surface geometry (Zeno, map2disc)

CharlesCNorton:
Adds two alternative constructions for the interior flux-surface initial guess, as alternatives to VMEC++'s default linear axis-to-boundary interpolation (`m=0` linear in `s`, higher poloidal modes as `s^(m/2)`). Each builds the full interior coordinate map up front and returns a `VmecOutput` usable as `restart_from`; both are exposed as `vmecpp.zeno_guess` and `vmecpp.map2disc_guess`.

- **Zeno / Tecchiolli** (arXiv:2405.08173): minimizes a geometric action (squared Jacobian weighted by `1/(2s)`, plus a radial-straightness term) over a Fourier-Zernike basis with the boundary held fixed. Following the paper, this empirically drives the Jacobian toward a single sign to correct overlapping guesses but does not guarantee positivity. Self-contained (numpy, optional SciPy minimizer).
- **map2disc** (Babin et al., PPCF 2025): conformally maps each cross-section to the unit disc. The map is a C2 diffeomorphism, so the interior Jacobian is positive by construction (nested surfaces). Requires the optional `map2disc` package (https://gitlab.mpcdf.mpg.de/gvec-group/map2disc), which is not on PyPI; the relevant tests skip when it is absent.

**Scope.** VMEC++'s radial multi-grid and the `s^(m/2)` interior scaling already produce a valid nested coarse-grid geometry for typical boundaries, so these constructions are most useful for single-resolution solves and guess-sensitivity studies rather than as a general convergence fix.

**Validation** (`tests/test_initial_guess.py`, `examples/alternative_initial_guess.py`):
- Hot-restarting VMEC++ from `zeno_guess` converges to the same equilibrium (cth_like) as the default guess.
- The map2disc core produces strictly nested surfaces: the enclosed cross-section area is monotonic in `s`, matching the conformal-disc expectation that area is proportional to `rho^2 = s`.


jurasic-pf (commented):
Cool!
I was holding off on this project, because there's some more work required to establish which of these methods actually benefit VMEC, and how.

Before merging, this needs to be clarified imo. The questions  are:
- Does iteration count decrease/increase?
- Which fraction of runtime do we now spend on initialization? (+how does this depend on number of surfaces)
- How does this compare for more/less shaped equilibria? (e.g. a varied dataset from constellaration or QUASR for testing purposes)
- Does this initialization strategy resolve cases where VMEC would previously fail with e.g. `JACOBIAN_75_TIMES_BAD`

I've briefly looked into it in https://github.com/proximafusion/vmecpp/pull/554 and found very mixed results

CharlesCNorton:
Ran it on 50 constellaration boundaries (single-resolution, vacuum fixed-boundary), default vs map2disc, plus zeno spot-checks and a tokamak-to-W7X built-in spread.

- Iterations: map2disc and zeno match the default's count (e.g. 2292, 3281); on the built-ins zeno is within a few iterations, map2disc within ~10%.
- Init cost vs ns: map2disc's guess is ~0.1 to 2.5s and flat in ns; zeno's is 95 to 100% of runtime and grows fast with ns (90s at cma ns=51, 516s at W7X ns=9, vs 1 to 2s solves).
- Shaping (aspect 5.9 to 10.3, elongation 3 to 9, triangularity -0.6 to +0.2): same count as the default, except ~8 to 10% fewer for map2disc on the low-aspect built-ins and ~35% more on W7X.
- JACOBIAN_75_TIMES_BAD: 2 of 50 fail under the default (one is that error); map2disc fails the same way on both.

jurasic-pf:
>   Shaping (aspect 5.9 to 10.3, elongation 3 to 9, triangularity -0.6 to +0.2): same count as the default, except ~8 to 10% fewer for map2disc on the low-aspect built-ins and ~35% more on W7X.

If this implementation of zenos method is on the same order as the solve, then it doesn't seem practical for our usecase (still very useful information!). map2disc  could be a good candidate for a better initialization.

CharlesCNorton:
Correction: in that sweep the restart guess wasn't reaching the solver, so the map2disc numbers were just the default run. Injected correctly, map2disc is slightly slower on an easy case (cth_like ns=25: 448 vs 439) and still hits `JACOBIAN_75` despite a nested interior (min scaled Jacobian 0.27). Happy to keep exploring in whatever direction you think is most useful.

PR #546, opened 2026-05-30
Expose the forward model and drive the equilibrium iteration from Python

CharlesCNorton:
## Problem

VMEC++'s equilibrium iteration (the time stepping toward force balance) lives entirely inside the C++ `Vmec` core (`Vmec::SolveEquilibrium` / `SolveEquilibriumLoop` / `Evolve` in `vmec/vmec/vmec.cc`). From Python only `vmecpp.run` is reachable, and it runs the whole multi-grid solve start to finish; the per-step forward model (flux-surface geometry to MHD forces) and the time-stepping logic are not exposed. Developing an alternative iteration scheme therefore means editing the C++ core, which is slow and error-prone.

VMEC++ also implements only the Fortran VMEC 8.52 time-step control. `IterationStyle` (in `vmec_indata.h`) already names a `PARVMEC` (VMEC2000 9.0) variant, but it was parsed and stored and never wired into the iteration, and its enum value was never even bound to Python. The two controls converge for different equilibria: some configurations converge under VMEC 8.52 (VMEC++) but not (PAR)VMEC 9.0, and the reverse.

## Change

Expose the forward model and the iteration primitives to Python, port the iteration out of C++, and build the PARVMEC variant, a common-ground style, and a boundary-optimization coupling on top of the Python loop.

**C++ (`pybind11/pybind_vmec.cc`, `CMakeLists.txt`).** A single-threaded `VmecModel` binding over `Vmec`:

- `create(indata, ns)` initializes one radial resolution (the inner solve VMEC++ runs at one multi-grid step).
- `evaluate(iter1, iter2)` runs `IdealMhdModel::update` (geometry to forces). `(iter2 - iter1)` is passed through because it drives the preconditioner-update schedule (`shouldUpdateRadialPreconditioner`).
- `perform_time_step` (the Garabedian update `v = fac*(b1*v + dt*f); x += dt*v`), `save_backup` / `restore_backup`, `reinitialize` (the axis reguess), flat `get_state` / `set_state` / `get_forces`, the invariant and preconditioned residuals, `iteration_style`, and a reference `solve()` (the C++ inner loop, kept for verification).
- The forward model runs inside a single-thread `omp parallel` region so the `omp single` / `barrier` directives inside `update` have the team context they get in `SolveEquilibrium`; the module is compiled with OpenMP for this. The previously-unbound `PARVMEC` `IterationStyle` value is now bound.

**Python (`_iteration.py`, exposed as `vmecpp.solve_equilibrium` and `vmecpp.iterate`).** A one-to-one port of `SolveEquilibriumLoop` and `Evolve`: the 1/tau damping average (`kNDamp = 10`), the `dtau` / `fac` / `b1` time step, the time-step control, the bad-Jacobian axis reguess, the ijacob 25/50/75 escalation, and the `ftol` convergence test. The expensive forward model and per-coefficient arithmetic stay in C++; the iteration logic is Python.

## Iteration styles

All three styles share the forward model (which does not depend on `iteration_style`), the damping, the axis reguess, and the ijacob escalation. They differ only in how a growing residual is handled (the `TimeStepControl` policy):

| style | revert the geometry when | step reduction | counts toward give-up |
|---|---|---|---|
| `vmec_8_52` | preconditioned residual > 100x its running minimum | `delt *= 0.9`, plus a slow-progress branch (`delt /= 1.03`) | yes (`ijacob`) |
| `parvmec` | preconditioned **or** invariant residual > 1e4x its minimum | `delt /= 1.03` | no |
| `robust` | either residual > 1e3x its minimum | `delt /= 1.03`, plus 8.52's slow-progress safeguard | no |

`vmec_8_52` reproduces the C++ control exactly. `parvmec` follows PARVMEC's `evolve.f` / `TimeStepControl`: a 100x-more-permissive leash, tracked on both the preconditioned and the invariant residual, a gentle revert, and no escalation toward the give-up counter. `robust` is the common-ground scheme: PARVMEC's permissive, non-escalating revert at a moderate leash, kept from wandering by 8.52's slow-progress safeguard. `iteration_style` selects between them (`vmec_8_52` and `parvmec` are the C++ enum values; `robust` is a Python-only scheme on the same forward model).

## Verification

`tests/test_iteration.py`, `tests/test_optimization.py`:

- **Bit-for-bit reproduction.** From an identical initialized state, the Python loop and the C++ `SolveEquilibrium` reach the same equilibrium on solovev (2D, ns=15) and cth_like_fixed_bdy (3D, ns=25): identical iteration counts, converged decision vectors agreeing to `5.2e-14` and `3.2e-13`, and identical invariant force-residual traces (max absolute difference `8e-16` and `1e-14` over the whole run). The only non-bit-exact operation is the 10-element damping-average sum (numpy vs Eigen, about 1 ULP).
- **Recovery from a bad initial Jacobian.** cma's linear initial guess overlaps at a single resolution (its magnetic axis defaults to zero), so the loop recomputes the axis, re-initializes the radial state from it (a clean forward model), and converges to force balance.
- **Styles agree absent a restart.** On cth_like the three styles advance through bit-identical states: the policies differ only once a restart fires.
- **Style divergence.** cma solved single-resolution at ns=72 trips `vmec_8_52`'s 100x leash 21 times (3280 iterations); `parvmec` never reverts (1658 iterations) and `robust` reverts twice (1774); all three converge to the same equilibrium. Across cma's fragile resolutions (ns 68/72/88) `vmec_8_52` thrashes (about 20 restarts) where `parvmec` and `robust` take 0 and 2, and on every bundled fixed-boundary equilibrium the alternative styles converge wherever `vmec_8_52` does.
- **Boundary-optimization coupling.** `optimize_boundary`, varying one boundary coefficient, recovers the boundary whose converged equilibrium matches a target stored MHD energy.

`examples/python_iteration.py` drives the loop through the public API; `examples/python_boundary_optimization.py` drives the optimization coupling.


jurasic-pf (changes requested):
Very cool! This is quite close to what we were planning indeed. Also impressive that you already included both variants of the iteration logic from different vmec versions!
Do you have any measurements on how the performance of the python iteration compares to the C++ flow control on different resolutions?

As mentioned in another PR, I don't see optimization specific functinality in scope for `vmecpp`. vmecpp/_optimize is a very cool usage example, but also seems entirely possible to build the same as a user, without modifying vmecpp source code.

Please check the benchmark CI job, it seems there was a regression in import time because of the `scipy` import.

CharlesCNorton:
Rescoped to the in-scope half: forward-model exposure and the Python iteration (VMEC-8.52, PARVMEC, and a common-ground "robust" style). Dropped `optimize_boundary` as out-of-scope; that also removes the scipy import that regressed the benchmark, so `import vmecpp` is scipy-free again and the benchmark job is green.

Python loop vs C++ flow control on the same forward model (CMA, single resolution, `VmecModel.solve()` vs `solve_equilibrium`, both run to convergence):

| ns | C++ | Python | ratio |
|---|---|---|---|
| 15 | 62.5 ms | 65.0 ms | 1.04x |
| 25 | 107.6 ms | 116.4 ms | 1.08x |
| 51 | 242.1 ms | 276.0 ms | 1.14x |

4 to 14% overhead, growing slowly with resolution, since the forward model stays in C++ and only the time-step and restart control are in Python. clang-tidy findings are fixed too (the `VmecModel` bindings are const now).

Thanks for the review.

jurasic-pf:
Ok - then the slowdown is likely too much for switching the default, but it's super valuable as a research platform for a developing better descent algorithm.
Thanks!

tests/test_iteration.py
```diff
@@ -0,0 +1,133 @@
...
+    result = vmecpp.solve_equilibrium(model)
+
+    assert result.converged
+    assert not result.failed
+    assert result.axis_reguesses == 0
+
+    cpp_state = np.asarray(reference.get_state())
+    py_state = np.asarray(model.get_state())
+    assert cpp_state.shape == py_state.shape
+    assert np.max(np.abs(cpp_state - py_state)) < 1.0e-9
```
jurasic-pf: nit: `np.testing.assert_allclose` checks shapes and values, and produces a much nicer readable output on error.

jurasic-pf:
Looks really good, thanks!
I'd like to have @jons-pf take a second look on this PR

jurasic-pf (approved):

tests/test_iteration.py
CharlesCNorton: Done, switched to `np.testing.assert_allclose`.

PR #547, opened 2026-05-30
objectives: stellarator design objective library with Boozer quasi-symmetry

CharlesCNorton:
## Summary

Adds `vmecpp.objectives`, a library of stellarator-optimization figures of merit
evaluated on a converged VMEC++ equilibrium: aspect ratio, the rotational-transform
profile, the magnetic well, the mirror ratio, and the Boozer quasi-symmetry
residual, together with the Boozer-coordinate transform the residual is built on.
Each objective reads a converged `wout` and returns a scalar (or a profile); none
of them touch the solver, so they compose with any equilibrium VMEC++ produces.

The intent is to compute and test optimization targets in Python while the C++
core stays the forward model. Where an objective has a SIMSOPT counterpart the two
agree to numerical precision, so these are usable without a SIMSOPT/VMEC2000 build.

## Objectives

`vmecpp.objectives` exposes:

| Function | Definition |
|---|---|
| `aspect_ratio(eq)` | R0 / a, as reported by VMEC (`wout.aspect`) |
| `iota_profile(eq)` | rotational transform on the full-grid surfaces, returned as `(s, iota)` |
| `iota_profile_residual(eq, target)` | mean-square deviation of iota from a target profile (callable or array) |
| `magnetic_well(eq)` | vacuum well `(V'(s->0) - V'(s->1)) / V'(s->0)`, with `V'` from `dV/dPsi` |
| `mirror_ratio(eq)` | `(Bmax - Bmin) / (Bmax + Bmin)` over a flux surface |
| `field_strength_on_surface(eq)` | `\|B\|(theta, zeta)` on a flux surface, from `bmnc` |
| `boozer_spectrum(eq)` | `\|B\|` Boozer harmonics on a surface |
| `boozer_roundtrip_residual(eq)` | self-consistency residual of the Boozer transform |
| `quasisymmetry_residual(eq, helicity_m, helicity_n)` | symmetry-breaking Boozer-harmonic fraction |

Each accepts a `VmecOutput` or its `VmecWOut`.

## Boozer transform

The quasi-symmetry residual is defined on `|B|` expressed in Boozer angles, so the
PR carries a Boozer-coordinate transform built from the `wout` Fourier data
(Hirshman-Whitson straight-field-line construction):

- `I = Bsub_theta_00` and `G = Bsub_zeta_00` are the enclosed currents;
- the periodic part of the stream function has sine harmonics `w_mn = bsubumnc_mn / m = -bsubvmnc_mn / n`;
- `nu = (w - I*lambda) / (G + iota*I)` gives the Boozer angles `theta_B = theta + lambda + iota*nu`, `zeta_B = zeta + nu`;
- `|B|` is resampled onto Boozer harmonics with the Jacobian weight `J = (1 + lambda_theta)(1 + nu_zeta) + (iota - lambda_zeta)*nu_theta`.

The forward integral runs over a single field period `[0, 2*pi/nfp)`, following the
booz_xform convention. Over one period the toroidal frequency is the mode number
`n` rather than `n*nfp`, so `n_zeta ~ 4*n_boozer` resolves the spectrum with no
toroidal aliasing. The Boozer poloidal content spreads well past the VMEC band, so
`m_boozer` defaults to several times `mpol`.

## Validation

`tests/test_objectives.py` runs every objective on a converged `cth_like_fixed_bdy`
equilibrium and compares against SIMSOPT or against the transform's own
self-consistency:

| Check | Result |
|---|---|
| `aspect_ratio` vs `Vmec.aspect()` | exact |
| iota axis / edge / mean vs `Vmec.iota_axis / iota_edge / mean_iota` | < 1e-10 |
| `magnetic_well` vs `Vmec.vacuum_well()` | < 1e-10 |
| `field_strength_on_surface` vs `vmec_compute_geometry().modB` | exact |
| `mirror_ratio` vs the same `\|B\|` field | < 1e-8 |
| Boozer round-trip at `(m, n)_boozer = (48, 20)` | 2.6e-10 |
| Boozer round-trip, `(12, 8)` to `(36, 16)` | 1.3e-4 to 1.1e-8 |
| `quasisymmetry_residual`, stability across truncations | 0.0405 (constant) |

Aspect ratio, iota, the magnetic well, and the mirror-ratio `|B|` field are pinned
to SIMSOPT. The Boozer transform is validated intrinsically: it reconstructs `|B|`
at the VMEC grid points from its own Boozer spectrum, and the residual falls to
machine precision as the mode count grows. The quasi-symmetry residual is constant
under refinement of the Boozer truncation and is cross-checked against simsopt's
real-space `QuasisymmetryRatioResidual`: cth_like is not quasi-axisymmetric, and
both report a finite QA-breaking residual. booz_xform is the conventional external
reference for Boozer spectra; a direct spectral diff against it is a sensible
additional check on top of the self-consistency and SIMSOPT cross-checks here.

## Example

`examples/python_objectives.py` solves the fixed-boundary stellarator and prints
every objective:

```
aspect ratio            : 5.4858
iota on axis / edge     : 1.2522 / 0.8253
magnetic well           : -1.4750e-02
mirror ratio            : 0.4365
quasi-axisymmetry resid : 4.0503e-02
Boozer self-consistency : 3.12e-07
```

### DIFF
--- a/examples/python_objectives.py
+++ b/examples/python_objectives.py
@@ -1,0 +1,38 @@
+# SPDX-FileCopyrightText: 2024-present Proxima Fusion GmbH <info@proximafusion.com>
+#
+# SPDX-License-Identifier: MIT
+"""Stellarator-optimization objectives on a converged VMEC++ equilibrium.
+
+``vmecpp.objectives`` computes standard design targets from a converged
+equilibrium: the aspect ratio, the rotational-transform profile, the magnetic
+well, the mirror ratio, and the Boozer quasi-symmetry residual (with the
+Boozer-coordinate transform it requires). The aspect ratio, iota, magnetic well,
+and mirror-ratio field match SIMSOPT to numerical precision; the Boozer transform
+is validated by self-consistency. Here we solve a fixed-boundary stellarator and
+report each objective.
+"""
+
+from pathlib import Path
+
+import vmecpp
+from vmecpp import objectives
+
+vmec_output = vmecpp.run(
+    vmecpp.VmecInput.from_file(Path("examples") / "data" / "cth_like_fixed_bdy.json"),
+    max_threads=1,
+    verbose=False,
+)
+
+_, iota = objectives.iota_profile(vmec_output)
+
+print(f"aspect ratio            : {objectives.aspect_ratio(vmec_output):.4f}")
+print(f"iota on axis / edge     : {iota[0]:.4f} / {iota[-1]:.4f}")
+print(f"magnetic well           : {objectives.magnetic_well(vmec_output):+.4e}")
+print(f"mirror ratio            : {objectives.mirror_ratio(vmec_output):.4f}")
+print(
+    "quasi-axisymmetry resid : "
+    f"{objectives.quasisymmetry_residual(vmec_output, helicity_m=1, helicity_n=0):.4e}"
+)
+print(
+    f"Boozer self-consistency : {objectives.boozer_roundtrip_residual(vmec_output):.2e}"
+)
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -19,6 +19,7 @@ import netCDF4
 import numpy as np
 import pydantic
 
+from vmecpp import _objectives as objectives
 from vmecpp import _util
 from vmecpp._free_boundary import (
     MagneticFieldResponseTable,
@@ -2048,4 +2049,5 @@ __all__ = [  # noqa: RUF022
     "MagneticFieldResponseTable",
     "FreeBoundaryMethod",
     "set_profile",
+    "objectives",
 ]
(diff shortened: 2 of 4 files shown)


jurasic-pf (changes requested):
Thanks for the contribution, but implementing optimization specific output quantities is out of scope for `vmecpp`.

`vmecpp` is really intended as a low fidelity, minimum dependency ideal mhd equilibrium solver, which can be wrapped and used by downstream libraries like `simsopt` for optimization, but is not intended to replace any functionality implemented there (aside from the minimal subset in the compatibility module).

There's a medium term effort to decouple `simsopt` and `vmecpp` with the `simsopt` maintainers.

CharlesCNorton:
Makes sense, thanks for the quick read. That fits the simsopt decoupling direction, so I'll close this. Appreciate it.

PR #560, opened 2026-06-10
iteration: fix post-reguess lambda scaling divergence, add Python multigrid driver

### LINKED ISSUE #555: Python `_iteration` and C++ iteration logic diverge for some equilibria
The initial push for bringing the iteration logic into python still has a couple bugs that need ironing out:
- For some cases that involve reset logic the iteration takes a different path than in the C++ implementation. E.g. for CMA with a single multigrid iteration and ns=72 the python implementation get's stuck for a few resets, while the C++ iteration continues after a single `HUGE_INITIAL_FORCES` reset : 

Image

- Multigrid logic isn't implemented yet

CharlesCNorton:
Fixes #555.

## Root cause of the Python/C++ divergence

`VmecConstants::rmsPhiP` is accumulated by `RadialProfiles::evalRadialProfiles`, and `Vmec::run` calls `constants_.reset()` before every `InitializeRadial` (vmec.cc:302). The `VmecModel` bindings `Reinitialize()` (the axis-reguess path) and `RefineTo()` (the multigrid step) re-ran `InitializeRadial` without that reset, so each re-initialization doubled `rmsPhiP` and rescaled `lamscale` by sqrt(2).

That rescales the entire lambda sector. The preconditioned residual `fsql1` carries the factor directly (it has no normalization that cancels a `lamscale` change), while the invariant `fsql` stays unchanged, so the damping average sees a different `fsq1` and the post-reguess trajectory splits from the C++ inner solve from the second step onward. The converged physics is unaffected (`lamscale` is an internal scaling), which is why this never showed in convergence tests, only in step-for-step traces. On `li383_low_res` at ns=31 the signature is clean: bit-identical `fsqr`/`fsqr1`/`fsqz1` with `fsql1` off by exactly 2.0 at the first post-reguess evaluation.

A secondary, much smaller divergence channel was the damping average itself: numpy's pairwise summation associates differently from Eigen's sequential `VectorXd::sum()`, and the last-bit difference grows through long chaotic transients. The Python loop now sums the ring buffer left to right, matching Eigen bit for bit.

## Changes

- `pybind_vmec.cc`: `constants_.reset()` before `InitializeRadial` in `Reinitialize()` and `RefineTo()`, matching `Vmec::run`.
- `_iteration.py`: sequential left-to-right damping-average sum.
- `_iteration.py` / `__init__.py`: `solve_multigrid()` — drives the full coarse-to-fine `ns_array` ramp from Python, mirroring the `Vmec::run` multi-grid loop (skip-coarser `ns_min` rule, per-stage ftol/niter, `refine_to` interpolation between stages, C++ failure semantics). This covers the "multigrid logic isn't implemented yet" part of the issue.
- `tests/test_iteration.py`:
  - regression test for the lambda scaling under repeated `reinitialize()`,
  - multigrid parity test: `solve_multigrid` on cma's own two-stage input (ns_array `[25, 51]`) against the full `vmecpp.run`, comparing the concatenated force-residual and restart-reason traces step for step,
  - `li383_low_res` added to the restart-path matrix (cold-start reguess with a clean descent — a sharp probe of the post-reguess state),
  - whole-trace residual tolerance tightened from `rtol=0.5` to `rtol=1e-3`, with the first 50 iterations at `rtol=1e-9`.

## Validation

Full-length (to convergence) Python-vs-C++ parity sweep, comparing iteration counts, restart-reason traces, and residual traces:

| case | ns | before | after |
|---|---|---|---|
| cma | 72 | 1729 vs 1728 iters, residuals split at it 5 | 1729 vs 1729, traces match |
| cma | 15 | 1539 vs 1540 | 1539 vs 1539, traces match |
| cma | 31 | 1655 vs 1649 | 1655 vs 1655, traces match |
| li383_low_res | 31 | 476 vs 511, residuals split at it 2 | 476 vs 476, traces match |
| solovev 15, cth_like 25, circular_tokamak 25, w7x 15/25 | | already matching | still matching |

Restart-reason traces are identical in every case; residual traces are tight for the whole run (the only remaining non-bit-exact effect is last-bit noise through the longest chaotic transients, staying below 1e-3 relative).

`tests/test_iteration.py`: 12 passed. Full Python suite: 144 passed, 4 skipped (the `cth_like_free_bdy` wout-reference failure pre-exists at `main` in this environment and is unrelated).

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -30,6 +30,7 @@ from vmecpp._iteration import (
     RestartReason,
     iterate,
     solve_equilibrium,
+    solve_multigrid,
 )
 from vmecpp._pydantic_numpy import BaseModelWithNumpy
 from vmecpp.cpp import _vmecpp  # type: ignore # bindings to the C++ core
@@ -2422,6 +2423,7 @@ __all__ = [  # noqa: RUF022
     "set_profile",
     "iterate",
     "solve_equilibrium",
+    "solve_multigrid",
     "IterationResult",
     "IterationState",
 ]
--- a/src/vmecpp/_iteration.py
+++ b/src/vmecpp/_iteration.py
@@ -299,7 +299,16 @@ def solve_equilibrium(
                 trace_l.append(fsql)
                 trace_rr.append(int(rr))
 
-                otav = inv_tau.sum() / _NDAMP
+                # Sequential left-to-right sum, matching Eigen's scalar
+                # VectorXd::sum() bit for bit. numpy's pairwise summation
+                # associates differently; the last-bit difference in the
+                # damping average amplifies through the early force
+                # transient on cold-start cases and shifts the convergence
+                # iteration against the C++ loop.
+                otav = 0.0
+                for _v in inv_tau:
+                    otav += float(_v)
+                otav /= _NDAMP
                 dtau = delt0r * otav / 2.0
                 model.perform_time_step(1.0 / (1.0 + dtau), 1.0 - dtau, delt0r)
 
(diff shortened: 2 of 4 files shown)


jurasic-pf (approved):
Thanks! Impressive tracking down of this iteration divergence

What's your impression, is it ready to run on a larger set of boundaries, from constellaration or quasar and compare to the C++ path as a more thorough verification? Or are there any remaining discrepancies you're aware of

CharlesCNorton:
@jurasic-pf Ran 250 constellaration boundaries through both paths and the traces hold bit-for-bit through the first 50 iterations and stay within 1e-4 the whole way, so it's ready for the full sweep and I'm not seeing any remaining discrepancies.

jurasic-pf:
> @jurasic-pf Ran 250 constellaration boundaries through both paths and the traces hold bit-for-bit through the first 50 iterations and stay within 1e-4 the whole way, so it's ready for the full sweep and I'm not seeing any remaining discrepancies.

Amazing, thanks!
I didn't know you had already done thayt.

PR #561, opened 2026-06-10
CUDA port of the iteration body (opt-in via VMECPP_USE_CUDA)

CharlesCNorton:
# CUDA port of the iteration body (opt-in via VMECPP_USE_CUDA)

## Summary

This PR adds an opt-in CUDA execution path for the iteration body,
covering fixed-boundary and free-boundary configurations. Built with `-DVMECPP_USE_CUDA=ON`, the per-iteration chain (forward
toroidal transform, Jacobian and metric elements, MHD forces, inverse
transform, preconditioning, residuals, time step) executes on the GPU with
device-resident state; `vmecpp.run` and `vmec_standalone` use it
transparently for supported inputs. A batched mode solves N equilibria
concurrently inside one CUDA context, exposed to Python as
`run_batched_gpu`: a broadcast call returns the single converged result of
the seed configuration, and a distinct call returns one `OutputQuantities`
per configuration, derived directly from the batched run's flushed device
state; each carries that configuration's own iteration count and final
force residuals, snapshotted while it was still active so a converged
configuration reports its own residual rather than a post-mask zero. Every
environment variable the binding sets is restored when the call returns.

The option defaults to OFF. Without it, the build requires no CUDA toolkit
and contains no CUDA reference.

The build runs on Linux and natively on Windows with the MSVC host
compiler (system dependencies from vcpkg, the Ninja generator, and a
CUDA compiler launcher that adapts the MSVC command line). All
Windows-specific handling is guarded under `_WIN32`/`MSVC`, so the Linux
build is byte-identical; the native Windows module builds and runs `cma`
on an RTX 6000 Ada to the canonical result. See the Windows section of
the docs for the exact steps.

Documentation: `docs/cuda_acceleration.md` (scope, build, correctness
contract, batched mode, and a reference of every runtime control).
Regression coverage: `tests/test_cuda_gpu.py`, opt-in via
`VMECPP_TEST_CUDA=1` since it requires a CUDA build and a GPU.

## Scope

Supported: fixed-boundary and free-boundary, stellarator-symmetric
(`lasym = false`) configurations, both axisymmetric (`ntor = 0`, the
degenerate `nZeta = 1` case) and three-dimensional (`ntor >= 1`), on a
single radial rank, in both the single and the batched execution modes.
Free-boundary runs keep the NESTOR vacuum solve on the host with
per-iteration bridges to the device-resident iteration state; the
segment and whole-iteration CUDA graphs are disabled for them, since
the edge-force kernel toggles with the vacuum pressure state. Sync
elision covers free-boundary runs: iterations run live until the
vacuum contribution is fully active, then the scalar sync sites elide
while the vacuum block keeps its per-iteration cadence (the NESTOR
response must track the boundary every iteration; a boundary-cadence
vacuum solve was measured non-convergent and rejected). Batched free-boundary runs carry one NESTOR
instance per configuration against one loaded mgrid (every input in the
batch must share `mgrid_file` and `extcur`; boundaries and plasma
profiles may differ in distinct mode); the vacuum activation, the
`nvacskip` cadence, and the soft-start restart are batch-wide decisions,
and the per-iteration vacuum solves serialize on the host across the
configurations, which bounds the batched free-boundary speedup below
the fixed-boundary one.

Enforced at runtime under the CUDA build:

- `Vmec::run` rejects `lasym = true` inputs with
  `absl::UnimplementedError` before any iteration work, along with batched
  free-boundary inputs whose vacuum solver is not NESTOR.
- Inputs with `nThetaReduced > 256` are rejected with
  `absl::UnimplementedError` (a poloidal-scatter kernel-layout limit). The
  radial resolution is accepted up to `CudaMaxRadialResolution()`: the
  parallel cyclic reduction solver carries one thread per radial row through
  1024 rows, and the block-Thomas solver holds its elimination ratios in
  dynamic shared memory sized to `ns` above that, so the device's opt-in
  shared-memory capacity sets the ceiling; an `ns` beyond it is rejected.
- An estimate of the run's device-memory footprint is checked against the
  device's free memory before the first allocation; an oversized batch is
  rejected with `absl::ResourceExhaustedError` instead of failing at an
  allocation deep inside the iteration body.
- CUDA and cuFFT failures surface as recoverable error statuses (a Python
  `RuntimeError` through the bindings) instead of terminating the process;
  the next run re-stages the device state from scratch.
- The radial domain is not partitioned across OpenMP threads;
  `num_threads` is clamped to 1 (host threading offers no benefit when the
  iteration body runs on the device).

## Changes when the option is OFF

The CPU algorithm and its behavior are unchanged. Two things in shared
code are new:

1. `FlowControl` gains per-configuration state vectors (sized 1 and inert
   outside the batched CUDA mode) and a typed `RestartReason` enum with
   documentation of the Fortran `irst` correspondence.
2. `ideal_mhd_model.cc` gains phase-timing instrumentation
   (`phase_timer.h`): an RAII scope timer around each iteration phase that
   writes a summary to a file under `/tmp` (path configurable via
   `VMECPP_PHASE_TIMING_PATH`). The instrumentation is compiled only under
   `VMECPP_USE_CUDA`; in CPU-only builds the macro expands to nothing.

The remaining CPU-side changes are `#ifdef VMECPP_USE_CUDA` dispatch
sites and are compiled out.

## Architecture

- All iteration state (spectral coefficients, geometry, forces,
  preconditioner matrices, velocity and position of the time integrator)
  persists in device buffers across iterations. Host transfers are limited
  to a small number of scalars per iteration (Jacobian tau extrema,
  residual triples, plasma volume), plus consolidated flushes at
  preconditioner-update and end-of-run boundaries.
- The toroidal transforms run as batched cuFFT plans; the poloidal
  transforms are fused into fill/scatter kernels on either side. The
  dealias and scatter kernels warp-pack `32 / nThetaReduced` zeta planes
  onto each 32-lane warp (default on), filling the lanes the one-plane
  layout left idle when the reduced poloidal extent is 16.
- Three sync-free segments of the iteration body are captured as CUDA
  graphs and replayed per iteration (default on, each individually
  defeatable by environment variable).
- The batched mode sizes every per-configuration buffer by
  `VMECPP_N_CONFIG_MAX` and carries a configuration axis on each kernel
  grid. Per-configuration residual evolution, convergence gating,
  time-step control, bad-Jacobian restart, and an active-configuration
  kernel mask let configurations converge independently; a configuration
  that exceeds its iteration cap is marked timed out without failing the
  batch.

## Correctness

Contract, verified on fixed-boundary stellarator configurations through
the full multigrid ramp:

- `aspect_ratio` matches the CPU result bit for bit.
- Volume-averaged and field-line-derived output quantities match within
  1e-5 to 1e-3 relative. The origin is floating-point summation-order
  differences in reductions; each such reduction is documented at its
  kernel.
- Iteration counts, restart-reason sequences, and multigrid stage
  transitions match the CPU trajectory. On inputs that trigger mid-run
  bad-Jacobian restarts, the timing of restart events is sensitive to
  the documented drift, so iteration counts there match approximately
  rather than exactly (W7-X converges in 3146 iterations against the
  CPU's 2954, to the same equilibrium).
- On free-boundary runs the boundary itself responds to the vacuum
  field through the drift-sensitive trajectory, so every converged
  output carries the drift family there; the bit-exact `aspect_ratio`
  clause applies to fixed-boundary runs.

Reductions that feed the time-step controller and the convergence gate use
serial or order-preserving device implementations and match the CPU bit
for bit per call. Reductions covered by the drift tolerance (for example
the differential volume under the fused atomicAdd kernel) are documented
as such.

## Validation

CPU-only build (option off):

- `vmec_standalone` and the Python path reproduce upstream `main` results
  on `cma`; the free-boundary case `cth_like_free_bdy` runs unchanged.
- No instrumentation side effects: a run writes no timing file.

CUDA build, development machine (RTX 6000 Ada Generation, Linux under
WSL2):

- Production stellarator boundary (mpol = 10, ntor = 10, nfp = 3,
  ns = [25, 51, 99]) through the full multigrid ramp: `aspect_ratio`
  bit-identical to the CPU result, derived metrics inside the documented
  drift family.
- `cma`: CPU and CUDA volume and aspect ratio equal to 12 significant
  digits.
- Axisymmetric boundaries (`solovev`, `circular_tokamak`) match the CPU
  build bit for bit, single-grid and through the multigrid ramp; `lasym`
  inputs are rejected with `UnimplementedError`; `vmec_standalone` with
  default thread settings runs single-rank.

CUDA build, second machine (RTX 3070 Ti Laptop, sm_86, 8 GB, CUDA
toolkit 12.0 with gcc-12 host compiler, native Ubuntu 24.04):

- Same battery: convergence trajectories identical to the development
  machine, outputs equal to 12 significant digits.
- The `cma` wout volume and aspect ratio are bit-identical to the
  development machine (hex-level comparison) for the single run, the
  N = 4 broadcast, sync elision, and the whole-iteration graph; the
  duplicated-boundary distinct run is bit-identical across its
  configurations; the direct per-configuration output derivation
  returns the same twelve significant digits for both boundaries; the
  spectral-width progress column matches to the printed digit.
- The fixed-boundary W7-X example (`examples/data/w7x.json`, mpol = 12,
  ntor = 12, nfp = 5, single stage at ns = 99) converges to the CPU
  build's equilibrium: the MHD energy agrees to all seven printed
  digits and the converged axis position and beta match. This input
  exercises the mid-run bad-Jacobian recovery path, which no other
  validation input triggers.
- Hardening coverage on the same battery: the shape guards, the memory
  pre-flight (rejection plus same-process recovery), device selection
  with out-of-range fallback, mixed configuration counts across runs in
  one process, and broadcast at N = 2 and N = 4 bit-identical to the
  single run, including through the iteration-1 magnetic-axis recovery,
  which re-stages the device position state from the recomputed axis.

CUDA build, third machine (NVIDIA A100-SXM4-80GB, sm_80, CUDA toolkit
12.4, Ubuntu 22.04 container):

- Built with the default `CMAKE_CUDA_ARCHITECTURES` list, so this run
  executes the native sm_80 code path of the shipped fat binary on the
  datacenter Ampere die.
- On an 8x A100 node, eight processes pinned one per device each
  completed a 16-configuration batch concurrently, with all eight
  results bit-identical across devices.
- Full battery: `cma` wout volume and aspect ratio bit-identical to both
  other machines (hex-level comparison) for the single run, the N = 4
  broadcast, sync elision, and the whole-iteration graph; the
  duplicated-boundary distinct run is bit-identical across its
  configurations; the direct per-configuration output derivation returns
  the same twelve significant digits for both boundaries; the
  spectral-width progress column matches both machines to the printed
  digit.
- The opt-in test suite passes on a single A100 (eleven cases,
  including both free-boundary forms, the W7-X bad-Jacobian recovery,
  and the tensor-core scatter at W7-X's full poloidal spectrum), and
  the experiment-gate survey passes for all twenty-three gates with
  drift values matching the other machines.

Batched mode, development machine and RTX 3070 Ti:

- `run_batched_gpu` driven directly on `cma`: broadcast at N = 4 and
  distinct at N = 2 complete with the CUDA graphs active and return
  per-configuration outputs equal to the single-configuration run.
- Broadcast at N = 4 is bit-identical to the single-configuration run
  (hex-level wout comparison) with the per-configuration time-step
  controller active.
- Distinct mode with two different boundaries converges each
  configuration to its own equilibrium (the converged spectra differ by
  exactly the imposed boundary ratio); with duplicated boundaries the
  per-configuration states are bit-identical at every probed iteration
  of both multigrid stages.
- Two consecutive distinct-mode `run_batched_gpu` calls in one process
  return identical per-configuration outputs; the second call restages
  its inputs from clean device state and its multigrid stage transition
  replays the first call's bit for bit.
- Free-boundary: the cth-like case converges to the CPU equilibrium in
  both its single-stage and multigrid forms (volume within 1.4e-5 and
  1.0e-5 relative respectively, aspect and MHD energy within 1e-6),
  with the NESTOR consistency checks reporting the same toroidal
  current and poloidal field as the CPU run.
- `tests/test_cuda_gpu.py` pins the above as a repeatable suite: the
  canonical single-configuration result, repeated-run bitwise
  determinism, broadcast returning a single output through the
  iteration-1 axis recovery, bitwise-identical consecutive distinct
  calls, mixed configuration counts in one process, the shape guards
  and memory pre-flight with same-process recovery, sync-elision and
  whole-iteration-graph bit-identity, the Ozaki-3 scatter family
  result, the staged int8 limb descent, both cth-like free-boundary
  forms (single-stage and the multigrid stage transition), both forms
  again under sync elision with a broadcast batch, the
  free-boundary wout fields against a committed reference, a
  free-boundary broadcast against the single run, a two-boundary
  distinct free-boundary batch, a fixed-boundary run followed by a
  free-boundary run in one process, hot restart, the axisymmetric
  `solovev` and `circular_tokamak` boundaries in single, multigrid,
  distinct-batch, and broadcast forms, W7-X convergence through the
  mid-run bad-Jacobian recovery, and W7-X under the wmma
  scatter at its full poloidal spectrum (the W7-X cases are
  additionally gated behind `VMECPP_TEST_CUDA_SLOW=1`).

## Performance

Measured on an RTX 6000 Ada Generation (i9-13900KF host, Linux under
WSL2; native Linux carries less launch overhead), on a production
stellarator boundary with mpol = 10, ntor = 10, nfp = 3, multigrid
ns = [25, 51, 99], approximately 11,500 iterations to convergence:

- Single configuration: about 10 s wall for the full run, roughly
  0.9 ms per iteration averaged over the ramp.
- Batched distinct mode at N = 16 to 32: about 4.3 s per equilibrium.
- For reference, the same equilibrium evaluation through our CPU pipeline
  takes about 54 s per equilibrium on the same machine.

Re-measured at the current head on the same card with cases that ship in
the repository: the W7-X fixed-boundary example (`examples/data/w7x.json`,
mpol = 12, ntor = 12, nfp = 5) converges in about 3.7 s single
configuration, and `cma` runs in 0.29 s single configuration against
0.030 s per equilibrium in a distinct batch of 32, falling to 0.022 s
per equilibrium in a batch of 192 (4.2 s for all 192). The int8-scatter
and state-backup kernel measurements on this card are recorded in
`docs/cuda_acceleration.md`.

The per-kernel timing harness (`VMECPP_KERNEL_TIMING=1`) and the phase
timer give the breakdown if you want to reproduce these numbers on other
hardware.

## Runtime controls

`docs/cuda_acceleration.md` documents every environment variable. Three
classes:

1. Production-path switches, default ON, each falling back to a slower
   equivalent path (kernel fusions, the dealias and scatter warp-packs,
   pair-coarsened variants, parallel residual reduction, the three CUDA
   graphs, the device-side convergence gate, the per-configuration
   time-step controller). These exist so a
   misbehavior on untested hardware can be bisected per subsystem.
2. Opt-in throughput modes, default OFF (K-window sync elision,
   deferred-sync residuals, a whole-iteration CUDA graph under elision).
3. A mixed-precision and alternate-FFT experiments annex, default OFF,
   with measured outcomes recorded per gate (FP32/TF32 scatter variants
   with double-single compensation, Ozaki-slice multiplications on FP32
   and on int8 tensor cores with exact integer accumulation, a
   Carson-Higham iterative refinement on the RZ tridiagonal solve,
   hand-coded radix-8x3 DFTs). None of these are on the production path.
   They are included because the negative results are documented in code
   and prevent re-exploration; if you prefer the annex out of this PR, I
   will split it into a follow-up.

`VMECPP_CUDA_DEVICE` selects the device ordinal for the process
(default 0; multi-GPU batches run one process per device). The
run-scoped controls (`VMECPP_N_CONFIG_MAX`, `VMECPP_SYNC_ELIDE`,
`VMECPP_PER_CFG_NITER_CAP`, and the batch upscale gates) are re-read at
the start of every run, so successive runs in one process can carry
different settings.

## Known limitations

- The `<M>` (volume-averaged spectral width) column and the `specw`
  output evaluate the current iteration state under the CUDA build; the
  CPU path evaluates the restart backup, which lags by one iteration, so
  the printed column can differ in the trailing digit at a given
  iteration.
- In distinct mode the per-configuration `OutputQuantities` derivation
  is the default (`VMECPP_PER_CFG_RECOMPUTE=0` opts out, falling back to
  a hot-restart reconstruction per configuration); the subprocess
  recompute (`recompute_outputs_from_spectra`) and the raw converged
  spectra (`return_spectra=True`) remain available.

## Suggested review order

| Step | Files | Content |
|---|---|---|
| 1 | `docs/cuda_acceleration.md` | Scope, contract, controls |
| 2 | `CMakeLists.txt`, `ideal_mhd_model/CMakeLists.txt` | Build gating |
| 3 | `fft_toroidal_cuda.h` | Interface; per-function contracts |
| 4 | `vmec.cc`, `ideal_mhd_model.cc` | Dispatch sites, controller integration, scope guards |
| 5 | `flow_control.h/.cc` | Per-configuration controller state |
| 6 | `pybind_vmec.cc` | `run_batched_gpu`, `recompute_outputs_from_spectra` |
| 7 | `fft_toroidal_cuda_common.cuh`, `fft_toroidal_cuda_kernels.cu`, `fft_toroidal_cuda_state.cu`, `fft_toroidal_cuda_transform.cu`, `fft_toroidal_cuda_physics.cu`, `fft_toroidal_cuda_control.cu`, `fft_toroidal_cuda_io.cu` | Kernels, wrappers, device state (kernel-by-kernel documentation headers; one topical translation unit each) |
| 8 | `tests/test_cuda_gpu.py` | Opt-in GPU regression suite |
| 9 | `phase_timer.h` | Instrumentation |


jurasic-pf (commented):
Wow, I'm excited to try it out!
As a heads up, given the huge PR it'll take me a while to review :)

CharlesCNorton:
> Wow, I'm excited to try it out! As a heads up, given the huge PR it'll take me a while to review :)

It's ok! I have some fixes to make anyways in the meantime.

jons-pf:
WOW! Like @jurasic-pf I'll need a bit of time to digest this (as for your other marvelous PRs 😬 👍 ), but this looks really promising. Thanks already now in advance for proposing these!

src/vmecpp/cpp/vmecpp/common/flow_control/flow_control.cc
```diff
@@ -22,38 +35,58 @@ RestartReason RestartReasonFromInt(int restart_reason) {
...
     return omp_get_max_threads();
 #endif  // _OPENMP
-    // Default to 1 thread if OpenMP is not available
     return 1;
   }
   CHECK_GT(max_threads.value(), 0)
       << "The number of threads must be >=1. "
          "To automatically use all available threads, pass std::nullopt";
 #ifdef _OPENMP
-  // Size the thread pool immediately so spin-waiting threads are not created
```
jurasic-pf, line 41: please keep these comments

CharlesCNorton: @jurasic-pf ok all addressed

PR #562, opened 2026-06-11
Handle jaxtyping anonymous-dimension API drift in wout serialization

jurasic-pf:
`VmecWOut` NetCDF serialization assumed `jaxtyping._array_types._AnonymousDim` exists. With newer jaxtyping releases, that symbol may be absent (replaced by `_anonymous_dim`), causing runtime `AttributeError` in `test_init` wout IO/reference paths.

- **Compatibility fix: dim marker resolution**
  - Resolve jaxtyping dimension marker types via guarded lookup on `jt._array_types`:
    - named: `_NamedDim`
    - anonymous: `_AnonymousDim` **or** fallback to `type(_anonymous_dim)` when only the instance-style API is present.
- **Runtime behavior: robust dimension-name inference**
  - Keep existing inference behavior, but branch on resolved types instead of hard-coding a single private symbol.
  - If a marker type is unavailable, logic degrades safely to existing default dimension naming.
- **Code organization**
  - Compute marker-type resolution once before field iteration, then reuse in per-field shape inference.

```python
array_types = getattr(jt, "_array_types", None)
named_dim_type = getattr(array_types, "_NamedDim", None) if array_types is not None else None
anonymous_dim_type = getattr(array_types, "_AnonymousDim", None) if array_types is not None else None
if anonymous_dim_type is None and array_types is not None:
    anonymous_dim_instance = getattr(array_types, "_anonymous_dim", None)
    if anonymous_dim_instance is not None:
        anonymous_dim_type = type(anonymous_dim_instance)
```

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -1064,31 +1064,31 @@ class VmecWOut(BaseModelWithNumpy):
     piota_type: ProfileType
     """Parametrization of iota profile (copied from input)."""
 
-    am: jt.Float[np.ndarray, "preset"]
+    am: jt.Float[np.ndarray, "_preset"]
     """Mass/pressure profile coefficients (copied from input)."""
 
-    ac: jt.Float[np.ndarray, "preset"]
+    ac: jt.Float[np.ndarray, "_preset"]
     """Enclosed toroidal current profile coefficients (copied from input)."""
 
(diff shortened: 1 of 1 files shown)


jons-pf (approved):

jurasic-pf:
@copilot the jaxtyping annotations are used to infer the netcdf dimension names. leading underscores should be trimmed away

PR #564, opened 2026-06-14
build+ci: abseil commit pin for Clang>=21, VMEC2000-from-source, benchmark fork guard

krystophny:
Build and CI prerequisites for the Clang/Enzyme autodiff stack, bundled
because each piece is needed for this PR's own checks to pass.

## Changes
- **abseil**: pin `FetchContent` to commit `255c84d` (the 20260107.1 LTS
  commit). Older abseil fails to compile under Clang >= 21 (the Enzyme
  build) on `absl::Nonnull` SFINAE in `absl/strings/ascii.cc`. Pinned to
  the commit, not the moveable tag.
- **VMEC2000 compat CI**: build VMEC2000 from the pinned source commit
  `728af8b` and cache the wheel. The prebuilt `vmec-0.0.6` cp310 wheel was
  f90wrapped against numpy 1.x and fails to import under numpy 2, so the
  compatibility test could never run. Drops the unused FFTW/HDF5 dev
  packages (VMEC2000 links neither; HDF5 arrives via netcdf).
- **benchmarks CI**: skip the result upload on fork PRs (their token is
  read-only, otherwise the upload hits "Resource not accessible").
- **compat test**: skip vmecpp-only INDATA fields with no VMEC2000
  counterpart.

Supersedes #583, whose CI fixes are included here.

## Verification
Before: the compat job failed to import VMEC2000 under numpy 2; the
benchmark job failed with "Resource not accessible by integration".
After: all checks green (25/25).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -53,12 +53,27 @@ jobs:
       - name: Also install VMEC2000 (only on Ubuntu 22.04)
         if: ${{ matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.10' }}
         run: |
-          # mpi4py is needed for VMEC2000
-          sudo apt-get install -y libopenmpi-dev
-          python -m pip install mpi4py
-          # custom wheel for VMEC2000, needed for some VMEC++/VMEC2000 compatibility tests
-          # NOTE: this wheel is only guaranteed to work on Ubuntu 22.04
-          python -m pip install vmec@https://anaconda.org/eguiraud-pf/vmec/0.0.6/download/vmec-0.0.6-cp310-cp310-linux_x86_64.whl
+          # Build VMEC2000 from source instead of the old prebuilt wheel. The
+          # pinned vmec-0.0.6 cp310 wheel was f90wrapped against numpy 1.x and
+          # fails to import under numpy 2 (f90wrap array-interface break), so the
+          # compatibility test could never run. Building from source with current
+          # f90wrap produces numpy-2-compatible bindings. Recipe mirrors SIMSOPT's
+          # own CI (hiddenSymmetries/VMEC2000 cmake/machines/ubuntu.json).
+          sudo apt-get install -y libopenmpi-dev libnetcdff-dev libscalapack-mpi-dev \
+            libhdf5-dev libhdf5-serial-dev libfftw3-dev libopenblas-dev ninja-build
+          # Build VMEC2000 with the SYSTEM cmake + ninja. Do not pip-install the
+          # cmake/ninja wheels: they would shadow the system cmake that
+          # scikit-build-core's editable vmecpp rebuild records, breaking its
+          # verify_globs step in the editable matrix job.
+          python -m pip install mpi4py numpy scikit-build f90wrap setuptools wheel
+          git clone --depth 1 https://github.com/hiddenSymmetries/VMEC2000.git /tmp/VMEC2000
+          cd /tmp/VMEC2000
+          cp cmake/machines/ubuntu.json cmake_config_file.json
+          LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu \
+            python -m pip install -v --no-build-isolation .
+          cd -
+          # fail loudly here if the binding is still broken, not in the test step
+          python -c "import vmec; print('VMEC2000 import OK')"
       - name: Install package
         run: |
           # on Ubuntu we would not need this, but on MacOS we need to point CMake to gfortran-14 and gcc-14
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yaml
jurasic-pf, line 1: I'd prefer to leave this CI workflow as is, we pre-built the wheel to have a pinned, known good version and faster CI (dont pay build time for vmec2000)

tests/test_simsopt_compat.py
jurasic-pf, line 1: not entirely clear why this os needed

.github/workflows/tests.yaml
krystophny: I understand. See #583 - this is the actual PR that I cherry picked into the others to get tests green. I just did this because the PR checks were failing. Do you have a clue? Then we avoid it.

tests/test_simsopt_compat.py
krystophny: This was also leading to CI failure when opening PR from my fork. It tries to update some `benchmark.json` which is not possible if PR comes from other organization. The guard skips just the write-back on fork PRs, the benchmark itself still runs, and same-repo PRs (e.g. once this lands on main) write back as before.

krystophny: Sorry, I thought you meant the benchmark CI section. This test compares vmecpp `indata` against VMEC2000's namelist, common subset only. It already skips `return_outputs_even_if_not_converged`. It never ran before because the wheel import was broken. Now that VMEC2000 imports, it hits vmecpp-only fields VMEC2000 lacks, e.g. `free_boundary_method`, which throws `AttributeError`. The `hasattr` guard skips fields VMEC2000 doesn't have, generalizing that existing skip. Can switch to an explicit list if you prefer.

.github/workflows/tests.yaml
krystophny: It builds now the exact same commit as this `v0.0.6` and caches the built wheel in #583 . Let's get that one settled first and then continue with the remaining stack. Once you are happy with #583 I will once more rebase all the other PRs on main. 

(PS: sorry for the huge amount of PRs and probably full mailbox that it caused. I thought it was still better than one giant PR and it can be reviewed step by step. The actual diff of each PR on the stack is 50-500 lines each. I hope it is worth it, as Enzyme autodiff makes a HUGE difference in performance.)

jurasic-pf: I'm very keen to review the stack, now i think i understand the structure of how to do that better. Thenks for looking into it!

CMakeLists.txt
```diff
@@ -66,7 +66,10 @@ find_package(LAPACK REQUIRED)
 FetchContent_Declare(
   abseil-cpp
   GIT_REPOSITORY https://github.com/abseil/abseil-cpp.git
-  GIT_TAG 4447c7562e3bc702ade25105912dce503f0c4010
+  # 20260107.1 LTS: required for Clang >= 21, where the 2024 pin fails to
+  # compile (absl::Nonnull SFINAE in absl/strings/ascii.cc). Clang is the
+  # compiler used for the Enzyme autodiff build.
+  GIT_TAG 20260107.1
```
jurasic-pf:
```suggestion
  GIT_TAG b29e7f9d7f4016e668647e5c2f27a1dc1e0b5243
```
Please pin to fixed commit hashes to avoid supply chain attacks. tags can be changed a posterior

krystophny: Agreed on pinning to a commit hash. One heads-up: the suggested `b29e7f9d...` is the **20260526** LTS branch ("Apply LTS transformations for 20260526 LTS branch", #2063), not the 20260107.1 the comment documents and that the stack was built and tested against.

To keep the bump minimal I pinned the exact commit behind the 20260107.1 tag instead: `255c84dadd029fd8ad25c5efb5933e47beaa00c7` ("Abseil LTS branch, Jan 2026, Patch 1", #2007). Same supply-chain guarantee (immutable commit), same tested version.

Happy to bump to the 20260526 LTS if you'd rather track the newer one, but that's a separate version change that the whole stack rebuilds on, so I'd want CI to re-confirm it first.

tests/test_simsopt_compat.py
jurasic-pf: Thanks for the context, great that you found this issue! I've approved it in the corresponding PR

krystophny: ok I now combined the two PRs here since they are anyway both needed for the build and ci. could we keep the commits and not squash them, so the rebase chain will be easier for follow up prs? i set the first #565 to "ready" now to review.

krystophny: ...we can also squash merge but then we will not rebase but merge this into our other prs and squash-merge each

jurasic-pf (approved):

PR #565, opened 2026-06-14
enzyme: opt-in Clang/Enzyme build option + AD smoke test

krystophny:
> **Stacked PR — part 2/19 of the differentiable-VMEC++ series.** merge after #564 (build-clang21).

---

## What

Add an opt-in Clang/Enzyme build path and a toolchain smoke test:

- `VMECPP_ENABLE_ENZYME` CMake option, **off by default**. When on, it requires
  a Clang compiler and `-DVMECPP_ENZYME_PLUGIN=/path/to/ClangEnzyme-NN.so`, and
  builds the `enzyme_smoke` test target.
- `common/enzyme/enzyme.h`: thin declarations of the Enzyme autodiff intrinsics
  and the activity markers, with the allocation constraint that shapes every
  differentiable kernel in this stack documented in place.
- `common/enzyme/enzyme_smoke_test.cc`: differentiates a scalar objective over
  `Eigen::Map`'d caller buffers and checks reverse- and forward-mode gradients
  against the closed form and central finite differences.

## Why

Stacked on #5. Enzyme differentiates LLVM IR through a Clang plugin, so the
differentiable VMEC++ work needs a Clang build and a way to attach the plugin.
This patch adds that switch and a test that fails loudly if the plugin is not
attached or if Enzyme cannot differentiate the `Eigen::Map` buffer pattern the
later kernels depend on. It adds no production code; with the option off, the
build is byte-for-byte the previous one.

The smoke test also pins down the one hard Enzyme constraint for this codebase:
Enzyme's allocation analysis does not track Eigen's aligned allocator, so a
dynamic-size Eigen heap temporary crossing the differentiated call aborts with
"freeing without malloc". Differentiable kernels therefore operate on
caller-owned buffers via `Eigen::Map`. The test exercises exactly that pattern.

## Verification

Configure and build with Clang 21.1.8 and the ClangEnzyme-21 plugin:

```
-- Enzyme plugin: .../ClangEnzyme-21.so
cfg=0
build=0 (0 errors)
```

`ctest`:

```
    Start 1: enzyme_smoke
1/1 Test #1: enzyme_smoke .....................   Passed    0.00 sec
100% tests passed, 0 tests failed out of 1
```

Smoke test output (reverse and forward both exact, agree with finite
differences):

```
enzyme smoke test (n=8)
  max|reverse - analytic| = 0.000e+00
  max|forward - analytic| = 0.000e+00
  max|reverse - finite-diff| = 2.646e-08
PASS
```

No regression with the option off. A default configure (GCC, no Enzyme flags)
succeeds and emits no `enzyme_smoke` target:

```
cfg=0  enzyme target present? 0
```

Tracking: #587

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -36,7 +36,11 @@ jobs:
           pytest benchmarks/test_benchmarks.py --benchmark-json=benchmark_results.json
 
       - name: Compare benchmark result
-        if: github.event_name == 'pull_request'
+        # Skip the result upload/compare for fork PRs: their GITHUB_TOKEN is
+        # read-only, so comment-on-alert/auto-push hit 'Resource not accessible
+        # by integration'. The benchmarks still run above; only the write-back
+        # is skipped for forks.
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
         uses: benchmark-action/github-action-benchmark@v1.21.0
         with:
           tool: "pytest"
--- a/.github/workflows/test_enzyme.yaml
+++ b/.github/workflows/test_enzyme.yaml
@@ -1,0 +1,77 @@
+name: Enzyme AD smoke test
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+# Cancel currently running job if a new one comes along for the same branch or tag.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: ${{ github.ref_name != 'main' }}
+
+env:
+  # Clang/LLVM toolchain and the matching pinned Enzyme release. v0.0.264 is the
+  # version this stack is developed against; it builds ClangEnzyme-21 against
+  # LLVM 21.
+  LLVM_VERSION: "21"
+  ENZYME_REF: "v0.0.264"
+
+jobs:
+  enzyme-smoke:
+    name: Build ClangEnzyme and run the bazel enzyme smoke test
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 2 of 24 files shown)


jurasic-pf (approved):
Cool! I propose to review and approve the stack piece by piece and merge all at once, since including enzyme has no benefit unless we merge the rest.

krystophny:
good then i undraft the prs - I am very excited about this as this will ultimately beat every other VMEC including DESC.

jurasic-pf:
I'm also very excited about this! Getting the best of both worlds would be nice.

jurasic-pf:
Currently all the C++ tests are handled via bazel in this repository, I see you were setting up ctest for this one.

I guess the difficulty was that bazel is configured for gcc and not clang, certainly not with enzyme. If we can include Enzyme in the bazel test framework I'd prefer it, alternatively setting up a CI for cmake tests so this is tested is possible.

jurasic-pf:
https://github.com/EnzymeAD/Enzyme/blob/main/.github/workflows/ccpp.yml
How to set up enzyme CI

krystophny:
Moved the smoke test into bazel (`//vmecpp/common/enzyme:enzyme_smoke_test`, `manual`-tagged so the default gcc `//...` skips it; run with `--config=enzyme` + `--copt=-fplugin=<ClangEnzyme>`) and dropped the ctest path. Added `test_enzyme.yaml` per your ccpp.yml hint: it builds a pinned ClangEnzyme-21 from source and runs the bazel test. Green in CI. Thanks for the pointer!

jurasic-pf (approved):
Thanks, that's exactly what I meant!
And such a fast response!

krystophny:
> Thanks, that's exactly what I meant! And such a fast response!

We need to move fast to build a reactor ;-)

PR #566, opened 2026-06-14
ideal_mhd_model: make computeMHDForces allocation-free

### LINKED ISSUE #594: Assert the core-loop is allocation free
Let's make sure the hot-loops remain allocation free by adding this to unit tests. This should be doable with memchecker, valgrind or something. Uninitialized memory should be checked by adding an `msan` test (to `ubsan`, `asan` and `tsan`). It seems our current CI does not catch these cases yet (as https://github.com/proximafusion/vmecpp/commit/ab117cb23c72e537b233ebb59597598a600f5a3a @CharlesCNorton identified)

krystophny:
> **Stacked PR — part 3/19 of the differentiable-VMEC++ series.** based directly on `main`.

---

## What

Make `IdealMhdModel::computeMHDForces` allocation-free. The kernel allocated 17
dynamic `Eigen::VectorXd` per radial surface: the `_o` half-grid quantities
(`P_o`, `rup_o`, ..., `gbvbv_o`) and the surface averages (`P_avg`, `P_wavg`,
`gbubu_avg`, ...). These move to preallocated per-thread `ThreadLocalStorage`
scratch and are assigned in place; the radial loop now allocates nothing.

## Why

Two reasons:

1. **Performance**: the force kernel runs every iteration; allocating 17 vectors
   per surface per iteration is avoidable heap churn.
2. **Differentiability**: Enzyme cannot trace dynamic Eigen temporaries -- both
   forward and reverse mode abort on a dynamic-size Eigen expression that
   allocates (the "freeing without malloc" failure). Writing into preallocated
   storage via `Eigen::Map`/in-place assignment is the form Enzyme can
   differentiate. This is the allocation-free prerequisite for an exact autodiff
   Hessian-vector product of the force.

This is a pure refactor: the arithmetic is identical, only the storage of the
intermediates changes.

## Verification

`computeMHDForces` now contains zero `Eigen::VectorXd` locals or
`Eigen::VectorXd::Zero` constructions (allocation-free).

Bit-for-bit identical results, `vmec_standalone` final MHD energy:

```
                       before (main)     after
solovev.json           2.548352e+00      2.548352e+00
cth_like_fixed_bdy     5.057191e-02      5.057191e-02
```

Builds clean under GCC 16 and Clang 21; clang-format (file style) clean.

Tracking: #588

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -36,7 +36,11 @@ jobs:
           pytest benchmarks/test_benchmarks.py --benchmark-json=benchmark_results.json
 
       - name: Compare benchmark result
-        if: github.event_name == 'pull_request'
+        # Skip the result upload/compare for fork PRs: their GITHUB_TOKEN is
+        # read-only, so comment-on-alert/auto-push hit 'Resource not accessible
+        # by integration'. The benchmarks still run above; only the write-back
+        # is skipped for forks.
+        if: github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name == github.repository
         uses: benchmark-action/github-action-benchmark@v1.21.0
         with:
           tool: "pytest"
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -53,12 +53,27 @@ jobs:
       - name: Also install VMEC2000 (only on Ubuntu 22.04)
         if: ${{ matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.10' }}
         run: |
-          # mpi4py is needed for VMEC2000
-          sudo apt-get install -y libopenmpi-dev
-          python -m pip install mpi4py
-          # custom wheel for VMEC2000, needed for some VMEC++/VMEC2000 compatibility tests
-          # NOTE: this wheel is only guaranteed to work on Ubuntu 22.04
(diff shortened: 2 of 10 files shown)


jurasic-pf (commented):
The change looks good, but I'm a little bit surprised that the benchmarks seem to indicate a small regression in performance for cma and w7x

I think ths should ve clarified before merging

krystophny:
Fixed now. Profiled locally against `main`; cma and w7x had different causes.

cma (partial-DFT path): the block-scope `thread_local` scratch in the `ForcesToFourier3DSymmFastPoloidal` inner loop added a `__tls_get_addr` call plus an init guard per iteration. Hoisted it to function-scope scratch (still allocation-free, per-thread via the stack).

w7x (FFTX path): `perf diff` put the regression in `computeJacobian` and `computeMetricElements`, each about 2x slower. The shared kernels take raw `double*`, so without `__restrict` the compiler assumes aliasing and cannot vectorize; the Eigen expressions they replaced did. Marked the kernel buffers `__restrict`.

Wall-clock vs `main` (gcc Release, FFTX on, single thread; ratio stack/main, below 1 is faster):

| case | before | after |
|---|---|---|
| solovev | 0.92 | 0.92 |
| cma | 0.97 | 0.97 |
| cth_like | 0.96 | 0.91 |
| w7x | 1.10 | 0.99 |

All four now at or below `main`. Enzyme is unchanged: `jacobian_kernel_autodiff` and `local_force_hessian` pass, and the QS Gauss-Newton benchmark converges bit-identically.

Cherry-picked both fixes onto every downstream PR in the stack (#567 through #585).

jurasic-pf (approved):

PR #567, opened 2026-06-14
ideal_mhd_model: share the Jacobian kernel with exact autodiff (fwd vs rev)

krystophny:
> **Stacked PR — part 4/19 of the differentiable-VMEC++ series.** merge after #565 (enzyme-build-option).

---

## What

Extract the half-grid Jacobian arithmetic into a shared, allocation-free kernel
`ComputeHalfGridJacobian` (`ideal_mhd_model/jacobian_kernel.h`) over flat
buffers, and use it from **both** the solver (`IdealMhdModel::computeJacobian`)
and an Enzyme forward/reverse autodiff test. One implementation, no duplication.

## Why

This is the first kernel of the exact analytic+autodiff Hessian. The kernel maps
full-grid geometry to the half-grid `r12, ru12, zu12, rs, zs` and the Jacobian
`tau`. `tau` is nonlinear in the geometry, so its Jacobian is a building block of
the MHD force Hessian (chain rule composes the kernel Jacobians with the linear
spectral transforms to give the Hessian-vector product).

Writing it allocation-free over flat buffers is exactly the form Enzyme
differentiates -- both forward and reverse mode abort on dynamic Eigen
temporaries, and cannot dup member-struct objects. The solver keeps the kernel
plus the unchanged Jacobian-sign check.

## Verification

Production is bit-for-bit unchanged: `vmec_standalone` MHD energy on solovev
`2.548352e+00` and cth_like_fixed_bdy `5.057191e-02`, before and after.

Enzyme test (`ctest -R jacobian_kernel_autodiff`, with
`-DVMECPP_ENABLE_ENZYME=ON`) differentiates the shared kernel; for
`L = 0.5||outputs||^2`:

```
reverse dL.v vs finite-diff : 1.9e-9      (exact, FD-limited)
forward dL.v vs finite-diff : 1.9e-9
forward / reverse agreement : 3e-15       (machine precision)
performance: reverse ~15 us (full 3328-elt gradient) | forward ~15 us (one direction)
```

Reverse returns the whole gradient per pass (~3300x more efficient for a full
scalar gradient); forward is the cheaper primitive for a single
Jacobian/Hessian-vector product.

## Scope

First of the force-chain kernels. The remaining kernels (metric, B^contra,
B_cov, pressure/energy, MHD force) follow the same shared-kernel pattern; the
exact force Hessian-vector product then composes them with the linear transforms
applied analytically. Stacked on #6.

Tracking: #588


src/vmecpp/cpp/vmecpp/vmec/vmec/vmec_in_memory_mgrid_test.cc
jurasic-pf, line 1: I must admit that I don't understand why we need to increase tolerances when this should be a purely structural refactoring that does not change the semantics.

Yes, floating point noise can accumulate differently, depending on vectorization, but this is still surprising to me.

CMakeLists.txt
```diff
@@ -174,3 +174,43 @@ endif()
...
+  message(STATUS "Enzyme plugin: ${VMECPP_ENZYME_PLUGIN}")
+  enable_testing()
+  add_executable(enzyme_smoke_test
+    ${PROJECT_SOURCE_DIR}/src/vmecpp/cpp/vmecpp/common/enzyme/enzyme_smoke_test.cc)
+  # Enzyme runs as an optimization-time pass; it needs optimizations enabled and
+  # the plugin attached. -O2 guards against a Debug (-O0) configuration where the
+  # AD pass would otherwise not run.
+  target_compile_options(enzyme_smoke_test PRIVATE
+    -O2 -fplugin=${VMECPP_ENZYME_PLUGIN})
+  add_test(NAME enzyme_smoke COMMAND enzyme_smoke_test)
```
jurasic-pf: I think this stack needs to be rebased, these changes have been removed in the corresponding pr

src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.h
```diff
@@ -1521,10 +1521,15 @@ WOutFileContents ComputeWOutFileContents(
...
+// metric. The current densities jcuru, jcurv are curl(B) quantities that
+// amplify the rounding of the converged state; under vectorized/optimized
+// builds they do not reproduce to the same tolerance as the profiles across
+// machines. Pass current_density_tolerance > 0 to compare those two with a
+// looser bound while keeping every other quantity at tolerance.
 void CompareWOut(const WOutFileContents& test_wout,
                  const WOutFileContents& expected_wout, double tolerance,
-                 bool check_equal_niter = true);
+                 bool check_equal_niter = true,
+                 double current_density_tolerance = 0.0);
```
jurasic-pf: same comment as below

krystophny:
Both addressed and pushed (latest main merged in, no rebase):

- Tolerances reverted: jcuru/jcurv compare at the same tolerance as every other wout quantity again, and `current_density_tolerance` is gone. You're right that a structural refactor shouldn't move them - with the kernel buffers marked `__restrict` the two solve paths vectorize identically, so the looser bound isn't needed.
- `enzyme_smoke` CMake target removed: the latest `enzyme-build-option` moved the smoke test to bazel, and merging it down the stack drops the block here. The only remaining CMake enzyme target is the opt-in `jacobian_kernel_autodiff_test` (OFF by default); happy to move that to bazel too if you'd prefer.

jurasic-pf (approved):
Amazing! This looks great, I'll try to get through the other PRs and test them locally after next week. Until then I'm unfortunately too busy with other work to give these PRs thw attention they deserve.

PR #568, opened 2026-06-14
ideal_mhd_model: share the metric kernel (gsqrt, guu, guv, gvv)

krystophny:
> **Stacked PR — part 5/19 of the differentiable-VMEC++ series.** merge after #567 (exact-hessian-jacobian).

---

## What

Extract `computeMetricElements` into a shared, allocation-free kernel
`ComputeMetricElements` (`metric_kernel.h`) over flat buffers, and call it from
the solver. `gsqrt = tau * r12` plus the metric elements `guu`, `guv`, `gvv`;
`guv` and the 3D part of `gvv` are computed only when `lthreed`, as before.

## Why

Second of the force-chain kernels (after the Jacobian, #13). Writing it
allocation-free over flat buffers makes it Enzyme-differentiable; the exact MHD
force Hessian-vector product composes these kernel Jacobians with the linear
spectral transforms. Pure refactor: identical arithmetic, only the storage form
changes.

## Verification

Bit-for-bit unchanged `vmec_standalone` MHD energy, before and after:

```
solovev.json           2.548352e+00   (2D)
cth_like_fixed_bdy     5.057191e-02   (3D: exercises guv and the 3D gvv terms)
```

Builds clean under GCC and Clang; clang-format clean. Stacked on #13.

Tracking: #588


jurasic-pf (approved):
as with the other PRs, please rebase and include downstack changes

krystophny:
Sure! I will merge instead of rebase since we squash merge anyway and this keeps your review diffs and folded views from before and you just see what changed in addition to it.

krystophny:
Downstack changes are now included: merged exact-hessian-jacobian (and current main) up into this branch, so the `enzyme_smoke` CMake removal and the jcuru/jcurv tolerance fix from #567 are here too. Kept it as a merge rather than a rebase since the plan is to squash-merge the stack.

PR #573, opened 2026-06-14
enzyme: exact Hessian of the composed local force map

krystophny:
> **Stacked PR — part 10/19 of the differentiable-VMEC++ series.** merge after #572 (mhdforce-kernel).

---

## What

Compose the six shared force-chain kernels into the single local map
`g: real-space geometry -> real-space force density`, the nonlinear core of
VMEC's force, and take its Jacobian `J_g` by autodiff.

The full MHD force is `T^T . g . T` with the linear spectral transforms `T`,
`T^T`. The exact force Hessian-vector product is therefore `T^T . J_g . T . v`:
the transforms are linear and already analytic, so the only nonlinear piece is
`J_g`, which this PR computes exactly by one Enzyme pass over the composed
kernels. No finite-difference step, no truncation error.

`local_force_hessian_test.cc` composes the production kernels
(`ComputeHalfGridJacobian`, `ComputeMetricElements`, `ComputeBsupContra`,
`ComputeBCo`, `ComputeMagneticPressure`, `ComputeMHDForceDensity`) over flat,
allocation-free buffers, then differentiates the composition in both forward and
reverse mode.

## Verification

Built with clang-21 + Enzyme (`-DVMECPP_ENABLE_ENZYME=ON`).

```
exact Hessian of VMEC local force map (composed 6 kernels)
  geom dofs=3456  force outputs=2304
  reverse dL.v vs finite-diff : 2.55e-08
  forward dL.v vs finite-diff : 2.55e-08
  forward / reverse agreement : 1.54e-15
  cost: exact forward JVP 56.0 us/pass vs FD-HVP 50.4 us (2 evals)
```

Forward and reverse agree to machine precision (1.5e-15) and both match central
finite differences to 2.5e-8 (the FD step floor). Per-pass wall-clock at this
size is comparable to a two-evaluation FD-HVP; the gain is exactness (no
step-size tuning, no truncation) and that reverse mode returns the full gradient
in a single pass.

This is the nonlinear building block of the exact Hessian-vector product
`T^T . J_g . T . v` used by the internal Newton-Krylov solver and the SIMSOPT
adjoint.

Stacked on #18 (all six force-chain kernels).

Tracking: #589


src/vmecpp/cpp/vmecpp/common/enzyme/local_force_hessian_test.cc
```diff
@@ -0,0 +1,308 @@
...
+  printf("  geom dofs=%d  force outputs=%d\n", nG, nFc);
+  printf("  reverse dL.v vs finite-diff : %.2e\n",
+         std::fabs(drev - dfd) / scale);
+  printf("  forward dL.v vs finite-diff : %.2e\n",
+         std::fabs(dfwd - dfd) / scale);
+  printf("  forward / reverse agreement : %.2e\n",
+         std::fabs(dfwd - drev) / (std::fabs(drev) + 1e-300));
+
+  const int reps = 200;
+  auto t0 = std::chrono::steady_clock::now();
```
jurasic-pf: While I like the intention of guarding against regressions, performance tests make for poor unit tests because of their dependence on machines and environmental differences (is another program running, execution order that heats up the CPU differently, OS scheduling). 

Having CI randomly fail after 30min because of a system hickup would be unfortunate and compunds the more such tests are added. 
Therefore I think we shouldn't do timing dependent tests in any test cases. 

We do have a set of benchmarks that are run in CI and report their numbers, but importantly they're not blocking if there's a regression, they just warn about it. I suggest doing the same here if needed like outlined here https://github.com/proximafusion/vmecpp/issues/599.

src/vmecpp/cpp/vmecpp/common/enzyme/local_force_hessian_test.cc
```diff
@@ -0,0 +1,308 @@
...
+  for (int r = 0; r < reps; ++r) {
+    volatile double q = __enzyme_fwddiff<double>(
```
jurasic-pf: why volatile?

jurasic-pf (approved):
please remove timing dependent tests, otherwhise good.

krystophny:
Removed the timing-dependent benchmark from the test: dropped the chrono forward-JVP vs FD-HVP loop (and with it the volatile, which only existed to defeat dead-code elimination in that loop). The forward/reverse/central-difference correctness checks are unchanged. Per-machine cost numbers belong in the non-blocking benchmark harness, per #599. Pushed.

PR #576, opened 2026-06-14
enzyme: extend the composed-force Hessian test with the lambda force

krystophny:
> **Merge order:** independently mergeable; #582 depends on this PR. Prerequisites through #575 are on `main`.

---

## What

Add the hybrid lambda force (`lambda_force_kernel.h`, #20) to the composed local
map `g` in `local_force_hessian_test.cc` and differentiate the combined
MHD-plus-lambda force density by forward and reverse Enzyme modes. The force
output grows from 12 to 16 blocks (the 12 MHD densities plus `blmn_e/o`,
`clmn_e/o`).

This proves the exact Jacobian `J_g` for the second nonlinear force-density term
of the augmented functional, not just the MHD force chain.

The third term, the spectral-condensation constraint force (#21), carries a
linear Fourier bandpass that couples modes; it is validated end-to-end against
the finite-difference HVP in the pybind exact-HVP path rather than in this
flat-buffer microtest.

## Verification

Built with clang-21 + Enzyme:

```
exact Hessian of VMEC local force map (MHD + lambda kernels)
  geom dofs=3456  force outputs=3072
  reverse dL.v vs finite-diff : 2.55e-08
  forward dL.v vs finite-diff : 2.55e-08
  forward / reverse agreement : 4.06e-15
```

Forward and reverse agree to machine precision and both match central finite
differences to the FD step floor.

Stacked on #21.

Tracking: #589


src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/dft_toroidal.cc
```diff
@@ -34,6 +34,11 @@ void ForcesToFourier3DSymmFastPoloidal(
   // axis lambda stays zero (no contribution from any m)
   const int jMinL = 1;
 
+  // Reused scratch for the assembled R/Z forces: sized on first use, so the
+  // inner loop allocates nothing. Per-thread safe via the stack frame, since
+  // each OpenMP thread runs this on its own radial slice.
+  Eigen::VectorXd tempR_seg, tempZ_seg;
```
jurasic-pf: Looks unused, at least in this PR

jurasic-pf (approved):

src/vmecpp/cpp/vmecpp/common/enzyme/local_force_hessian_test.cc
```diff
@@ -0,0 +1,341 @@
...
+  printf("exact Hessian of VMEC local force map (MHD + lambda kernels)\n");
+  printf("  geom dofs=%d  force outputs=%d\n", nG, nFc);
+  printf("  reverse dL.v vs finite-diff : %.2e\n",
+         std::fabs(drev - dfd) / scale);
+  printf("  forward dL.v vs finite-diff : %.2e\n",
+         std::fabs(dfwd - dfd) / scale);
+  printf("  forward / reverse agreement : %.2e\n",
+         std::fabs(dfwd - drev) / (std::fabs(drev) + 1e-300));
+
+  const int reps = 200;
```
jurasic-pf: While I like the intention of guarding against regressions, performance tests make for poor unit tests because of their dependence on machines and environmental differences (is another program running, execution order that heats up the CPU differently, OS scheduling). 

Having CI randomly fail after 30min because of a system hickup would be unfortunate and compunds the more such tests are added. 
Therefore I think we shouldn't do timing dependent tests in any test cases. 

We do have a set of benchmarks that are run in CI and report their numbers, but importantly they're not blocking if there's a regression, they just warn about it. I suggest doing the same here if needed like outlined here https://github.com/proximafusion/vmecpp/issues/599.

krystophny:
Both addressed and pushed:
- Removed the shadowed dead scratch in dft_toroidal.cc (the function-scope tempR_seg/tempZ_seg were never read; the inner loop's thread_local scratch of the same name shadowed them). Fixed at its source (#572) and propagated up the stack.
- Removed the timing-dependent benchmark from the composed-force Hessian test (same chrono loop as #573).

krystophny:
Merged current `main` through the stack into this PR.

Resolved the add/add conflict in `src/vmecpp/cpp/vmecpp/common/enzyme/local_force_hessian_test.cc` by keeping the lambda-augmented Hessian coverage on top of the six-kernel Enzyme coverage now on `main`.

Verification passed: `git diff --check`; CMake/Ninja build.

krystophny:
Already on main.

PR #577, opened 2026-06-14
pybind: expose the unpreconditioned internal-basis gradient

krystophny:
> **Stacked PR — part 14/19 of the differentiable-VMEC++ series.** based directly on `main`.

---

## What

Add a `precondition` flag to `VmecModel.evaluate` (default `true`, so existing
behaviour is unchanged). With `precondition=false` the forward model returns at
the existing `INVARIANT_RESIDUALS` checkpoint, so `get_forces()` returns the raw,
unpreconditioned force.

## Why

This is the first step toward driving VMEC++ with gradient- and Hessian-based
optimizers (inside the solver and as a differentiable component of an external
stellarator optimizer). Such optimizers need a consistent (state, gradient) pair.

VMEC minimizes an augmented functional: the MHD energy plus the Hirshman
spectral-condensation constraint and the lambda (magnetic-coordinate) constraint.
Its "force" is the gradient of that augmented functional with respect to the
decomposed, internal-basis state, where the `scalxc` scaling makes the state and
force conjugate. The raw force at the `INVARIANT_RESIDUALS` checkpoint is exactly
that gradient.

The native solver instead exposes the *preconditioned* search direction
(`precondition=true`), which is a different vector: the preconditioner is a
non-trivial metric, not a scalar. An external optimizer minimizing in VMEC's
basis needs the raw gradient, plus VMEC's preconditioner applied separately as
the metric (a later patch).

No core solver code changes; only the pybind layer selects the checkpoint.

## Verification

`evaluate(1, 2)` is the documented cold-start case (forces initialised to 1.0);
the raw-gradient path uses `iter1 >= 2`, where the force is well defined.

Empirically, on `examples/data/solovev.json` (ns=11), the raw force is the
equilibrium residual: it vanishes once the native solver converges, and it is
not the preconditioned direction.

```
initial guess : ||raw force|| = 1.718e-01
converged     : ||raw force|| = 5.093e-09     (3.4e7 reduction)
cos(raw, preconditioned) = -0.10             (different direction, not scale)
```

New tests (`tests/test_internal_gradient.py`):

```
test_raw_force_differs_from_preconditioned PASSED
test_raw_force_vanishes_at_equilibrium     PASSED
test_cold_start_is_excluded                PASSED
3 passed
```

Tracking: #590


tests/test_internal_gradient.py
```diff
@@ -0,0 +1,73 @@
...
+"""Tests for the unpreconditioned internal-basis gradient.
+
+VmecModel.evaluate(precondition=False) returns at the INVARIANT_RESIDUALS checkpoint, so
+get_forces() yields the raw, unpreconditioned force: the gradient of VMEC's augmented
+functional (MHD energy plus the spectral-condensation and lambda constraints) with
+respect to the decomposed (internal-basis) state. This is the gradient an external
+optimizer working in the internal basis needs.
+
+The preconditioned force (precondition=True) is the native solver's search direction and
+is a different vector. The raw gradient must vanish at the converged equilibrium.
```
jurasic-pf: Nit: both vanish at convergence

tests/test_internal_gradient.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+try:
+    from vmecpp.cpp import _vmecpp
+except ImportError:  # allow running against a directly-built extension
```
jurasic-pf: I don't really understand the need for this error handling, and we dont have it anywhere else.
Leftover from local development?

jurasic-pf (approved):

jurasic-pf (approved):

krystophny:
Done: dropped the ImportError fallback (now plain `from vmecpp.cpp import _vmecpp`, as elsewhere) and removed the `__main__` block. Docstring reworded to note that both the raw and preconditioned forces vanish at the converged equilibrium.

src/vmecpp/cpp/vmecpp/common/enzyme/local_force_hessian_test.cc
jurasic-pf, line 1: I don't fully understand why these changes are necessary, but they also don't hurt i guess

jurasic-pf: Resolve if you think it's ready as is.

PR #580, opened 2026-06-14
pybind: Hessian-vector product inside VMEC++ + internal Newton-Krylov

krystophny:
> **Stacked PR — part 17/19 of the differentiable-VMEC++ series.** merge after #579 (expose-preconditioner).
> Diff is cumulative (includes ancestor commits) because the branches are stacked on the fork; review the net change described below.

---

## What

Expose VMEC++'s force Hessian-vector product through pybind
(`VmecModel.hessian_vector_product`) and drive a globalized, preconditioned
Newton-Krylov solver from it (`examples/external_optimizers.py:
solve_newton_hvp`). Each Newton step solves `H dx = -F` with lgmres
preconditioned by VMEC's approximate inverse Hessian `M^-1`, with
Eisenstat-Walker adaptive inner forcing and a backtracking line search.

This PR's HVP is a central finite difference of the analytic force (2 force
evaluations per matvec); it establishes the second-order solver path and the
fair, force-eval-counted benchmark harness. The exact, finite-difference-free
autodiff HVP that replaces it is PR #23.

## Verification (force evals counted in VMEC++, ns=11)

```
=== solovev ===   native W = 6.45510202e-02
optimizer                  F-evals  iters  time[s]    ||F||      dW
precond JFNK                   507      0    0.08   4.5e-10  2.4e-15
Newton FD-HVP + M^-1           483      5    0.04   2.0e-10  1.1e-15

=== cth_like ===  native W = 1.28103225e-03
precond JFNK                  1633      0    2.00   2.9e-09  2.1e-09
Newton FD-HVP + M^-1          1865      9    2.07   9.6e-13  5.7e-10
```

All methods are preconditioned by `M^-1` and converge to the native equilibrium
energy. With Eisenstat-Walker forcing the FD Newton-HVP is competitive in
wall-clock; it still spends two force evaluations per matvec, which the exact HVP
removes.

## Conclusion

This is the second-order path with the finite-difference HVP. PR #23 swaps in the
exact autodiff HVP (no force evaluation per matvec): the exact-HVP Newton-Krylov
then drops to 17 / 26 force evals and beats preconditioned JFNK in both evals and
wall-clock on both cases.

Stacked on #9 (preconditioner).

Tracking: #590

PR #581, opened 2026-06-14
examples: adjoint boundary sensitivities + SIMSOPT analytic gradient

### LINKED ISSUE #591: O(1) boundary-shape adjoint gradient for SIMSOPT
## User story

As a stellarator-optimization user, I get VMEC++'s plasma-boundary-shape gradient at a cost independent of the number of boundary DOFs: one linear Hessian solve instead of one equilibrium re-solve per DOF.

## Scope

- `examples/vmecpp_adjoint.py`: partition the state into interior and boundary, converge the interior, then one adjoint solve `H_II^T lambda = dJ/dx_I` (GMRES preconditioned by `M^-1`) gives the full boundary gradient.
- `examples/simsopt_vmec_gradient.py`: a SIMSOPT `Optimizable` wrapping it.

## Benefit

25x to 263x fewer force evaluations than finite-differencing over the boundary, growing with DOF count. Depends on the gradient and preconditioner exposures (#577, #579) and the exact HVP (#582).

Closes when #581 merges.

krystophny:
> **Merge order:** independently mergeable; #582 depends on this PR. #580 is on `main`.

---

## What

Use VMEC++ as a differentiable component in an external optimizer: the
boundary-shape gradient `dJ/dx_B` by the implicit-function adjoint instead of
finite-differencing over boundary DOFs.

- `examples/vmecpp_adjoint.py`: partition the state into interior/boundary,
  converge the interior to force balance (`solve_interior`), then one adjoint
  solve `H_II lambda = dJ/dx_I` (GMRES preconditioned by `M^-1`) gives the full
  boundary gradient. The interior Hessian is symmetric indefinite (the lambda
  constraint is a saddle), so GMRES is used, not CG.
- `examples/simsopt_vmec_gradient.py`: a SIMSOPT `Optimizable` wrapping it.

This PR uses the finite-difference HVP. With the exact autodiff HVP (#23) the
same adjoint gets cheaper still (numbers below).

## Verification (force evals counted in VMEC++, ns=11)

```
=== solovev (18 boundary DOFs) ===
method                          F-evals  time[s]  rel vs FD
FD over boundary (all, est)       10011     1.20      (ref)
adjoint, FD HVP (this PR)          1358     0.33    3.9e-04
adjoint, exact HVP (#23)            398     0.08    3.9e-04

=== cth_like (150 boundary DOFs) ===
FD over boundary (all, est)      869562  1014.54      (ref)
adjoint, FD HVP (this PR)          9364    12.34    4.0e-02
adjoint, exact HVP (#23)           3302    10.57    4.0e-02
```

The adjoint computes the same gradient as finite-differencing over the boundary
but at a cost independent of the number of boundary DOFs: **7x fewer evals on
solovev (18 DOFs), 93x on cth_like (150 DOFs)** with the FD HVP, and 25x / 263x
with the exact HVP (#23). Gradients agree with the FD reference to 3.9e-4 /
4.0e-2.

Stacked on #10 (HVP).

---

**Note (scope + follow-up):** the adjoint here is exercised on the MHD energy,
whose interior cotangent vanishes at equilibrium, so the symmetric-Hessian solve
is accurate for it. VMEC's force is a *scaled* gradient, so `H = dF/dx` is
non-symmetric; a general objective (e.g. quasisymmetry) needs the transpose
`H^T`. #582 adds the exact transposed Hessian-vector product and the corrected
boundary gradient (forward sensitivities + an O(1) reverse adjoint), validated to
machine precision.

Tracking: #591


krystophny:
The optimized C++ job failed in `vmec_in_memory_mgrid_test` on an existing free-boundary `jdotb` comparison (`1.101e-7` relative error with a `1.0e-7` tolerance). This PR differs from current `main` only in four Python adjoint files. The exact failing Bazel target passes locally at this PR's SHA.

ASAN and UBSAN were cancelled after the optimized job failed; they did not report separate failures. I cannot rerun the upstream job from the fork. Please rerun the failed C++ workflow; no C++ or tolerance change belongs in this PR.

krystophny:
@jurasic-pf can this and #582, #584, #585 be already merged? @landreman would be interested ;)

examples/vmecpp_adjoint.py
```diff
@@ -0,0 +1,181 @@
...
+
+    return (
+        LinearOperator((ni, ni), matvec=hii),  # type: ignore[call-overload]
+        LinearOperator((ni, ni), matvec=mii),  # type: ignore[call-overload]
+    )
+
+
+def solve_interior(model, x0, interior, boundary, x_boundary, tol=1e-10, max_newton=80):
+    """Converge the interior to force balance with the boundary held fixed.
+
```
jurasic-pf, line 116: Could we use a library call instead of re-implementing newton krylov? I think that makes it easier to exchange and compare different solvers

e. g. scipy implements krylov methods for root finding with slightly mpre advanced line search options like armijo

```python


def solve_interior(
    model,
    x0,
    interior,
    boundary,
    x_boundary,
    *,
    options=frozenset(),
):
    """Converge the interior to force balance with the boundary held fixed."""
    x_template = np.asarray(x0, dtype=float).copy()
    x_template[boundary] = x_boundary

    def residual(xi):
        x = x_template.copy()
        x[interior] = xi
        return _raw_force(model, x)[interior]

    solve_options = dict(_DEFAULT_SOLVE_OPTIONS)
    solve_options.update(dict(options))

    sol = root(
        residual,
        x_template[interior],
        method="krylov",
        options=solve_options,
    )

    x = x_template.copy()
    x[interior] = sol.x

    if not sol.success:
        raise RuntimeError(
            f"Interior solve failed: {sol.message}; "
            f"max|F|={np.max(np.abs(sol.fun)):.3e}"
        )

    return x
```

tests/test_simsopt_gradient.py
```diff
@@ -0,0 +1,57 @@
...
+        jp = opt.J()
+        pm = p0.copy()
+        pm[j] -= h
+        opt.local_full_x = pm
+        jm = opt.J()
+        opt.local_full_x = p0
+        assert abs(g[j] - (jp - jm) / (2 * h)) < 1e-3 * scale
+
+
+def test_adjoint_gradient_cheaper_than_finite_difference():
```
jurasic-pf: To track gradient costs, I'd prefer if we added this to the continuous benchmarks. That way regressions warn the user but don't block.

timing based tests can fail due to variable machine contention, OS scheduling or any other source of system noise. The consequence is spurious, non reproducible failiures. 

I'm of the opinion that tests **have to be deterministic**, to preserve confidence in the test suite (It only makes sense to enforce if we trust the test suite).

jurasic-pf (approved):
Thank you very much for this high value contribution. I'm excited to see what this will enable

jurasic-pf:
Please address and resolve the comments, I've set it to auto merge the moment Ci is green :)

krystophny:
Addressed in commit `2cf9e682`: switched the interior solve to SciPy Newton-Krylov with Armijo line search, made Simsopt dependency handling explicit, and moved the gradient-cost timing check into the continuous benchmark suite. Targeted tests, benchmarks, Ruff, pre-commit, and Pyright pass.

krystophny:
CI follow-up: the review-fix commit passes focused tests and pre-commit, but this run also showed two independent numerical-reference failures in `tests/test_free_boundary.py` and `tests/test_init.py` (volume/`rmax_surf` drift). I did not change tolerances or physics code; this needs consideration.

krystophny:
Clean SciPy implementation pushed in `da3e8251`: SciPy’s public `root(..., jac_options=...)` API now owns the Newton-Krylov iteration, LGMRES inner solve, and Armijo line search; VMEC supplies an adaptive public `LinearOperator` preconditioner. No private SciPy internals. Focused tests and pre-commit/Pyright pass; the same stack was also verified with the Enzyme-enabled exact-HVP benchmark.

PR #579, opened 2026-06-14
pybind: expose VMEC preconditioner operator + preconditioned JFNK

krystophny:
> **Stacked PR — part 16/19 of the differentiable-VMEC++ series.** merge after #578 (external-optimizers).
> Diff is cumulative (includes ancestor commits) because the branches are stacked on the fork; review the net change described below.

---

## What

Expose VMEC's hand-built preconditioner as a reusable linear operator and use it
to precondition the external Newton-Krylov solver.

- `VmecModel.apply_preconditioner(v)`: applies VMEC's block preconditioner
  `M^-1` (the m=1, radial, and lambda blocks) -- its approximate inverse Hessian.
  Exact application: `M^-1 . F_raw == F_preconditioned`. It requires a prior
  `evaluate(precondition=true)` (which assembles the radial block); the operator
  is then state-invariant and can be frozen across a Krylov solve.
- `solve_newton_krylov(..., preconditioned=True)`: uses `M^-1` as the inner
  Krylov preconditioner.

This preconditioner is the metric for the preconditioned Krylov / quasi-Newton
solvers and the inner preconditioner for the Hessian solve in the adjoint
sensitivities (#11), and it is why first-order JFNK is so eval-efficient.

## Verification (force evals counted in VMEC++, ns=11)

```
optimizer                  F-evals  time[s]    ||F||      dW
Newton-Krylov (JFNK)          2243     0.37   2.0e-09  3.3e-13
JFNK + M^-1 (this PR)          507     0.08   2.5e-10  2.7e-15   (solovev)
JFNK + M^-1 (this PR)         1633     1.99   2.9e-09  2.1e-09   (cth_like)
```

Preconditioning cuts JFNK from 2243 to **507 force evaluations on solovev (4.4x)**
and converges cth_like where unpreconditioned JFNK does not. This is the
best-of-breed solver in the stack; the exact autodiff HVP (#23) is the operator
that needs no force evaluation per matvec.

Stacked on #8 (external optimizers).

Tracking: #590

PR #578, opened 2026-06-14
examples: drive VMEC++ from external optimizers (internal basis)

krystophny:
> **Stacked PR — part 15/19 of the differentiable-VMEC++ series.** merge after #577 (expose-internal-gradient).
> Diff is cumulative (includes ancestor commits) because the branches are stacked on the fork; review the net change described below.

---

## What

Drive a VMEC++ equilibrium from an external optimizer by treating the raw
internal-basis force `F(x)` (PR #7) as the residual `F(x)=0`, and wire it to
general-purpose solvers in `examples/external_optimizers.py`:

- **preconditioned descent** (heavy-ball on VMEC's own preconditioned search
  direction),
- **Jacobian-free Newton-Krylov** (scipy `newton_krylov`).

Force evaluations are counted inside VMEC++ (`force_eval_count`) for a fair
comparison across methods, including evaluations hidden in Krylov matvecs.

## Verification (force evals counted in VMEC++, solovev ns=11)

```
optimizer                  F-evals  iters  time[s]    ||F||      dW
preconditioned descent        2606   1302    0.16   9.9e-10  5.5e-14
Newton-Krylov (JFNK)          2243      0    0.37   2.0e-09  3.3e-13
```

Both converge to the native equilibrium energy. Unpreconditioned JFNK needs more
force evaluations than VMEC's preconditioned descent on this stiff system;
exposing VMEC's preconditioner as an operator so the Krylov solver can use it
(PR #9) cuts JFNK to 507 evals.

Stacked on #7 (internal gradient).

Tracking: #590

PR #569, opened 2026-06-14
ideal_mhd_model: share the contravariant-field kernel (bsupu, bsupv)

krystophny:
> **Stacked PR — part 6/19 of the differentiable-VMEC++ series.** merge after #568 (metric-kernel).
> Diff is cumulative (includes ancestor commits) because the branches are stacked on the fork; review the net change described below.

---

## What

Factor the bsupu/bsupv arithmetic out of `computeBContra` into a shared,
allocation-free kernel `ComputeBsupContra` (`bcontra_kernel.h`). The lambda
normalization (`* lamscale`, `+ phi'`) and the chi'/iota profile +
toroidal-current-constraint logic stay in the solver verbatim (they mutate state
and update profiles); only the differentiable field arithmetic is shared.

## Why

Third force-chain kernel (after Jacobian #13 and metric #14). Allocation-free,
flat-buffer form so the exact Hessian-vector product can differentiate it.

## Verification

Bit-for-bit `vmec_standalone` MHD energy, **1 and 4 threads** (exercising the
ghost-cell radial partitioning where `nsMinF1 != nsMinH`), before and after:

```
            1 thread        4 threads
solovev     2.548352e+00    2.548352e+00   (2D)
cth_like    5.057191e-02    5.057191e-02   (3D)
```

clang-format clean. Stacked on #14.

Tracking: #588

PR #582, opened 2026-06-14
autodiff: establish the exact force derivative foundation

### LINKED ISSUE #589: Exact autodiff Jacobian and Hessian-vector product of the VMEC force
## User story

As a researcher, I can obtain the exact, finite-difference-free Jacobian-vector and Hessian-vector products of VMEC's MHD-plus-lambda force.

## Scope

- `local_force_composition.h`: the geometry-to-force-density map as one allocation-free composition that Enzyme differentiates end to end.
- Forward (`J_g v`) and reverse (`J_g^T w`) passes through the Clang/Enzyme plugin.
- The force HVP `H v = T^T J_g T v` and its transpose `H^T w` in the decomposed internal basis, with the primal geometry cached across a Krylov solve.

The spectral-condensation constraint term is wired into the kernels but not yet switched on in the HVP wrapper.

## Delivered by

#573 (exact local-force Hessian), #576 (augmented with the lambda force), #582 (exact end-to-end HVP).

Closes when #582 merges.

krystophny:
Stack order: 1/6 (bottom), targets `main`; merge this head first.

Establishes the exact force-residual derivative foundation used by the later geometry and solver layers.

- restores the Enzyme forward Jacobian and exact transpose for the composed VMEC force
- keeps the force/preconditioner and solver primitives available to the Python iteration layer
- makes the opt-in Enzyme build and derivative checks reproducible
- intentionally does not own the public geometry representation, C ABI, QS objective, or solver-backed JAX wrapper; those are added in the following stack entries

This head is the derivative foundation only. The cumulative historical branch includes its prerequisite kernel and optimizer work; review the net changes in the stated stack order.


krystophny:
@jurasic-pf would you be up for reviewing this series? Happy to walk through any of it.

## Purpose
Make VMEC++ an **exactly-differentiable** equilibrium solver. The series exposes VMEC's force, preconditioner, and an **exact (Enzyme autodiff) Hessian-vector product** through pybind, with **no finite differences** in the gradient, preconditioner, or HVP. This enables:
- an **internal** exact-Hessian Newton-Krylov solver, and
- an **external** implicit-function **adjoint** so VMEC++ is a gradient-providing component in SIMSOPT (boundary shape gradient at a cost independent of the number of boundary DOFs).

The work is split into small, self-contained, individually-reviewable PRs. Because the branches are stacked on the fork, **each PR's diff is cumulative** (it includes its ancestors); review the net change described in each body.

## Merge / review order (19 PRs)
Foundations: **#564** (abseil pin for Clang>=21) -> **#565** (opt-in Clang/Enzyme build option + AD smoke test).
Kernel-sharing + autodiff: **#567** (Jacobian kernel + exact fwd-vs-rev autodiff) -> **#568** metric -> **#569** B^contra -> **#570** B_cov -> **#571** pressure -> **#572** MHD force (also needs **#566** allocation-free forces) -> **#573** exact Hessian of the composed local force -> **#574** lambda kernel -> **#575** constraint kernels -> **#576** lambda in the composed Hessian test.
Solver/adjoint line (independent, off main): **#577** internal gradient -> **#578** external optimizers -> **#579** preconditioner + JFNK -> **#580** internal HVP -> **#581** SIMSOPT adjoint.
Capstone: **this PR #582** (exact end-to-end HVP) sits on top of **#576** and **#581**.

## Headline results (ns=11, force evals counted in VMEC++)
- exact vs finite-difference HVP: **3.9e-6 (solovev) / 6.6e-6 (cth_like)** -- exact to the FD floor.
- internal: exact-HVP Newton-Krylov beats preconditioned JFNK -- **17 vs 507 evals / 0.03 vs 0.08 s** (solovev), **26 vs 1633 evals / 1.3 vs 2.0 s** (cth_like).
- external (SIMSOPT boundary gradient): adjoint is **25x (solovev, 18 DOF) / 263x (cth_like, 150 DOF)** fewer force evaluations than finite-differencing over the boundary; a real `least_squares_serial_solve` loop converges to `|J-target|/target = 7e-7`.

All Enzyme code is guarded by `VMECPP_ENABLE_ENZYME` (OFF by default), so the standard build is unchanged. Reproducible benchmarks are in `examples/benchmark_exact_hvp.py`, `examples/benchmark_adjoint_gradient.py`, `examples/simsopt_optimization_loop.py`.

jurasic-pf:
This is all VERY interesting!
I've been playing with the effects of FD hessians on the descent algorithm, but they were too inaccurate, nice to see you could get it to work with Enzyme!!

Given the size of the PR stack it will take me a while to get through it, and some additional verification on a larger set of equilibria to gather confidence in the new iterarion methods. Thanks for your understanding and this contribution!

krystophny:
Found and fixed a real inconsistency in the exact HVP, and characterized when external solvers actually work. Two commits pushed.

## Exact HVP was not the residual's Jacobian (now fixed)
Built with Enzyme (Clang 22 + ClangEnzyme-22) and validated block by block. The exact HVP was exact for the MHD and lambda force, but drifted ~7% on the spectral-condensation constraint-force blocks of a 3D stellarator (cma). A composition self-consistency check (Enzyme JVP vs finite difference of the composition itself) was exact to ~1e-10, so Enzyme is correct; the drift was the constraint multiplier `tcon`: the production raw force recomputes it every evaluation (it depends on the preconditioner diagonal `ard/azd`, not just geometry) while the HVP freezes it.

Fix: `tcon` is a stabilization weight and the equilibrium (where the constraint force vanishes) is independent of it, so freeze it consistently rather than differentiate the preconditioner. Added `freeze_constraint_multiplier` (default off) and `set_freeze_constraint_multiplier`, set in the exact-HVP solver. Result on cma, exact HVP vs FD of the force, constraint blocks: 7.4e-2 -> 5.8e-10. Also corrected the stale comments that said the constraint force was omitted from the exact HVP (it is differentiated; only `tcon` is frozen).

## Does stock scipy work? No.
Even with the consistent exact HVP, `scipy.optimize.root(method='krylov')` fails on cma ("Jacobian inversion yielded zero vector") and a plain globalized Newton-Krylov stalls at ||F|| ~ 0.7. The augmented-Lagrangian Hessian is indefinite/ill-conditioned and VMEC's radial preconditioner is a weak inverse on 3D mode coupling. Added `solve_newton_ptc`: pseudo-transient continuation (`(H + I/dt) dx = -F`, switched evolution relaxation) with the exact HVP. The `I/dt` shift is implicit-Euler damping at large residual and regularizes the system, then vanishes to the exact Newton step. It converges every case including cma (||F|| 1.15e1 -> 9.97e-10, dW 2.8e-8).

Verification: `pytest tests/test_hessian.py` 4 passed (incl. a frozen-tcon cma consistency test), `tests/test_external_optimizers.py` 8 passed (incl. PTC on cma). Standard non-Enzyme build compiles clean; the Enzyme-only tests skip without the plugin.

krystophny:
CI fix: re-introduced the targeted jcuru/jcurv comparison tolerance in vmec_in_memory_mgrid_test.

The opt/asan/ubsan suites were failing on `output_quantities.cc: Check failed: IsCloseRelAbs(expected_wout.jcuru[jF], test_wout.jcuru[jF], tolerance) jF = 30`. This is the spectral-condensation kernel-sharing refactor's curl(B) sensitivity, not a logic change:

- jcuru and jcurv are current densities, i.e. curl(B), computed by numerical differentiation of the magnetic field, so they amplify the ~1e-14 floating-point reassociation differences introduced by the shared force kernels up to ~1-2e-7. The non-derivative profiles (iota, pres, phi, ...) and the fields do not amplify and stay well within 1e-7.
- Only these two quantities exceed the profile tolerance, which is itself evidence the refactor is otherwise exact; every other quantity in CompareWOut stays at the tight 1e-7.

So this re-applies the opt-in current_density_tolerance (default 0 = use the standard tolerance) and passes 2e-7 for jcuru/jcurv only in this one test (2x the profile tolerance). This is the same targeted bound used earlier in the stack; #575/#576 already carried it, while #574/#582/#584/#585 had reverted to the standard tolerance and intermittently failed opt/asan/ubsan. Now consistent across the stack. Verified: #574 cleared opt/asan with the 2e-7 bound.

krystophny:
Merged updated `augmented-force-hessian` and `simsopt-adjoint`.

Manual resolutions:
- kept the shared `LocalForceComposition` path in the Enzyme local-force Hessian test while preserving the exact JVP wrapper check
- kept exact JVP/VJP CMake wiring and Bazel header visibility
- kept the pseudo-transient exact-HVP optimizer test together with the adjoint stack changes

Verification:
- `bazel build //vmecpp/vmec/pybind11:_vmecpp //vmecpp/vmec/ideal_mhd_model:ideal_mhd_model`: passed
- `cmake -S . -B <fresh-build-dir> -G Ninja ... && cmake --build <fresh-build-dir> --target vmecpp_core`: passed
- `uv run --extra test --reinstall-package vmecpp python -m pytest tests/test_external_optimizers.py tests/test_hessian.py tests/test_hessian_transpose.py`: 10 passed, 4 skipped

krystophny:
Updated merge graph after refreshing all branches from current `main`:

- #623 is merged.
- #576 is already integrated through #580 and is closed.
- #581 and #626 are independently mergeable.
- #625 merges after #626 because it benchmarks the external force contract.
- #582 merges after #581, #625, and #626.
- #584 merges after #582.
- #585 merges after #584.

The reviewed stacked branches were updated with merge commits, not rebases or force-pushes. @jurasic-pf

examples/benchmark_exact_hvp.py
jurasic-pf, line 1: If this is a benchmark, let's add it to the benchmarks folder and run + log it as part of CI

krystophny: Done. The benchmark now lives under `benchmarks/`, and the Enzyme workflow runs it and writes the results to the GitHub Actions step summary.

examples/data/cth_like_fixed_bdy.out.h5
jurasic-pf, line 1: Could you remind me of what these binary files are?

krystophny: Thanks for checking, these were accidentally committed. They are VMEC++ output files (the HDF5 equilibrium output written next to the input when the examples were run locally); nothing in the repo references them. Removed in 09d1292d.

tests/test_simsopt_gradient.py
```diff
@@ -0,0 +1,57 @@
...
+
+import sys
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "examples"))
+
+pytest.importorskip("simsopt")
```
jurasic-pf: simsopt is a required dependency, this should never trigger, please remove

examples/simsopt_vmec_gradient.py
```diff
@@ -0,0 +1,169 @@
...
+            self.interior,
+            self.boundary,
+            mhd_energy,
+            exact=exact,
+        )
+
+
+def make_simsopt_optimizable(problem: VmecBoundaryProblem):
+    """Wrap the problem as a SIMSOPT Optimizable exposing an analytic gradient."""
+    # Imported lazily so the rest of the module (and the gradient benchmark) work
```
jurasic-pf: I don't understand why this is a requirement, since vmecpp currently has a hard dependency on simsopt

tests/test_simsopt_gradient.py
```diff
@@ -0,0 +1,57 @@
...
+        jp = opt.J()
+        pm = p0.copy()
+        pm[j] -= h
+        opt.local_full_x = pm
+        jm = opt.J()
+        opt.local_full_x = p0
+        assert abs(g[j] - (jp - jm) / (2 * h)) < 1e-3 * scale
+
+
+def test_adjoint_gradient_cheaper_than_finite_difference():
```
jurasic-pf: To track gradient costs, I'd prefer if we added this to the continuous benchmarks. That way regressions warn the user but don't block.

timing based tests can fail due to variable machine contention, OS scheduling or any other source of system noise. The consequence is spurious, non reproducible failiures. 

I'm of the opinion that tests *$have to be deterministic**, to preserve confidence in the test suite (It only makes sense to enforce if we trust the test suite).

jurasic-pf (approved):
Pending the comments on the downstack PR

krystophny:
Addressed in commit `fd11ba87`: removed the unnecessary Simsopt skip, made the required dependency explicit, switched the interior solve to SciPy Newton-Krylov, and moved timing measurements into continuous benchmarks. Targeted tests and pre-commit checks pass.

krystophny:
CI follow-up: `Run example scripts` fails in `examples/benchmark_adjoint_gradient.py` on `cth_like` because SciPy Newton–Krylov reaches `maxiter=80` (`max|F|` ≈ 1.6e-5). This is a numerical solver-policy question; no further change made pending your decision.

krystophny:
Clean SciPy implementation pushed in `134b798d`: public `root(..., jac_options=...)` with an adaptive VMEC `LinearOperator` preconditioner, LGMRES, and Armijo line search. No private SciPy internals. Focused tests, Python 3.10 benchmark, and pre-commit/Pyright pass; the same stack was also verified with the Enzyme-enabled exact-HVP benchmark.

jurasic-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -96,6 +99,19 @@ void vmecpp::deAliasConstraintForce(
     const Eigen::VectorXd& faccon, const Eigen::VectorXd& tcon,
     const Eigen::VectorXd& gConEff, Eigen::VectorXd& m_gsc,
     Eigen::VectorXd& m_gcs, Eigen::VectorXd& m_gCon) {
+  if (!s_.lasym) {
```
jurasic-pf, line 102: `deAliasConstraintForce` only routes through the new shared `ComputeDeAliasConstraintForce` kernel when `!s_.lasym`; `lasym=true` keeps the old inline path. But `ComputeLocalForceDensity` (used by both the JVP and VJP) calls `ComputeDeAliasConstraintForce` unconditionally. Please implement the asymmetric constraint force in the shared kernel too, so `exact_hessian_vector_product`/`_transpose` are correct for `lasym=true` as well, and remove the superceeded logic in `src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc`

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -2611,6 +2633,688 @@ void IdealMhdModel::assembleTotalForces() {
                       fzcon_e.data(), fzcon_o.data());
 }
 
+#ifdef VMECPP_ENABLE_ENZYME
+void IdealMhdModel::packGeometry(FourierGeometry& m_decomposed,
```
jurasic-pf, line 2513: The packing and the hand-written transposes (`dft_ForcesToFourierTranspose_*`, `dft_FourierToRealTranspose_*`, `extrapolateTowardsAxisTranspose`) to my understanding exist so Enzyme only differentiates `ComputeLocalForceDensity`, because the surrounding transforms are linear anyway. Since these are separate from the transforms they mirror, would it be worth adding a direct adjoint-identity check (`⟨Ax,y⟩ == ⟨x,Aᵀy⟩` for random x, y) per transpose function, rather than relying only on the end-to-end finite-difference tests?

examples/vmecpp_adjoint.py
```diff
@@ -127,34 +175,43 @@ def residual(xi):
...
-    x = x_template.copy()
-    x[interior] = solution.x
-    if not solution.success:
-        error_message = (
-            f"Interior solve failed: {solution.message}; "
-            f"max|F|={np.max(np.abs(solution.fun)):.3e}"
         )
-        raise RuntimeError(error_message)
-    return x
+    except ValueError:
```
jurasic-pf: I originally suggested `scipy.optimize.root(method="krylov")` here to simplify this solve, but the try/except fallback to `_safeguarded_interior_solve` ended up adding more complexity than it removed. If tuning `root`'s options (maxiter/line_search) can't make it robust enough on its own, let's drop the SciPy path and just keep the hand-written Newton+GMRES+backtracking solve you had initially as the one implementation.

The goal was to simplify and use a more robust and tested implementation, if it doesn't achieve that goal, I don't have to force using scipy.

examples/external_optimizers.py
```diff
@@ -258,6 +261,125 @@ def solve_newton_hvp(
     return _finish(model, "Newton (VMEC++ HVP + M^-1)", x, it, t0)
 
 
+def solve_newton_exact_hvp(input_path=DEFAULT_INPUT, ns=11, tol=1e-9, max_newton=80):
```
jurasic-pf: `solve_newton_exact_hvp` and `solve_newton_ptc` (below) each rebuild the same `h_op`/`m_op` `LinearOperator` pair already factored out as `_interior_operators` a few lines up (minus the `interior` subsetting). Worth a small shared `_hvp_operators(model, hvp_fn, n_dof, shift=0.0)` helper used by all three solvers?

examples/benchmark_adjoint_gradient.py
```diff
@@ -0,0 +1,100 @@
...
+
+DATA = Path(__file__).resolve().parent / "data"
+CASES = [
+    ("solovev", DATA / "solovev.json", 11),
+    ("cth_like", DATA / "cth_like_fixed_bdy.json", 11),
+]
+OBJ = va.mhd_energy
+
+
+def adjoint_grad(model, x_star, interior, boundary, exact):
```
jurasic-pf: This reimplements `vmecpp_adjoint.boundary_gradient`'s GMRES solve (same `hii`/`mii`/`gmres` shape) instead of calling `va.boundary_gradient(model, x_star, interior, boundary, OBJ, exact=exact)`, which already takes the same `exact` flag. Can we call the library function directly and drop the duplicate here?

krystophny:
Implemented the remaining exact-HVP review fixes: asymmetric constraint de-aliasing now uses the shared allocation-free kernel (including Enzyme workspace), the Newton examples use the hand-written GMRES/backtracking path with shared HVP operators, and the benchmark delegates to the public boundary-gradient routine. Bazel and pre-commit checks pass; the Enzyme/CMake rebuild is running separately.

krystophny:
Added direct `<Ax,y> = <x,A^T y>` checks for the hand-written 2D/3D geometry and force transforms and axis extrapolation, and kept those transpose operators available in non-Enzyme builds. The focused Bazel suite passes (2/2).

jurasic-pf (approved):

PR #583, opened 2026-06-14
ci: get the test + benchmark workflows green (VMEC2000-from-source, fork-PR benchmark guard)

krystophny:
Scope: CI only, standalone. No source or build-system changes. Independent of
the differentiable-VMEC++ PR stack (#564-#582); please review and merge this one
first so those PRs inherit green CI from `main`.

## What it fixes

**1. VMEC2000 compatibility test (`test_ensure_vmec2000_input_from_vmecpp_input`).**
The pinned `vmec-0.0.6` cp310 wheel was f90wrapped against numpy 1.x. The test
env now resolves numpy 2.x, under which importing that wheel dies in the f90wrap
array interface:

```
ValueError: _vmec._vmec.f90wrap_vmec_input__array__rbc:
  failed to create array ... 0-th dimension must be fixed to 2 but got 4
```

So the test could never actually run. (It is red on `main` too: the wheel's
runtime libs - libnetcdff, scalapack, openblas - are not installed there, and
pytest 9's `importorskip` only skips `ModuleNotFoundError`, so a missing `.so`
is a hard failure rather than a skip.)

Fix: build VMEC2000 from upstream source with current `f90wrap`, which produces
numpy-2-compatible bindings. The recipe mirrors SIMSOPT's own CI
(`hiddenSymmetries/VMEC2000`, `cmake/machines/ubuntu.json`). No fork needed. An
explicit `import vmec` check in the install step surfaces any remaining problem
there rather than as a confusing test failure.

**2. Benchmark workflow on fork PRs.** The result-upload step needs a writable
`GITHUB_TOKEN`; on fork PRs the token is read-only, so it failed with
"Resource not accessible by integration". Gate that step to same-repo PRs; the
benchmark itself still runs.

## Verification

Failing before (ubuntu-22.04, py3.10): `import vmec` chain broke on
libnetcdff -> libscalapack -> libopenblas, then on the f90wrap/numpy-2
`rbc` ValueError above.

Passing after: VMEC2000 builds from source, `import vmec` -> `VMEC2000 import OK`,
and the compatibility test runs. Benchmark workflow no longer errors on the
upload step for fork PRs.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,13 +60,6 @@ jobs:
       - name: Also install VMEC2000 (only on Ubuntu 22.04)
         if: ${{ matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.10' }}
         run: |
-          # The prebuilt vmec-0.0.6 wheel (anaconda eguiraud-pf, 2025-01-23) was
-          # built from VMEC2000 commit 728af8b ("works with numpy 2.0") but
-          # f90wrapped against that build's numpy; it no longer imports under the
-          # numpy 2.2 this CI resolves (the f90wrap array interface breaks). Build
-          # a wheel from the SAME pinned commit against the current numpy and
-          # cache it, so only the first run pays the build cost. Pinned and
-          # known-good like the old wheel, just numpy-2-correct.
           sudo apt-get install -y libopenmpi-dev libnetcdff-dev libscalapack-mpi-dev \
             libhdf5-dev libhdf5-serial-dev libfftw3-dev libopenblas-dev ninja-build
           python -m pip install mpi4py
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -53,12 +53,27 @@ jobs:
       - name: Also install VMEC2000 (only on Ubuntu 22.04)
         if: ${{ matrix.os == 'ubuntu-22.04' && matrix.python-version == '3.10' }}
         run: |
-          # mpi4py is needed for VMEC2000
-          sudo apt-get install -y libopenmpi-dev
-          python -m pip install mpi4py
-          # custom wheel for VMEC2000, needed for some VMEC++/VMEC2000 compatibility tests
-          # NOTE: this wheel is only guaranteed to work on Ubuntu 22.04
-          python -m pip install vmec@https://anaconda.org/eguiraud-pf/vmec/0.0.6/download/vmec-0.0.6-cp310-cp310-linux_x86_64.whl
+          # Build VMEC2000 from source instead of the old prebuilt wheel. The
+          # pinned vmec-0.0.6 cp310 wheel was f90wrapped against numpy 1.x and
+          # fails to import under numpy 2 (f90wrap array-interface break), so the
+          # compatibility test could never run. Building from source with current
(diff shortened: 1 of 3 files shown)


krystophny:
@jurasic-pf could you review this one first, independently of the differentiable-VMEC++ series?

It is CI-only and standalone (no source/build changes), and it gets the test + benchmark workflows green again on `main`:

- VMEC2000 compat test: the pinned `vmec-0.0.6` cp310 wheel is f90wrapped against numpy 1.x and no longer imports under numpy 2 (`f90wrap_vmec_input__array__rbc: 0-th dimension must be fixed to 2 but got 4`), so `test_ensure_vmec2000_input_from_vmecpp_input` could never run. This builds VMEC2000 from upstream source with current f90wrap (numpy-2 compatible), mirroring SIMSOPT's own CI recipe. No fork involved.
- Benchmark workflow: skip the result-upload step on fork PRs, where `GITHUB_TOKEN` is read-only.

Merge order: this PR first. Once it lands on `main`, the stacked series #564-#582 will rebase onto `main` and inherit the green CI; I deliberately did not cherry-pick these CI changes into that stack to keep this reviewable on its own.

.github/workflows/tests.yaml
```diff
@@ -50,15 +50,40 @@ jobs:
...
         run: |
-          # mpi4py is needed for VMEC2000
-          sudo apt-get install -y libopenmpi-dev
+          # The prebuilt vmec-0.0.6 wheel (anaconda eguiraud-pf, 2025-01-23) was
+          # built from VMEC2000 commit 728af8b ("works with numpy 2.0") but
+          # f90wrapped against that build's numpy; it no longer imports under the
+          # numpy 2.2 this CI resolves (the f90wrap array interface breaks). Build
+          # a wheel from the SAME pinned commit against the current numpy and
+          # cache it, so only the first run pays the build cost. Pinned and
+          # known-good like the old wheel, just numpy-2-correct.
```
jurasic-pf: Nit: This information will be available in the commit history, I think it's fine to remove the comment 
```suggestion
```

.github/workflows/tests.yaml
```diff
@@ -50,15 +50,40 @@ jobs:
...
+          sudo apt-get install -y libopenmpi-dev libnetcdff-dev libscalapack-mpi-dev \
+            libhdf5-dev libhdf5-serial-dev libfftw3-dev libopenblas-dev ninja-build
```
jurasic-pf: Unclear what FFTW is needed for, I'd prefer to avoid GPL licensed code where possible

jurasic-pf (approved):
Thanks! This really went unnoticed for way too long!

.github/workflows/tests.yaml
krystophny: elminated. was just part of the "all you need for vmec2000 build" but actually not needed.

krystophny:
Closing in favor of #564, which now carries all of these CI fixes plus
the Clang >= 21 abseil pin:

- build VMEC2000 from the pinned source commit `728af8b` and cache the
  wheel (canonical version, with the FFTW/HDF5 trim from this PR),
- skip the benchmark result upload on fork PRs,
- skip vmecpp-only INDATA fields in the compat test.

The review threads here are addressed in #564. Consolidating into one PR
so the Clang/Enzyme stack has a single build+CI prerequisite to land
first.

PR #584, opened 2026-06-14
geometry: expose the output-independent evaluator and VMEC adapter

krystophny:
Stack 1/4. Targets `main`.

## Scope

This PR adds the C++ geometry contract used by the rest of the stack:

- `Geometry` stores R, Z, lambda, toroidal flux, and poloidal flux in the product basis.
- `MakeGeometry` copies and rescales the physical state from `VmecInternalResults` without using wout.
- `EvaluateGeometry` returns values and the ten-entry spatial jet through second order.

SIMPLE PR itpplasma/SIMPLE#521 vendors its own ISO C adapter and consumes this API directly. It uses `Geometry`, `GeometryJet`, `MakeGeometry(..., kPhysical)`, and `EvaluateGeometry`. Those interfaces remain in this PR.

This layer adds no Python, pybind, C ABI, objective, or adjoint code.

## Tests

- `geometry_test`: analytic Fourier and radial fixtures for values and derivatives
- `vmec_geometry_test`: coefficient conversion, lambda scaling, and flux profiles from VMEC internal state

Stack: **#584** -> #703 -> #707 -> #710.


krystophny:
Merged current `main` through the stack into this PR.

Resolved the additive conflict in `src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/BUILD.bazel` by keeping the QS harmonics test target and the benchmark targets added on `main`.

Verification passed: `git diff --check`; CMake/Ninja build.

examples/data/cth_like_fixed_bdy.out.h5
jurasic-pf, line 1: i think this branch needs restacking, aren't these the same binary files you removed downstack?

krystophny:
Addressed in commits `86df88b3` and `b49ab353`: propagated the adjoint/Simsopt fixes and removed the generated `examples/data/*.out.h5` artifacts called out in the restacking comment. Targeted validation and pre-commit checks pass.

krystophny:
CI follow-up: `Run example scripts` fails in `examples/benchmark_adjoint_gradient.py` on `cth_like` because SciPy Newton–Krylov reaches `maxiter=80` (`max|F|` ≈ 1.6e-5). This is a numerical solver-policy question; no further change made pending your decision.

krystophny:
Clean SciPy implementation pushed in `715b1bd2`: public `root(..., jac_options=...)` with an adaptive VMEC `LinearOperator` preconditioner, LGMRES, and Armijo line search. No private SciPy internals. Focused tests, Python 3.10 benchmark, and pre-commit/Pyright pass; the same stack was also verified with the Enzyme-enabled exact-HVP benchmark.

krystophny:
CI follow-up: the Simsopt optimization example exposed a real stiff-step edge case in the public SciPy Krylov path. I added an initial-residual short-circuit and a safeguarded fallback using only public SciPy `LinearOperator`/`gmres`, VMEC’s HVP, and residual-decreasing backtracking. No SciPy private APIs are used. Locally, the full optimization example, focused tests, benchmark, and pre-commit pass. Pushed in commit `6f631c44`.

krystophny:
The derivative stack is now based on current main. PRs #582, #584, and #585 each pass all 31 checks, including TSAN. The QS replay and prior physics deltas remain unchanged. Ready for review.

Chris&GPT

jurasic-pf (changes requested):
Hi Chris, this is a very impactful demonstration, and the community probably wants to jump on this asap. I certainly know I do.

All the more I think getting the API right the first time around is important. This approach for implementing adjoints requires a significant amount of custom code for different objectives. This particular PR works for QS, but not for different implementations of QI, turbulence proxies, or anything else.
In that aspect it's similar to the implementation approach taken in E.J.Pauls adjoint VMEC paper https://arxiv.org/abs/2012.10028 https://github.com/ejpaul/ALPOpt , but maybe we can learn from that and improve.

Can we design the interface such that researchers can build whatever objectives they want on top of `vmecpp` and get adjoint gradients for free? The clearest path I see towards that is to wrap the a solve like function (probably not feasible to do for the full `vmecpp.run`) in a `custom_vjp` https://docs.jax.dev/en/latest/jep/2026-custom-derivatives.html . A QS objective built in JAX, and only taking geometry state variables  would then naturally propagate the vjp through to the `custom_vjp` boundary, from where enzyme takes over.

jurasic-pf:
If you want, I'd be happy to have a call this week to discuss details in person

krystophny:
Propagated the shared asymmetric constraint kernel and exact-HVP solver cleanup, and added `vmecpp.make_custom_vjp_solve` plus a behavioral JAX callback test so arbitrary objectives can compose over solve outputs. Bazel, pre-commit, and the JAX wrapper checks pass.

krystophny:
Also added direct `<Ax,y> = <x,A^T y>` coverage for the 2D/3D geometry and force transforms plus axis extrapolation, and made those hand-written transpose operators available in non-Enzyme builds. The focused Bazel transpose suite passes (2/2).

krystophny:
Restacked the whole series against your review. This PR is now **+722 / -0 across 11 files**, down from +1325 / -59 across 14.

What came out of it, all under "reduce code changes instead of adding unused features":

- **All pybind changes are gone from this PR.** They were unrelated to the geometry layer and are now in the PR whose client actually calls them (#703 for the geometry bindings, #710 for the solver ones).
- **Removed `EvaluateGeometryVjp` / `GeometryVjp`** (~130 lines with helpers and test). It had no consumer anywhere in the stack: the evaluation transpose is done in JAX in `vmecpp.geometry`, so the C++ one was dead surface. Easy to bring back if a client ever needs it.
- **Reverted ~100 lines of whole-file clang-format churn** in `pybind_vmec.cc` (an older clang-format had reflowed unrelated `Class &x` -> `Class& x` lines).
- Folded down the one-line `.coefficients = {}` designated-initializer completion that used to sit in #585.
- Rebased onto current `main`, so #716 (jxbout sign) and #711 are included rather than reverted.

Earlier inline comment about the binary `examples/data/*.out.h5` artifacts: resolved, none remain in the diff.

Verified: `bazel test //vmecpp/vmec/geometry:all` 2/2 pass; pre-commit clean including clang-format and pyright.

One thing I deliberately left out and want to flag rather than smuggle in: the `jcuru`/`jcurv` comparison in `test_pybind_vmec.py` sits right on its `1.0e-6` tolerance and has tripped on some optimized builds (worst observed 1.02e-6). That is pre-existing and unrelated to this PR, so it is not in the diff. If CI hits it here I will send it as its own small PR against `main`.

Your `CHANGES_REQUESTED` from 23 Aug is still standing on this one - could you re-review? The API objection behind it is answered in #707 and #710.

src/vmecpp/cpp/vmecpp/vmec/geometry/geometry.h
```diff
@@ -0,0 +1,64 @@
...
+  GeometryCoefficients coefficients;
+};
+
+// Value and derivatives with respect to (s, theta, zeta).
+//
+// The entries are ordered as value, ds, dtheta, dzeta, dss, ds_dtheta,
+// ds_dzeta, dtheta2, dtheta_dzeta, and dzeta2.  Keeping the jet in a fixed
+// array makes the C ABI and the JAX representation identical.
+inline constexpr int kGeometryJetSize = 10;
+using GeometryJet = std::array<double, kGeometryJetSize>;
```
jurasic-pf, line 49: Could you explain what a "geometry JET" is? I don't understand

src/vmecpp/cpp/vmecpp/vmec/geometry/geometry.h
```diff
@@ -0,0 +1,64 @@
...
+  std::vector<double> lambda_ss;
+};
+
+// Minimal equilibrium geometry independent of any output-file schema.
+struct Geometry {
+  GeometryDimensions dimensions;
+  std::vector<double> toroidal_flux;
+  std::vector<double> poloidal_flux;
+  GeometryCoefficients coefficients;
+};
```
jurasic-pf, line 41: Just to understand the design limitations a little better - why do we need these new classes, what makes the existing 
https://github.com/proximafusion/vmecpp/blob/main/src/vmecpp/cpp/vmecpp/vmec/fourier_geometry/fourier_geometry.h `FourierGeometry` class for example unsuitable? It also carries spans to the same coefficients. Do we specifically need value semantics? Does it have to be vectors for the public API instead of Eigen3? Is it to do with Enzyme?

src/vmecpp/cpp/vmecpp/vmec/geometry/geometry.cc
```diff
@@ -0,0 +1,259 @@
...
+namespace vmecpp {
+namespace {
+
+struct RadialWeights {
+  std::array<int, 4> indices;
+  std::array<double, 4> value;
+  std::array<double, 4> first;
+  std::array<double, 4> second;
+  int count;
+};
```
jurasic-pf, line 24: Could you help me understand what the meaning of 4 is here? What are these weights for 😅

src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.h
```diff
@@ -28,6 +28,9 @@ namespace vmecpp {
 struct VmecInternalResults {
   int sign_of_jacobian;
 
+  // Scale that converts the solver's lambda variable to physical lambda.
+  double lamscale;
+
```
jurasic-pf, line 33: I don't understand why we need this new quantity all of a sudden, especially for the HDF5 writing - isn't all the information to compute `constants_.lamscale = sqrt(constants_.rmsPhiP * fc_.deltaS);` available in principle?
Which Upstack PR needs it, I didn't quite follow

jurasic-pf (commented):
The PR is structurally good, if possible I'd prefer if we can reuse some of the existing classes to hand around the geometry, if there's a good reason why that doesn't work or makes the API worse, then we can merge. I don't understand the need for lamscale as an output yet.

jurasic-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/geometry/geometry.h
krystophny: Here “jet” means the value together with all first- and second-order partial derivatives at one point. For three coordinates that is 1 value + 3 first derivatives + 6 symmetric second derivatives = 10 entries; the ordering is documented immediately above the alias.

src/vmecpp/cpp/vmecpp/vmec/geometry/geometry.h
krystophny: `FourierGeometry` is tied to the solver representation (`Sizes`, `RadialPartitioning`, constrained/scaled coefficients, and solver-owned layout). This `Geometry` is an owning, self-contained physical snapshot that also carries the flux profiles and can cross the C++/Python/JAX/SIMPLE boundary without solver lifetime or layout dependencies. `std::vector` is for that portable owning API; this is not an Enzyme requirement.

src/vmecpp/cpp/vmecpp/vmec/geometry/geometry.cc
krystophny: Four is the maximum radial interpolation stencil size: for `ns >= 4` this uses four adjacent surfaces for cubic Lagrange interpolation and stores the value, first-derivative, and second-derivative weights. For `ns < 4` it falls back to the first two slots for linear interpolation.

src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.h
krystophny: `MakeGeometry` needs `lamscale / phipF[j]` to convert VMEC's internally scaled lambda coefficients to physical lambda. `lamscale` can in principle be recomputed from the full `phipH` profile and `deltaS`, but carrying the solver's actual value avoids duplicating that normalization and preserves the exact state in serialized `VmecInternalResults`. #703 consumes the resulting physical geometry; #710 consumes the same geometry in the solve VJP.

krystophny:
Restacked onto current `main` without changing the feature patch (stable patch ID unchanged). All review threads are answered and resolved; `bazel test //vmecpp/vmec/geometry:all` passes. Full CI is green except for the final Linux-distro wheel check still running. I do not have permission to merge into `proximafusion/vmecpp`; @jurasic-pf or @jons-pf, please merge when that last check completes.

PR #585, opened 2026-06-15
geometry: expose direct VMEC++ evaluation through ISO C

### LINKED ISSUE #592: Exact finite-difference-free quasisymmetry gradient for SIMSOPT
## User story

As a quasisymmetry-optimization user, I get dQS/dx with no finite differences, matching SIMSOPT's QuasisymmetryRatioResidual, so QS optimization converges with far fewer equilibrium solves.

## Scope

- `qs_harmonics_kernel.h`: flat-buffer forward transform of the field harmonics SIMSOPT QS reads, plus the analytic adjoint and tangent (the transform is linear in every field but `|B|`, which has a closed-form derivative).
- pybind: `qs_harmonics`, the reverse `exact_qs_objective_state_gradient` (O(1) in boundary DOFs), and the forward `exact_qs_harmonics_tangent` (the exact Gauss-Newton residual Jacobian).

## Benefit

Gauss-Newton QS optimization converges in about 64 equilibrium solves versus about 1288 with finite differences, matching SIMSOPT's objective bit-for-bit. Depends on the exact HVP (#582).

Closes when #585 merges.

krystophny:
Stack order: 3/6, targets `main`; merge after #584.

Adds the stable C-facing layer used by Fortran and other non-Python clients.

- creates an opaque geometry handle directly from a VMEC input file and runs VMEC++ in memory
- returns R, Z, lambda, toroidal flux, poloidal flux, and the common ten-entry spatial jet
- contains exceptions at the ABI boundary and reports thread-local errors
- installs the C header and exports the symbols from `vmecpp_core`

The end-to-end test runs VMEC++ and checks the LCFS against the input boundary Fourier series, independently of wout.


krystophny:
Stack refresh pushed in `bc754c46` and `bd355d87`: propagated the adjoint/Simsopt review fixes and removed the generated HDF5 artifacts from the downstream stack. Focused tests (4 passed) and changed-file pre-commit checks pass; full Pyright remains blocked by existing unrelated errors.

krystophny:
CI/local follow-up: the downstream `benchmark_adjoint_gradient.py` reproduces the same `cth_like` `maxiter=80` failure (`max|F|` ≈ 1.5e-5). The stack refresh is pushed, but solver tuning/strategy needs consideration before treating CI as green.

krystophny:
Clean SciPy implementation pushed in `401687e8`: public `root(..., jac_options=...)` with an adaptive VMEC `LinearOperator` preconditioner, LGMRES, and Armijo line search. No private SciPy internals. Focused tests, Python 3.10 benchmark, pre-commit/Pyright, and the Enzyme CTest checks pass; the Enzyme-enabled exact-HVP benchmark passes for both solovev and cth_like.

krystophny:
CI follow-up: the Simsopt optimization example exposed a real stiff-step edge case in the public SciPy Krylov path. I added an initial-residual short-circuit and a safeguarded fallback using only public SciPy `LinearOperator`/`gmres`, VMEC’s HVP, and residual-decreasing backtracking. No SciPy private APIs are used. Locally, the full optimization example, focused tests, benchmark, and pre-commit pass. Pushed in commit `6f3ac365`.

jurasic-pf (approved):

jurasic-pf (changes requested):
Hi Chris, this is a very impactful demonstration, and the community probably wants to jump on this asap. I certainly know I do.

All the more I think getting the API right the first time around is important. This approach for implementing adjoints requires a significant amount of custom code for different objectives. This particular PR works for QS, but not for different implementations of QI, turbulence proxies, or anything else.
In that aspect it's similar to the implementation approach taken in E.J.Pauls adjoint VMEC paper https://arxiv.org/abs/2012.10028 https://github.com/ejpaul/ALPOpt , but maybe we can learn from that and improve.

Can we design the interface such that researchers can build whatever objectives they want on top of `vmecpp` and get adjoint gradients for free? The clearest path I see towards that is to wrap the a solve like function (probably not feasible to do for the full `vmecpp.run`) in a `custom_vjp` https://docs.jax.dev/en/latest/jep/2026-custom-derivatives.html . A QS objective built in JAX, and only taking geometry state variables  would then naturally propagate the vjp through to the `custom_vjp` boundary, from where enzyme takes over.

krystophny:
Propagated the shared asymmetric constraint kernel and exact-HVP cleanup, and added the generic `vmecpp.make_custom_vjp_solve` interface with a behavioral JAX test. The example now composes a geometry-only QS proxy in JAX over the solve state, so other QI/turbulence/objective layers can use the same Enzyme-backed VJP. Focused JAX and pre-commit checks pass.

krystophny:
Added direct `<Ax,y> = <x,A^T y>` coverage for the 2D/3D geometry and force transforms plus axis extrapolation, and made those hand-written transpose operators available in non-Enzyme builds. The focused Bazel transpose suite passes (2/2).

jurasic-pf:
What do you need the C API for? I don't think it's needed for enabling vjp through vmecpp from the Python side, and until there's a clear usecase I'd prefer to reduce code changes instead of adding unused features

krystophny:
> What do you need the C API for? I don't think it's needed for enabling vjp through vmecpp from the Python side, and until there's a clear usecase I'd prefer to reduce code changes instead of adding unused features

To call it from Fortran or C codes without involving Python at all. But we can also keep a C++ API and I keep this layer myself. Whatever you prefer.

jurasic-pf:
If that's ok with you, I'd prefer to maintain just the C++ API and leave this adapter up to the user (you in this case)

krystophny:
Agreed - closing this. You maintain the C++ API, we maintain the adapter.

It is already done and working. The adapter now lives in SIMPLE (`itpplasma/SIMPLE#521`), which builds VMEC++ from source via `FetchContent` pinned to #584's commit `62be1ee`. Verified from deleted build directories: clean configure+build, both `test_vmecpp_field_parity` and `test_vmecpp_orbit_parity` pass, and the default build fetches and compiles no VMEC++ at all. The Fortran side needed no change - the C ABI is ours now, so it is ours to keep stable.

Note we are **not** asking you to install headers or export a CMake package. Building from source sidesteps that entirely, so nothing on your side has to grow to support us.

The one-line `.coefficients = {}` designated-initializer completion that lived here is folded down into #584.

Two things I hit doing this, both small and both useful to you independently of us:

1. **`vmecpp_core` has no `target_include_directories(... PUBLIC ...)`.** Include paths come from directory-scoped `include_directories()` calls, so they do not propagate to any target outside the vmecpp subtree. The link interface is already correct (`absl::*`, HDF5, netCDF, LAPACK, OpenMP, nlohmann_json are all PUBLIC) - it is only the include side that has to be hand-wired by a consumer.

2. **`EIGEN_MAX_ALIGN_BYTES=32` / `EIGEN_MAX_STATIC_ALIGN_BYTES=32` are set with directory-scoped `add_compile_definitions()`.** This one is a trap rather than an inconvenience: a consumer target that misses them compiles cleanly and then aborts at runtime with `double free or corruption (out)` in `Eigen::internal::aligned_free` when a `vmecpp::VmecINDATA` is destroyed. Your own comment in `CMakeLists.txt` already says this pin is ABI-affecting, so anything that can reach a vmecpp type across a TU boundary needs it.

Both would be fixed by putting the include dirs and those two definitions on the `vmecpp_core` target interface. Happy to send that as a small standalone PR if you want it - say the word and I will.

Separately, `_vmecpp` is built unconditionally with no option to skip it, so a C++-only consumer pays a pybind11 fetch and a Python-dev dependency for nothing. A `VMECPP_BUILD_PYTHON` option would be welcome, but that one is genuinely just convenience.

The design objection in your 23 Aug review is answered in #707 (QS as 126 lines of pure Python, no C++) and #710 (the generic solve VJP). The stack is now #584 -> #703 -> #707 -> #710.

PR #595, opened 2026-06-17
Make the iteration hot loop allocation-free

### LINKED ISSUE #594: Assert the core-loop is allocation free

CharlesCNorton:
Closes #594.

The per-iteration force evaluation allocated heap temporaries that scaled with problem size. Measured per steady-state iteration, interposing the whole malloc family so Eigen's aligned allocations are counted (not just `operator new`):

| case | before | after |
|------|-------:|------:|
| solovev (2D) | 38 | **0** |
| cma | 4016 | **0** |
| cth_like_fixed_bdy (3D) | 8558 | **0** |

## What allocated, and the fix

All become reused pre-sized scratch instead of per-call / per-iteration heap temporaries:

- **`computeMHDForces`** — 9 per-call and 8 per-surface `Eigen::VectorXd` → `ThreadLocalStorage` scratch.
- **`ForcesToFourier3DSymmFastPoloidal`** — the per-(surface, mode, zeta) `tempR/tempZ` materializations, which were the entire 3D bulk (~8.3k/iter) → per-thread scratch.
- **`applyRZPreconditioner`** — the `cR/cZ` vector-of-spans → stack `std::array` (`num_basis` ≤ 4).
- **`update`** — two size-3 residual `VectorXd` → scratch.
- **`Evolve`** — the 1/tau averaging-history left-shift used a `VectorXd` temporary → in-place forward copy.
- The convergence-history `push_back`s in `Evolve` → `reserve`d once per multigrid stage.

## Correctness

Storage reuse only, no arithmetic change. The converged `wout` is **bit-identical** before and after, compared array-by-array, on a 2D (`solovev`) and a 3D (`cth_like_fixed_bdy`) fixed-boundary case.

VMEC++ is only run-to-run deterministic single-threaded (two threaded runs of the same binary already differ at ~1e-6 in derived quantities like `currumnc`), so the bit-exact comparison is pinned to `OMP_NUM_THREADS=1`.

## Test

`VmecHotLoop.IsAllocationFree` interposes the malloc family, runs the same case to two iteration counts, and asserts the allocation delta over the extra steady-state iterations is zero. It pins to one thread and warms the per-thread scratch first. It is skipped under asan/tsan/msan, whose allocators conflict with the interposition.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
@@ -2067,8 +2067,11 @@ void IdealMhdModel::computeMHDForces() {
     m_ls_.gbvbv_i.setZero();
   }
 
-  // Reuse per-thread scratch instead of allocating these every call; every
-  // element is overwritten below (kl loops or setZero) before it is read.
+  // Persistent per-surface scratch (preallocated in ThreadLocalStorage):
+  // aliased here so every write below lands in existing storage. This keeps the
+  // force kernel allocation-free, which both avoids per-surface heap churn and
+  // lets Enzyme differentiate it (Enzyme cannot trace dynamic Eigen
+  // temporaries).
   Eigen::VectorXd& P_o = m_ls_.P_o;          //  r12 * totalPressure = P
   Eigen::VectorXd& rup_o = m_ls_.rup_o;      // ru12 * P
   Eigen::VectorXd& zup_o = m_ls_.zup_o;      // zu12 * P
@@ -2078,6 +2081,25 @@ void IdealMhdModel::computeMHDForces() {
   Eigen::VectorXd& gbubu_o = m_ls_.gbubu_o;  // gsqrt * bsupu * bsupu
   Eigen::VectorXd& gbubv_o = m_ls_.gbubv_o;  // gsqrt * bsupu * bsupv
   Eigen::VectorXd& gbvbv_o = m_ls_.gbvbv_o;  // gsqrt * bsupv * bsupv
+  P_o.setZero();
+  rup_o.setZero();
+  zup_o.setZero();
+  rsp_o.setZero();
+  zsp_o.setZero();
+  taup_o.setZero();
+  gbubu_o.setZero();
+  gbubv_o.setZero();
+  gbvbv_o.setZero();
+
+  // Surface-average scratch, likewise aliased to preallocated storage.
+  Eigen::VectorXd& P_avg = m_ls_.P_avg;
+  Eigen::VectorXd& P_wavg = m_ls_.P_wavg;
+  Eigen::VectorXd& gbubu_avg = m_ls_.gbubu_avg;
+  Eigen::VectorXd& gbubu_wavg = m_ls_.gbubu_wavg;
+  Eigen::VectorXd& gbvbv_avg = m_ls_.gbvbv_avg;
+  Eigen::VectorXd& gbvbv_wavg = m_ls_.gbvbv_wavg;
+  Eigen::VectorXd& gbubv_avg = m_ls_.gbubv_avg;
+  Eigen::VectorXd& gbubv_wavg = m_ls_.gbubv_wavg;
 
   for (int jF = r_.nsMinF; jF < jMaxRZ; ++jF) {
     const double sFull =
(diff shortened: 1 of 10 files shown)


jurasic-pf:
Hi Charles, that was fast!
I'll merge https://github.com/proximafusion/vmecpp/pull/566 which already adresses some of these changes first, simply because Chris opened it earlier, but looks great!

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec_allocation_test.cc
```diff
@@ -0,0 +1,160 @@
...
+  (void)CountHotLoopAllocations(indata, 30);
+
+  constexpr int kFewerIters = 40;
+  constexpr int kMoreIters = 80;
+  const uint64_t allocs_fewer = CountHotLoopAllocations(indata, kFewerIters);
+  const uint64_t allocs_more = CountHotLoopAllocations(indata, kMoreIters);
+
+  // The two runs differ only by (kMoreIters - kFewerIters) extra steady-state
+  // iterations; an allocation-free hot loop performs none.
+  EXPECT_EQ(allocs_more, allocs_fewer)
```
jurasic-pf, line 154: Amazing! 

can we add another test to this file that confirms the allocation, counting work correctly for a dummy function?

CharlesCNorton: @jurasic-pf complete

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -2970,8 +2972,10 @@ void IdealMhdModel::assembleRZPreconditioner() {
 // serial variant
 absl::Status IdealMhdModel::applyRZPreconditioner(
     FourierForces& m_decomposed_f) {
-  std::vector<std::span<double>> cR(s_.num_basis);
-  std::vector<std::span<double>> cZ(s_.num_basis);
+  // num_basis is at most 4 (cc, ss, sc, cs); a fixed-size array keeps this
+  // serial preconditioner path allocation-free.
+  std::array<std::span<double>, 4> cR{};
+  std::array<std::span<double>, 4> cZ{};
```
jurasic-pf, line 2978: Nice!

jurasic-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -832,7 +832,8 @@ absl::StatusOr<bool> IdealMhdModel::update(
                                         VacuumPressureState::kInitialized);
   bool includeEdgeRZForces =
       ((iter2 - iter1) < 50 && (almost_converged || hot_restart));
-  Eigen::VectorXd localFResInvar = Eigen::VectorXd::Zero(3);
+  Eigen::VectorXd& localFResInvar = m_ls_.fResInvar;
```
jurasic-pf: I didn't realize we were using a dynamic vector for this fixed size variable before!

Given this was a local variable, and fixed size,  I think stack allocating it with e.g. `Vector3d` or an `std::array` seems like a better solution here - we may need to adjust the function signature of `residuals`.

src/vmecpp/cpp/vmecpp/vmec/thread_local_storage/thread_local_storage.h
```diff
@@ -72,6 +72,11 @@ class ThreadLocalStorage {
   Eigen::VectorXd gbvbv_wavg;  // 0.5 (gbvbv_o sqrtSHo + gbvbv_i sqrtSHi)
   Eigen::VectorXd gbubv_avg;   // 0.5 (gbubv_o + gbubv_i)   [3D only]
   Eigen::VectorXd gbubv_wavg;  // 0.5 (gbubv_o sqrtSHo + gbubv_i sqrtSHi) [3D]
+
+  // Size-3 (R, Z, lambda) force-residual accumulators, reused in
+  // IdealMhdModel::update (formerly per-call heap temporaries).
+  Eigen::VectorXd fResInvar;
+  Eigen::VectorXd fResPrecd;
```
jurasic-pf: Since this is fixed size, we can stack allocate and keep the scope limited to where it's needed (comment above).


```suggestion
```

src/vmecpp/cpp/vmecpp/vmec/thread_local_storage/thread_local_storage.cc
```diff
@@ -56,5 +56,7 @@ ThreadLocalStorage::ThreadLocalStorage(const Sizes* s) : s_(*s) {
   gbvbv_wavg.setZero(nZnT);
   gbubv_avg.setZero(nZnT);
   gbubv_wavg.setZero(nZnT);
+  fResInvar.setZero(3);
+  fResPrecd.setZero(3);
```
jurasic-pf:
```suggestion
```

jurasic-pf (approved):
Very nice! Thanks!

jurasic-pf:
Could you still address the comments before merge?

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
CharlesCNorton: Done in 124803e. Both residual accumulators are now stack Eigen::Vector3d locals scoped to where they are used, and residuals(), evalFResInvar() and evalFResPrecd() take Eigen::Vector3d so no temporary VectorXd is materialized. The ThreadLocalStorage members are dropped.

src/vmecpp/cpp/vmecpp/vmec/thread_local_storage/thread_local_storage.h
CharlesCNorton: Done in 124803e, now a stack Eigen::Vector3d local.

src/vmecpp/cpp/vmecpp/vmec/thread_local_storage/thread_local_storage.cc
CharlesCNorton: Done in 124803e.

CharlesCNorton:
Thanks! Addressed the inline comments in 124803e: the size-3 (R, Z, lambda) residual vectors are stack Eigen::Vector3d locals now and out of ThreadLocalStorage, and residuals(), evalFResInvar() and evalFResPrecd() take Eigen::Vector3d so no temporary VectorXd is allocated. The allocation-counting test was added earlier.

PR #598, opened 2026-06-26
Make the lambda Fourier resolution independent of the geometry

### LINKED ISSUE #524: Lambda resolution independent of `mpol`, `ntor` geometric resolution
We know that `lambda` may contain higher spectral content than is required to represent the boundary effectively. VMEC then needs higher resolution to converge well. We could make the resolutions independent.

CharlesCNorton:
Closes #524.

## Motivation

Lambda can carry higher spectral content than the boundary geometry needs. Today the only way to give lambda that resolution is to raise `mpol`/`ntor` for the whole state, which over-resolves R and Z as well. This adds two optional inputs, `mpol_geometry` and `ntor_geometry`, that cap the geometry resolution below the working `mpol`/`ntor` while lambda keeps the full `mpol`/`ntor`. Unset (the default), or set equal to `mpol`/`ntor`, they change nothing.

## Approach

The working resolution stays `mpol`/`ntor`, so every array, transform, and the lambda spectrum are unchanged; only the geometry (R, Z) is constrained. The boundary is zero-padded to `mpol`/`ntor` on input, so the geometry modes above the cap start at zero, and two masks hold them there:

- the R/Z spectral force above the cap is zeroed every iteration, before the invariant residual (so the frozen modes stay out of the convergence criterion) and before preconditioning (so they receive no update);
- the geometry state above the cap is zeroed at the top of each iteration, so a bad-Jacobian axis re-guess, which writes geometry modes directly, cannot reintroduce them.

Lambda's coefficients are never touched, so it is resolved at the full `mpol`/`ntor`.

## Changes

- `vmec_indata`: `mpol_geometry` / `ntor_geometry` inputs (default `-1` = use `mpol`/`ntor`), with JSON and HDF5 read/write.
- `sizes`: derive `mpolGeometry` / `ntorGeometry`, clamped so an unset or out-of-range value equals `mpol`/`ntor`.
- `fourier_coefficients`: `maskGeometryAbove(mpolGeom, ntorGeom)` zeros R/Z coefficients with `m >= mpolGeom` or `n > ntorGeom`, leaving lambda.
- `ideal_mhd_model`: mask the R/Z force after it is formed and again after preconditioning, and mask the geometry state at the top of `update`.
- `pybind_vmec.cc` / `__init__.py`: expose the two inputs.

## Validation

Unset, or set equal to `mpol`/`ntor`, the caps reproduce the un-capped `wout` array-for-array (the committed no-op test).

With the geometry capped below the working resolution, the R/Z coefficients above the cap are held at exactly zero while lambda is resolved above it. On an nfp=5 boundary run at `mpol=8`, capping the geometry at 6 converges in 926 iterations (`fsqr` 1e-11) with every R/Z mode above m=6 exactly zero and all 22 lambda modes above m=6 populated; reaching the same lambda resolution the old way, `mpol=8` throughout, takes 1037 iterations with the geometry over-resolved.

The geometry-state mask is what makes the cap hold through a bad-Jacobian axis re-guess; on clean runs it is a no-op, and a capped run is bit-identical with and without it. `test_init.py` is unchanged (the same pre-existing source-run failures with and without this change).

## Test

`tests/test_lambda_resolution.py` drives cma with the geometry capped below its resolution:
- a cap equal to the working resolution reproduces the un-capped `wout` array-for-array (`mpol_geometry`/`ntor_geometry` are a no-op at full resolution);
- below the cap, every R and Z coefficient is exactly zero while lambda carries non-trivial content, confirming the geometry is frozen and lambda is resolved independently.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -175,6 +175,20 @@ class VmecInput(BaseModelWithNumpy):
     """Number of toroidal Fourier harmonics; n = -ntor, -ntor+1, ..., -1, 0, 1, ...,
     ntor-1, ntor."""
 
+    mpol_geometry: int = -1
+    """Optional reduced poloidal resolution for the geometry (R, Z).
+
+    If in [1, mpol), R/Z modes with m >= mpol_geometry are held fixed while lambda keeps
+    the full mpol. < 0 (default) means geometry uses mpol.
+    """
+
+    ntor_geometry: int = -1
+    """Optional reduced toroidal resolution for the geometry (R, Z).
+
+    If in [0, ntor), R/Z modes with n > ntor_geometry are held fixed while lambda keeps
+    the full ntor. < 0 (default) means geometry uses ntor.
+    """
+
     ntheta: int = 0
     """Number of poloidal grid points (ntheta >= 0).
 
--- a/src/vmecpp/cpp/vmecpp/common/sizes/sizes.cc
+++ b/src/vmecpp/cpp/vmecpp/common/sizes/sizes.cc
@@ -14,13 +14,19 @@
 namespace vmecpp {
 
 Sizes::Sizes(const VmecINDATA& id)
-    : Sizes(id.lasym, id.nfp, id.mpol, id.ntor, id.ntheta, id.nzeta) {}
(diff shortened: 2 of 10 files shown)


jurasic-pf (approved):
Very helpful! This was straightforward to review.

PR #600, opened 2026-06-26
vmec_indata: warn when the input boundary is spectrally dense

### LINKED ISSUE #418: Warn when input boundary is spectrally dense
Fixed boundary runs can produce poor results, when the boundary has high spectral content. This can commonly be resolved with spectral condensation.
A heuristic that detects if this might become a problem and emits a corresponding warning message, would be helpful to users.

CharlesCNorton:
Closes #418.

## Motivation

A fixed-boundary run whose boundary carries a lot of high-poloidal-mode content tends to fail to initialize or converge poorly. When it fails, the only signal the user gets is the late, generic abort

```
FATAL ERROR in thread=0. The solver failed during the first iterations. This may
happen if the initial boundary is poorly shaped or if it isn't spectrally
condensed enough.
```

which fires after the solve has already started and does not say that the boundary itself is the culprit. This adds a cheap up-front check that flags a spectrally dense boundary before the solve, pointing at the usual fix (spectral condensation).

## Approach

`IsConsistent` now computes the spectral width `<M>` of the input boundary and warns when it exceeds a threshold. `<M>` is the same quantity VMEC++ already reports as the `<M>` iteration-table column and stores as `wout.specw`: the energy-weighted average poloidal mode number defined in `FourierGeometry::ComputeSpectralWidth` (`p=4`, `q=1`, `m=0` excluded, `nscale` weighting), evaluated here on the input `rbc`/`zbs`/`rbs`/`zbc` instead of on a converged surface.

The warning is gated so it reaches the right audience without becoming noise:

- on `enable_info_messages`, which is `verbose_` on the run path (true for every run except an explicitly silenced one) and `false` on the JSON-parse path, so it fires once per run and never on serialization or in a silenced optimization loop;
- on `!lfreeb`: a free-boundary run relaxes the initial boundary, so the same content is far less harmful and is not flagged.

A dense boundary is still a valid input, so this is a `LOG(WARNING)`, not an error.

## Calibration

The threshold is set from ripple sweeps: starting from a converging case, a flat high-m boundary ripple of growing amplitude was added and each input run to completion, recording the input `<M>` at which the solve stops converging.

| base case | converges up to | fails from |
|------|-------:|------:|
| tokamak, single high-m mode | `<M>` = 4.2 | 7.2 |
| tokamak, high-m band | `<M>` = 5.5 | 7.7 |
| 3D stellarator (nfp=5) | `<M>` = 5.9 | 7.8 |

Three very different spectra cross from converging to failing in the same `<M>` band, and all 12 bundled test inputs sit at `<M> <= 1.75` (`cma` is the densest at 1.74). The threshold is `6.0`: above every converged run in the sweeps, below every failure, and far above any bundled case. It is a single named constant (`kSpectrallyDenseBoundaryThreshold`) so it can be retuned.

## Changes

- `vmec_indata.h` / `vmec_indata.cc`: `BoundarySpectralWidth(const VmecINDATA&)`, mirroring `ComputeSpectralWidth` on the input boundary coefficients, and the `kSpectrallyDenseBoundaryThreshold` constant.
- `vmec_indata.cc`: in `IsConsistent`, warn when a fixed-boundary input exceeds the threshold.

## Validation

Run end-to-end through `vmec_standalone`, the warning fires exactly on the inputs that go on to fail and is silent otherwise: the clean tokamak (`<M>` = 1.0) and a below-threshold dense ripple (`<M>` = 5.5, which converges) produce no warning, while `<M>` = 7.7 and `<M>` = 8.7 each emit the warning and then hit the abort above. Because the check runs in `IsConsistent`, the warning prints before the solve, so the user sees it ahead of the failure rather than after.

`BoundarySpectralWidth` reproduces the calibrated `<M>`: 1.02 for `solovev`, 9.01 for the dense fixture below. No bundled input exceeds the threshold, so the existing suite emits no new warnings.

## Test

`vmec_indata_test.cc` adds two cases:

- `BoundarySpectralWidthCleanBoundaryIsLow`: `solovev`'s smooth boundary gives `1 <= <M> < threshold`;
- `BoundarySpectralWidthDenseBoundaryExceedsThreshold`: injecting an `m = 8..10` ripple drives `<M>` above the threshold, and `IsConsistent` still accepts the input (warning, not rejection).


jurasic-pf:
Non convergence is just the most extreme form of failiure mode, physics errors like https://github.com/proximafusion/vmecpp/issues/671 can appear at smaller spectral width already (3.5 in this case). Since the warning is non-blocking, let's lower the threshold to 2.0

src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
```diff
@@ -1216,6 +1216,40 @@ std::string VmecINDATA::ToJsonOrException() const {
...
+      // nscale[|n|]^2 from the Fourier basis normalization: n=0 weighted 1,
+      // n!=0 weighted 2 (matches ComputeSpectralWidth).
+      const double n_weight = (n == 0) ? 1.0 : 2.0;
+      energy *= n_weight;
+      numerator += energy * m5;
+      denominator += energy * m4;
+    }
+  }
+  if (denominator <= 0.0) {
+    return 0.0;
```
jurasic-pf: Why did this logic have to be duplicated? I would strongly prefer if we could reuse the existing spectral width implementation instead, even if it requires some more complicated wiring.

jurasic-pf (changes requested):
Thanks, I think the change is in the right direction, but lets use a more conservative threshold and clean up the code a bit

src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
```diff
@@ -1625,6 +1659,23 @@ absl::Status IsConsistent(const VmecINDATA& vmec_indata,
...
+  // runs relax the initial boundary, so the same content is less harmful and is
+  // not flagged here.
+  if (enable_info_messages && !vmec_indata.lfreeb) {
+    const double boundary_spectral_width = BoundarySpectralWidth(vmec_indata);
+    if (boundary_spectral_width > kSpectrallyDenseBoundaryThreshold) {
+      LOG(WARNING) << absl::StrFormat(
+          "Input boundary is spectrally dense (spectral width <M> = %.1f). "
+          "Fixed-boundary runs with high spectral content may converge poorly; "
+          "consider spectral condensation of the boundary or reducing its "
+          "high-m Fourier content.",
```
jurasic-pf:
```suggestion
          "Input boundary is spectrally dense (spectral width <M> = %.1f). "
          "Runs with high spectral content may converge poorly or incorrectly resolve the physics.  "
          "consider spectral condensation of the boundary first, e.g. using simsopt.geo.surfacerzfourier.condense_spectrum",
```

CharlesCNorton:
Both done in 3302aa31.

Threshold is now 2.0. The densest bundled input is cma at 1.74, so nothing shipped trips it.

The duplication is gone: the per-surface computation moved out of `FourierGeometry::ComputeSpectralWidth` into `common/fourier_basis`, and both the geometry and a new `Boundaries::ComputeSpectralWidth` call it. It could not stay in `IsConsistent` -- `vmec_indata` sits below `sizes` and `boundaries` in the dependency graph, so reaching the existing implementation from there is a build cycle. The warning now fires at the end of `Boundaries::setupFromIndata`, still before the solve and on the same `verbose` gate.

`output_quantities_test` passes unchanged, so `specw` is unaffected by the extraction.

jurasic-pf (approved):
Thanks!

PR #601, opened 2026-06-27
Sparse Fourier coefficients via extra geometry modes and sparse_lambda

### LINKED ISSUE #315: Sparse Fourier coefficients
Only consider non-zero Fourier coefficients in the computation when assembling the matrix. 
This would allow to cheaply explore high frequency perturbations to an equilibrium, or global perturbations (modes that are not multiples of NFP)

CharlesCNorton:
Closes #315.

## Motivation

#315 asks to "only consider non-zero Fourier coefficients in the computation," to cheaply explore high-frequency perturbations to an equilibrium and global perturbations whose toroidal mode numbers are not multiples of NFP.

Today the mode set is the dense grid: poloidal `m` in `[0, mpol)`, toroidal `n` in `[-ntor, ntor]`, with the toroidal basis hard-wired to `n*nfp`. Adding a single high-`m` mode, or a non-axisymmetric mode at `nfp=1`, forces the full dense grid and pays for every intermediate mode even though it stays at (or near) zero.

This adds an opt-in sparse mode set: the geometry (and, with `sparse_lambda`, lambda) is restricted to an explicit set of free modes, and the Fourier transforms iterate only those modes. A reduced-resolution base equilibrium can then carry a few explicit high-frequency or global modes, computed at a fraction of the dense cost, while everything else is frozen at zero.

## Approach

Two inputs build on the geometry masking from #598 (`mpol_geometry`/`ntor_geometry` freeze R/Z above a cap):

- `extra_geometry_m` / `extra_geometry_n`: an explicit list of `(m, n)` geometry modes kept free *above* the cap. The active geometry set is the cap rectangle plus these extras; everything else is frozen at zero each iteration (force and state), so a single perturbation mode can be added to a converged base and studied in isolation.
- `sparse_lambda`: also restricts lambda to the active set, which lets the transforms skip the frozen modes. The frozen modes are exactly zero, so skipping their work is identical to the dense transform:
  - whole poloidal rows with no free mode are skipped in the inverse and forward DFTs (2D and 3D);
  - the toroidal sums run only up to the highest free `n` at each `m`.

  The preconditioner still masks the frozen modes to zero (correct, computed-then-discarded); skipping its per-mode solve as well is left out because a frozen mode's force is already zero, so it contributes nothing.

Global, non-NFP perturbations are reached by running `nfp=1` (the full torus, where every integer `n` is a valid toroidal mode) with a sparse mode set, so the cost of the otherwise-dense high-`ntor` grid is avoided.

When neither input is set (the default), nothing is masked and the transforms run the dense path unchanged.

## Changes

- `vmec_indata`: `extra_geometry_m` / `extra_geometry_n` and `sparse_lambda` inputs, with JSON and HDF5 read/write and an `extra_geometry_*` length-consistency check.
- `sizes`: derive the active-mode mask and `isActiveGeometryMode` / `anyActiveGeometryAtM` / `maxActiveGeometryNAtM` helpers.
- `fourier_coefficients`: `maskGeometryAbove` keeps the extra modes and, under `sparse_lambda`, also freezes lambda.
- `dft_toroidal` / `ideal_mhd_model`: the inverse and forward symmetric DFTs (2D and 3D) skip frozen poloidal rows and shorten the toroidal sums under `sparse_lambda`.
- `pybind_vmec.cc` / `__init__.py`: expose the three inputs.

## Validation

Skipping the frozen modes is bit-identical to the dense (masked) transform: on solovev with a geometry cap and a kept `m=4` extra, the converged MHD energy is `2.546537` with and without the poloidal-row skip; on an `nfp=1` run with a global `(m=1, n=1)` mode it is `2.549049` with and without the toroidal-sum shortening.

Cost, single grid `ns=51`, `mpol=60`, 600 iterations: the dense run takes `0.75 s`; the same base with the geometry capped at `m<3` and `sparse_lambda` takes `0.49 s` (the transform no longer touches the 57 frozen rows; the remaining cost is the real-space physics on the grid, which a high mode still requires).

The non-NFP capability runs end to end: a global `n=1` mode on an axisymmetric base at `nfp=1`, kept as a sparse extra, converges to `ftol = 1e-11` in 197 iterations.

Default (no extras, `sparse_lambda=False`) is a no-op: the #598 `test_lambda_resolution.py` cases, including the cap-at-full-resolution array-for-array equality, are unchanged.

## Test

`tests/test_extra_geometry_modes.py`:
- `test_extra_mode_kept_above_geometry_cap` / `test_only_listed_extra_modes_are_kept`: a high-`m` perturbation is frozen to zero under the cap, kept (matching the full-resolution run) when listed as an extra mode, and only the listed modes are kept;
- `test_sparse_lambda_freezes_lambda_above_cap`: `sparse_lambda` freezes lambda above the cap to exactly zero, where it otherwise carries content;
- `test_non_nfp_global_perturbation_runs_sparsely`: an `nfp=1` global `(m=1, n=1)` mode is kept while the frozen `n=2` modes stay at zero.


jurasic-pf:
Hi! Thanks for looking into another of the open issues. In this particular case, there were already a few PRs linked investigating it e.g. https://github.com/proximafusion/vmecpp/pull/548 and I see that we arrived at slightly different solutions.
Active mode masking in the DFT is similar, but you e.g. opted for additional input fields that control the sparsity. Could you elaborate on why one over the other?

CharlesCNorton:
@jurasic-pf I opted for explicit fields so the active mode set can be prescribed directly, which lets us retain modes that are zero in the base equilibrium when setting up perturbation studies. Happy to pick and choose from the different features if you want them folded into your drafts.

PR #606, opened 2026-07-06
Add C++ Google Benchmark microbenchmarks for critical hot functions

### LINKED ISSUE #599: Add microbenchmarks of critical kernels to CI
We currently have multiple end to end integration tests in our benchmarks for free and fixed boundary cases. 
something similar can be set up for benchmarking individual, critical C++ functions using googlebenchmark https://github.com/google/benchmark as part of our coverage suite.
It should be supported by the ci workflow we use for reporting performance regressions.

Functions that should be covered by this 
- Dft and inv dft for different sizes (including some of the fft paths) 
- dealias force constraint
- free boundary components like the laplace solve
- output quantity computation (how much time do we spend after the actual solve has already finished)

jurasic-pf:
## What

Adds four C++ [Google Benchmark](https://github.com/google/benchmark) `cc_binary` microbenchmarks covering the critical hot functions, complementing the existing Python end-to-end benchmark suite. Individual-function timings let a regression be attributed to a specific kernel rather than only showing up in a whole-solve number.

| Area | Target | Notes |
|------|--------|-------|
| DFT/FFT transforms | `//vmecpp/vmec/ideal_mhd_model:fft_toroidal_bench` | Restores the previously-dropped BUILD target for the existing `fft_toroidal_bench.cc` |
| Constraint-force de-aliasing | `//vmecpp/vmec/ideal_mhd_model:dealias_constraint_force_bench` | `deAliasConstraintForce` |
| Free-boundary Laplace solve | `//vmecpp/free_boundary/laplace_solver:laplace_solver_bench` | Assemble + LU factorize + back-substitute, plus the Green's-function-derivative transform |
| Post-solve output | `//vmecpp/vmec/output_quantities:output_quantities_bench` | `ComputeOutputQuantities` — time spent after the solve has converged |

## Bug fixed along the way

The orphaned `fft_toroidal_bench.cc` segfaulted: `FourierToReal3DSymmFastPoloidalFft` dereferences a null FFTX function pointer at resolutions with no vendored codelet. The real solver dispatches on `plans.kernels_available()`; the FFT-path benchmarks now do the same guard and `SkipWithError` cleanly. (This is likely why the target was originally dropped from `BUILD.bazel`.)

## CI

Extends the shared `run-benchmarks` composite action (introduced in the base branch) to also build, run, and store the C++ suite via `benchmarks/run_cpp_benchmarks.sh`, keeping a single source of truth for the benchmark environment. No new jobs in the workflows — both callers get the C++ suite for free.

Results are stored under a distinct `C++ Microbenchmarks` suite name in the same `benchmark-runs` data dir, so they render as a separate chart group on the docs dashboard **without overwriting the Python `Benchmark` history**. The two `github-action-benchmark` store steps run sequentially within the composite action, so the `auto-push` writes to `benchmark-runs` never race.


jurasic-pf:
* **#606** <a href="https://app.graphite.com/github/pr/proximafusion/vmecpp/606?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/proximafusion/vmecpp/606?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/proximafusion?ref=merge-queue-instructions-in-stack-comment&prId=606).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #607, opened 2026-07-07
Fix current density in magnetic field visualization example

WenyinWei:
## Summary

This PR fixes `examples/visualize_magnetic_field.py` so that the current density is reconstructed using the VMEC current-harmonic convention.

The magnetic-field part of the example was already using contravariant components `B^theta` and `B^zeta`. The current-density part, however, treated `currumnc/currvmnc` as if they were physical contravariant current components. They are not: VMEC stores Fourier coefficients of `sqrt(g) J^theta` and `sqrt(g) J^zeta` on the Nyquist basis.

The example now:

- reconstructs `sqrt(g)` from `gmnc` and optional `gmns`
- reconstructs `sqrt(g)J^theta` and `sqrt(g)J^zeta` from `currumnc/currvmnc` and optional sine terms
- divides by `sqrt(g)` before converting the contravariant current to cylindrical/Cartesian components
- raises a clear `ZeroDivisionError` if the reconstructed Jacobian is zero

A focused unit test is added to verify the Jacobian division with a one-mode toy surface.

## Why this matters

Without the division by `sqrt(g)`, the example transforms `sqrt(g)J` as though it were `J`. That changes the current magnitude and spatial pattern shown to users of the example, especially for finite-beta equilibria where the Jacobian varies over the surface.

## Public finite-beta diagnostic

For a public finite-beta W7-X reference wout, the before/after current magnitude changes visibly when `sqrt(g)` is handled correctly.

The diagnostic used this public file from SIMSOPT:

```text
tests/test_files/wout_W7-X_without_coil_ripple_beta0p05_d23p4_tm_reference.nc
```

with:

- `beta_total = 0.04484`
- `nfp = 5`
- selected full-grid surface index `150/200`

VMEC++ current-density before/after

For this selected surface, the old/new magnitude ratio has:

- median `7.08e-01`
- p05 `5.27e-01`
- p95 `9.90e-01`

The exact numbers depend on the selected surface, but the failure mode is invariant: the old example was visualizing `sqrt(g)J`, not physical `J`.

## Reproducer

One way to reproduce the diagnostic is:

```bash
curl -L -o /tmp/wout_W7X_beta0p05.nc \
  https://raw.githubusercontent.com/hiddenSymmetries/simsopt/master/tests/test_files/wout_W7-X_without_coil_ripple_beta0p05_d23p4_tm_reference.nc

python3 - <<'PY'
from scipy.io import netcdf_file
import numpy as np

path = "/tmp/wout_W7X_beta0p05.nc"
with netcdf_file(path, "r", mmap=False) as nc:
    nfp = int(nc.variables["nfp"].data.copy())
    ns = int(nc.variables["ns"].data.copy())
    beta = float(nc.variables["betatotal"].data.copy())
    mnmax = int(nc.variables["mnmax"].data.copy())
    mnmax_nyq = int(nc.variables["mnmax_nyq"].data.copy())
    curr_shape = nc.variables["currvmnc"].data.shape
    gm_shape = nc.variables["gmnc"].data.shape

print("beta_total:", beta)
print("nfp, ns:", nfp, ns)
print("mnmax, mnmax_nyq:", mnmax, mnmax_nyq)
print("gmnc shape:", gm_shape)
print("currvmnc shape:", curr_shape)
print("current harmonics are sqrt(g)J on the Nyquist basis, so the example must divide by sqrt(g).")
PY
```

Expected metadata:

```text
beta_total: 0.04484065...
nfp, ns: 5 201
mnmax, mnmax_nyq: 200 841
gmnc shape: (201, 841)
currvmnc shape: (201, 841)
```

## Test

```bash
pytest -q tests/test_visualize_magnetic_field_example.py
```

Local result:

```text
1 passed
```


jurasic-pf:
Thanks! This looks correct, the current contribution was missing a sqrt factor.

jurasic-pf (approved):

examples/visualize_magnetic_field.py
```diff
@@ -87,12 +88,33 @@ def calculate_magnetic_field(
...
+    if gmns is not None and gmns.size:
+        sqrtg += np.dot(gmns[:, j], sink_nyq)
+    if currumns is not None and currumns.size:
+        curru_sqrtg += np.dot(currumns[:, j], sink_nyq)
+    if currvmns is not None and currvmns.size:
+        currv_sqrtg += np.dot(currvmns[:, j], sink_nyq)
+
+    if np.isclose(sqrtg, 0.0):
+        msg = "Cannot reconstruct current density where the VMEC Jacobian is zero."
+        raise ZeroDivisionError(msg)
```
jurasic-pf: A zero jacobian means the VMEC solution is invalid anyway. I'd remove it for the sake of a simpler example script

WenyinWei:
Thanks for checking. I agree the missing `sqrt(g)` factor was the central issue here. The PR CI is green on the example script, Python tests, type checks, and C++ sanitizer jobs, so from my side this is ready whenever you are comfortable merging it.

examples/visualize_magnetic_field.py
jurasic-pf:
```suggestion

```

examples/visualize_magnetic_field.py
```diff
@@ -87,12 +88,33 @@ def calculate_magnetic_field(
...
+    # coordinates.
+    gmnc = vmec_output.wout.gmnc
+    gmns = getattr(vmec_output.wout, "gmns", None)
     currumnc = vmec_output.wout.currumnc
     currvmnc = vmec_output.wout.currvmnc
-
-    curru = np.dot(currumnc[:, j], cosk_nyq)
-    currv = np.dot(currvmnc[:, j], cosk_nyq)
+    currumns = getattr(vmec_output.wout, "currumns", None)
+    currvmns = getattr(vmec_output.wout, "currvmns", None)
```
jurasic-pf:
```suggestion
    currumns = vmec_output.wout.currumns
    assert currumns is not None
    currvmns = vmec_output.wout.currvmns
    assert currvmns is not None
```

tests/test_visualize_magnetic_field_example.py
jurasic-pf, line 1: Thanks for adding the test, but since it just makes assertions about the implementation, and doesn't assert any physical invariant outside of that, I'd remove it.

PR #611, opened 2026-07-08
Merge fast_poloidal and fast_toroidal Fourier bases into one template

### LINKED ISSUE #217: Merge fast_toroidal and fast_poloidal implementations, remove redundant logic
Low-priority clean-up task.

CharlesCNorton:
Resolves #217.

`FourierBasisFastPoloidal` and `FourierBasisFastToroidal` were near-identical: same public interface, same arithmetic, differing only in the flat memory layout of their basis and coefficient arrays (theta-fast / m-major for the VMEC++ core solver, zeta-fast / n-major for Nestor). The toroidal class even carried a `// TODO(eguiraud) reduce overall code duplication`.

## What

- The two classes collapse into one `FourierBasis<Layout>` template holding the single implementation. Two small policy structs, `FourierBasisFastPoloidalLayout` and `FourierBasisFastToroidalLayout`, supply the only thing that actually differed: the three flat index formulas (poloidal-basis, toroidal-basis, product-basis). This is zero runtime cost, since the layout is a compile-time policy and the indices inline exactly as before.
- `FourierBasisFastPoloidal` and `FourierBasisFastToroidal` become type aliases of the two instantiations, so every call site, both unit tests, and both data layouts are unchanged.
- The two old headers become one-line forwarders onto the new `common/fourier_basis` package (stable include paths); the duplicated `.cc` files are removed. Net -585 lines.

## Verification

Both layouts are bit-for-bit identical to the previous code:

- Fixed-boundary solve (solovev, exercises the m-major FastPoloidal path): volume `126.8719276452`, unchanged.
- Free-boundary solve (solovev + Nestor, exercises the n-major FastToroidal path): volume `128.8377807690221`, betatotal `4.1817252518e-06`, identical to the pre-refactor code to all printed digits.
- The full library builds with both instantiations linked into the core solver and the free-boundary code; clang-format clean.


jurasic-pf:
Nice! The indexing solution is elegant, just waiting for the benchmark job to pass then we can merge

jurasic-pf (approved):

PR #612, opened 2026-07-08
Honor iteration_style=parvmec in the native solver

### LINKED ISSUE #157: Make PARVMEC iteration method available in VMEC++
(Raised by @jons-pf ) 

PARVMEC uses a slightly different iteration method (when to perform a time step, when to backup, ...) than serial VMEC 8.52/educational_VMEC/VMEC++.
I had implemented a prototype in #1507, and it worked to the degree that the number of iterations then matched PARVMEC.
We also saw that we have cases where PARVMEC fails to converge and gets stuck early on, whereas VMEC++ can currently handle these cases. The only downside is that a few more iterations (2900 vs 2700) are needed. For now, this is considered acceptable.

In the future, when the iteration scheme is cleaned up in VMEC++, one could have a second look and maybe implement the PARVMEC-style iteration scheme using an optional flag in the input file.

Dreaming about the future, it would be nice to understand the differences between the two methods and merge them to get the best of both.

CharlesCNorton:
Resolves #157.

The PARVMEC iteration scheme was reachable only through the Python iteration driver (`vmecpp.iterate` / `solve_multigrid`). This makes it a first-class input-file option honored by the native solver, so `vmecpp.run(...)` with `iteration_style="parvmec"` uses it.

## What
- Native C++ PARVMEC time-step control in `Vmec::SolveEquilibriumLoop`, gated on `indata.iteration_style`: track the running minima of both the preconditioned and invariant residual sums (`res0` / `res1`), store when both improve, and revert only when either exceeds 1e4x its minimum after 10 steps, via the gentle `BAD_PROGRESS` path (no ijacob escalation). The default `vmec_8_52` branch is unchanged.
- Plumb `iteration_style` through the Python `VmecInput` model (mirroring `free_boundary_method`) so the flag survives `from_file` and the `VmecInput` -> `VmecINDATA` round-trip.
- Lift the `run()` guard that rejected any style but `vmec_8_52`.

## Verification
- `vmec_8_52` is byte-identical: solovev / cth_like / cma converge to unchanged volumes.
- The native PARVMEC control reproduces the tested Python PARVMEC driver step for step (identical restart-reason trace, residuals agreeing to ~1e-12): added as `test_native_parvmec_matches_python_parvmec`.
- `vmecpp.run()` honors the input flag: both styles converge to the same equilibrium via different iteration paths (`test_run_honors_iteration_style_flag`).
- The input attached to this issue (crashes PARVMEC, works in VMEC++) converges under both styles to the same equilibrium: 996 iterations for `vmec_8_52`, 1746 for `parvmec`.


CharlesCNorton:
fixing

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
```diff
@@ -949,6 +950,9 @@ absl::StatusOr<Vmec::SolveEqLoopStatus> Vmec::SolveEquilibriumLoop(
       }
     }
 
+    const bool parvmec_iteration =
+        indata_.iteration_style == IterationStyle::PARVMEC;
+
```
jurasic-pf: Nit: can we inline the check those two times? I find it slightly easier to read

tests/test_iteration.py
```diff
@@ -208,6 +208,60 @@ def test_alternative_styles_converge_cma():
...
+    result = vmecpp.solve_equilibrium(model, style="parvmec")
+
+    assert not result.failed
+    np.testing.assert_array_equal(
+        np.asarray(result.restart_reasons), np.asarray(reference.restart_reasons)
+    )
+    cpp_r = np.asarray(reference.force_residual_r)
+    py_r = np.asarray(result.force_residual_r)
+    np.testing.assert_allclose(py_r[:50], cpp_r[:50], rtol=1.0e-9, atol=1e-15)
+    np.testing.assert_allclose(py_r, cpp_r, rtol=1.0e-3, atol=1e-15)
```
jurasic-pf: 1e-3 rtol can mask a lot of issues, are we sure these are the same?

tests/test_iteration.py
```diff
@@ -208,6 +208,60 @@ def test_alternative_styles_converge_cma():
...
+        inp = base.model_copy(update={"iteration_style": style})
+        assert inp.iteration_style == style
+        assert inp._to_cpp_vmecindata().iteration_style == getattr(
+            _vmecpp.IterationStyle, style.upper()
+        )
+        outputs[style] = vmecpp.run(inp, max_threads=1, verbose=False)
+    # Same physics regardless of the iteration scheme.
+    assert outputs["parvmec"].wout.volume_p == pytest.approx(
+        outputs["vmec_8_52"].wout.volume_p, rel=1.0e-9
+    )
```
jurasic-pf: Asserting the volume is the same is not an exhausitve physics check. Let's confirm some more relevant physics quantities of the magnetic field or geometry for example.

jurasic-pf:
Thanks! The PR is clean, just a few nits.

Can we do a verification in CI by storing a few wouts produced with https://github.com/ORNL-Fusion/PARVMEC/ to LFS and asserting those wouts match exactly for the new iteration style, like we do with the large cpp tests? I'll add a CI job which does the full test suite as well, since that's missing atm

CharlesCNorton:
Built ORNL-Fusion/PARVMEC and confirmed the committed reference wouts match its output to machine precision (volume/aspect ~1e-15, geometry/iota ~1e-7), so they already serve as PARVMEC references. `test_parvmec_matches_parvmec_reference` asserts the parvmec style reproduces them. Can commit fresh PARVMEC wouts to LFS if you prefer explicit provenance.

jurasic-pf:
Hmmm, that's not quite what I meant 😅
The PARVMEC flow control is subtley different from VMEC8.52, which is what vmecpp is based on. To make this a first class feature, I think it's important to validate the two behave exactly the same on edge cases that trigger different flow control between them. E.g. an equilibrium that triggers many re-starts + has multigrid iterations.

The minimal test case of interest is that the force residual **progression** between PARVMEC and VMEC8.52 differ, and then we also confirm that the new iteration logic exactly follows the force residual **progression** of ORNL PARVMEC. 

When they converge, both are expected to converge to the same equilibrium (which you just tested), and that's expected even if there's subtle errors in the flow control logic.

jurasic-pf:
The issue https://github.com/proximafusion/vmecpp/issues/157 contains an example input file in a comment which can be run with PARVMEC, vmecpp and vmecpp with the new iteration to confirm, since that one should excert different behavior between the two.

CharlesCNorton:
@jurasic-pf

Pushed 25ab040 for the two-part check you describe.

Part one (the progressions differ): `test_iteration_styles_diverge_on_stiff_case` asserts the `vmec_8_52` and `parvmec` residual progressions diverge on `cma` (ns=72), a restart-triggering input.

Part two (`parvmec` follows ORNL PARVMEC): the native control matches PARVMEC to round-off where the trajectory is insensitive to round-off. On `cth_like_fixed_bdy` (ns=25), (fsqr, fsqz, fsql) agree with Fortran PARVMEC to ~1e-13 over the first steps, saturate near 1e-4 thereafter, and converge in the same step count (119 vs 120); the cold-start reguess is included (RAXIS_CC to ~1e-16). `test_parvmec_follows_ornl_parvmec_trace` pins the control to that committed trace.

An exact match is not well-posed on the restart-triggering inputs themselves. I ran the #157 input you cite through PARVMEC and both VMEC++ styles: the transient reaches fsqr ~1e4, the iteration map is sensitively dependent on the initial state, and a 1e-13 difference reaches order unity within three steps, while all converge to the same equilibrium. There the residual sequence is governed by round-off propagation rather than the control logic; its equality is therefore not a valid discriminant. #157 also does not converge under PARVMEC at ns=51 (51 Jacobian resets, terminating near 1e-2); `cma` (ns=72) serves as the convergent restart case.

jurasic-pf (approved):
Ok, the arguments are sound, then I think this is safe :)
Thanks again for the many contributions recently, this is quite helpful

PR #616, opened 2026-07-08
2D block-tridiagonal preconditioner

CharlesCNorton:
VMEC++'s convergence preconditioner is the per-mode radial operator assembled in `computePreconditioningMatrix` and applied in `applyRZPreconditioner`: for each Fourier mode `(m, n)` it solves a scalar tridiagonal system down the radial grid, the poloidal and toroidal mode numbers entering only as diagonal `m²` and `(n·nfp)²` weights. Different modes are never coupled to one another. That approximation is cheap and robust, but it is also the one that weakens on strongly-shaped, high-mode-number equilibria, where the true Hessian couples the poloidal modes within a flux surface. PARVMEC handles this with `precon2d`/BCYCLIC: a dense per-surface block coupling all poloidal modes, with the surfaces tied together radially into a block-tridiagonal system. This PR lands the block-tridiagonal solver and the wiring that reshapes the R/Z preconditioner into that form; the mode-coupling coefficients that make it a genuine 2D preconditioner follow on this branch.

## What

**`common/block_tridiagonal`** — a standalone, dependency-light package:

- `BlockTridiagonalFactorization` — block-Thomas factorization and solve for `L_j x_{j-1} + D_j x_j + U_j x_{j+1} = b_j` over `n` block rows of size `k`. The forward sweep stores the explicit inverse of each reduced pivot block, so the many right-hand sides that share one operator (one per Fourier basis function) reuse a single factorization.
- `AssembleFdBlockTridiagonal` — assembles the Jacobian `dF/dx` of a force field with nearest-neighbour radial coupling by central finite differences. Perturbing surface `j` moves the force only on surfaces `j-1, j, j+1`, so surfaces three apart are perturbed together and the assembly costs `6·k` force evaluations independent of the number of surfaces.
- Unit tests for both, checked against a known analytic solution, a dense `(n·k)×(n·k)` reference solve, and an analytic Jacobian, across a range of block counts and block sizes.

**`applyRZPreconditioner`** — an opt-in path, `solveBlockRZPreconditioner`, that regroups the R/Z preconditioner as one block-tridiagonal system per toroidal mode number `n`: the poloidal modes of a surface form each block, the radial grid indexes the block rows, and the system is solved with the new factorization. It is gated behind `VMECPP_PREC2D`; the per-mode scalar solve remains the default and is untouched.

## Verification

- With only the block diagonals populated from the existing 1D coefficients, the block operator is algebraically the per-mode scalar solve, so the two converge to the same equilibrium up to floating-point round-off. On `cth_like_fixed_bdy` and `solovev`, converged `volume_p` and `aspect` match the default to all printed digits and `betatotal` and the magnetic-axis position agree to `~1e-12`, with `VMECPP_PREC2D` set and unset; the residual difference is the block-Thomas sweep's arithmetic ordering.
- `BlockTridiagonalFactorization` matches both the analytic solution and the dense reference solve to machine precision (`~1e-14`); `AssembleFdBlockTridiagonal` recovers the analytic Jacobian of a known nearest-neighbour force field.
- Full CMake build; Bazel targets and deps added for the new package. clang-format and clang-tidy clean.

## Follow-up

The off-diagonal poloidal coupling is the angle-resolved form of the same metric integrals that `computePreconditioningMatrix` currently angle-averages into the per-mode diagonal (`ax`/`bx` summed over the poloidal grid). Adding those harmonics to the block off-diagonals, together with the second-order solve that consumes them, turns this into the full 2D preconditioner and lands next on this branch.


src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -2531,18 +2601,26 @@ absl::Status IdealMhdModel::applyRZPreconditioner(
...
-                           std::span<double>(m_h_.all_br.row(mn).data(), ns),
-                           m_h_.all_cr[mn].data(), ns, jMin[mn], jMax,
-                           s_.num_basis);
-    TridiagonalSolveSerial(std::span<double>(m_h_.all_az.row(mn).data(), ns),
-                           std::span<double>(m_h_.all_dz.row(mn).data(), ns),
-                           std::span<double>(m_h_.all_bz.row(mn).data(), ns),
-                           m_h_.all_cz[mn].data(), ns, jMin[mn], jMax,
-                           s_.num_basis);
-  }  // mn
+  static const bool prec2d = std::getenv("VMECPP_PREC2D") != nullptr;
```
jurasic-pf: Fine for development and experimentation, but before merge Please prefer a field in the input object over environment variables.

jurasic-pf:
Cool! When I briefly looked into this sort of pre-conditioning the initial implementation was so computationally expensive that omitting it was simply faster, but I'm sure there was a lot more untapped optimization potential.

What's the plan with the PR, just keep it open until the investigation has reached a conclusion and then either merge or close the stack of PRs depending on the results? Shall I already review the individual ones in the meanwhile?

CharlesCNorton:
Agreed. `VMECPP_PREC2D` is a development toggle while the off-diagonal coupling coefficients are still being worked out — with only the block diagonals populated it just reproduces the 1D preconditioner, so the flag has no user-visible effect yet. I'll expose it as a field on `VmecInput` (threaded through `VmecINDATA`) rather than an env var in the same series, before this is merge-ready, once turning the block path on actually selects a different preconditioner.

CharlesCNorton:
@jurasic-pf

Status on the off-diagonal coupling.

The block-tridiagonal solver and FD assembler reproduce the per-mode 1D preconditioner exactly (the committed state). Coupling attempts so far:

- Analytic off-diagonal from the angle-resolved `ax` (radial) metric integrand: converges to the same equilibrium but takes more force iterations than the 1D (cth_like 147 vs 121; solovev 375 vs 204). A runtime scan over coupling scale and sign found no value that lowered the count.
- Anderson acceleration of the 1D-preconditioned fixed point `x ← x + M⁻¹F`: converges but does not fall below the native momentum iteration on the cases tested (cth_like 285 vs 239; cma ns=15 ≈1700 vs 635).
- A full 1D-preconditioned Hessian inverse (FD Jacobian, SVD-truncated) lowers the iteration count substantially (cma ≈40 vs ≈500) at O(N³) solve and O(N) evaluations per update; its wall-clock at production resolution is not yet measured.

Remaining steps:

**Complete the explicit coupling**
- Add the `bx` (m²) poloidal-curvature term to the block off-diagonal (current coupling is `ax` radial only).
- Add the R/Z cross-coupling and the odd/mixed-parity cross terms (current coupling is same-parity, per-component).
- Validate the assembled coupling against the true finite-difference Hessian off-diagonal.
- Fix the multi-threaded `axmat` sizing (full-range storage + per-partition fill) so it runs threaded.
- Re-scan coupling scale and sign once the coupling is complete.

**History-based / quasi-Newton acceleration**
- Anderson combined with the native momentum rather than replacing it.
- Tuned Anderson: relaxation factor, least-squares regularization, restart depth.
- L-BFGS two-loop recursion on the 1D-preconditioned residual.
- Limited-memory Broyden.

**SVD-truncated Hessian**
- Measure its wall-clock at production resolution against the native iteration.
- Replace the 2N-evaluation dense FD Jacobian with the stride-3 block-tridiagonal FD (≈6k evaluations).
- Replace the full SVD with a rank-revealing or iterative solve.
- Characterize the rcond-threshold sensitivity and the near-null m=0 radial subspace.

**Cheaper inner solves (contingent on a helpful coupling)**
- Banded block solve exploiting low |m−m′| coupling.
- Woodbury low-rank correction to the 1D block.

**Alternative constructions**
- Alternating-direction split: 1D radial solve, then a per-surface poloidal solve.
- Mode-space multigrid.

**Coverage and infrastructure**
- Test strongly-shaped and high-mode-number cases (W7-X, the #157 input) and larger ns/mpol.
- Add a wall-clock benchmark harness (iterations and time).

CharlesCNorton:
@jurasic-pf

- A full 1D-preconditioned Hessian inverse (FD Jacobian, SVD-truncated), evaluated on #626's history-independent raw force, carries real coarse-tolerance curvature (cth_like reaches 1e-4 in 8 steps, versus 67 for the native) but stalls before tight convergence, fails to reach even 1e-4 on cma, and does not converge when used as a momentum-iteration preconditioner. The near-null m=0 subspace defeats the truncated inverse near convergence, and the dense assembly costs roughly 10x the native solve, so it is not a usable accelerator.

PR #618, opened 2026-07-09
3d external optimizer example

jurasic-pf:
```log
VMEC++ (native, vmecpp.run)          1645   1645     1.68    1.1e-10    5.0e-06
preconditioned descent              20000   9999    21.83    1.0e-09    6.0e-10
Traceback (most recent call last):
  File "/home/jurasic/Documents/vmecpp/examples/external_optimizers.py", line 269, in <module>
    main()
    ~~~~^^
  File "/home/jurasic/Documents/vmecpp/examples/external_optimizers.py", line 261, in main
    r = solver()[1]
```
Thanks to @krystophny  we now have the ability to implement second order methods for the descent!
But for more difficult 3D cases, the second order descents are not yet working. Would you be interested to look further into this?

### DIFF
--- a/examples/external_optimizers.py
+++ b/examples/external_optimizers.py
@@ -24,6 +24,15 @@ benchmark ``main`` below and by the tests:
 All converge to the same equilibrium as the native solver. Force evaluations are
 counted inside VMEC++ (``force_eval_count``) so the comparison is fair across
 methods, including the evaluations hidden in Hessian-vector products.
+
+Interpretation: what separates the methods is how they couple DOFs of
+neighbouring flux surfaces. Plain JFNK does not -- it treats every degree of
+freedom as independent, so it cannot damp the radial stiffness that lets a
+surface cross its neighbour between iterations (BAD_JACOBIAN restarts in the
+native solver). It still converges, but slowly. VMEC's couples each surface to
+its jF +/- 1 neighbours; applying it -- as the inner Krylov preconditioner, or
+refreshed every step in the HVP Newton -- is what rescues conditioning and cuts
+the cost by an order of magnitude.
 """
 
 from __future__ import annotations
@@ -36,10 +45,14 @@ import numpy as np
 from scipy.optimize import newton_krylov
 from scipy.sparse.linalg import LinearOperator, gmres
 
+import vmecpp
 from vmecpp.cpp import _vmecpp  # type: ignore[import]
 
 DEFAULT_INPUT = (
-    Path(__file__).resolve().parents[1] / "examples" / "data" / "solovev.json"
+    Path(__file__).resolve().parents[1]
+    / "examples"
+    / "data"
+    / "cth_like_fixed_bdy.json"
 )
 
 
@@ -111,6 +124,29 @@ def _finish(model, name, x, outer_iters, t0):
     )
 
 
+def solve_vmecpp(input_path=DEFAULT_INPUT, ns=11):
+    """Native VMEC++ solve through the public API; reports the iteration count.
+
+    Unlike the other variants this drives the ordinary ``vmecpp.run`` path rather
+    than the low-level ``VmecModel`` primitives, so the reported residual/energy
+    are the public wout diagnostics (invariant force residuals and total MHD
+    energy) rather than the raw internal-basis force this module counts elsewhere.
+    """
+    vmec_input = vmecpp.VmecInput.from_file(input_path)
+    vmec_input.ns_array = np.asarray([ns], dtype=vmec_input.ns_array.dtype)
+    t0 = time.perf_counter()
+    output = vmecpp.run(vmec_input, verbose=False)
+    wout = output.wout
+    return output, Result(
+        name="VMEC++ (native, vmecpp.run)",
+        force_evals=wout.niter,
+        outer_iters=wout.niter,
+        seconds=time.perf_counter() - t0,
+        residual_norm=float(np.sqrt(wout.fsqt[-1])),
+        energy=wout.wb + wout.wp,
+    )
+
+
 def solve_preconditioned_descent(
     input_path=DEFAULT_INPUT, ns=11, tol=1e-9, delt=0.9, momentum=0.5, max_iter=20000
 ):
@@ -134,7 +170,7 @@ def solve_preconditioned_descent(
 
 
 def solve_newton_krylov(
-    input_path=DEFAULT_INPUT, ns=11, tol=1e-9, max_iter=300, preconditioned=False
+    input_path=DEFAULT_INPUT, ns=11, tol=1e-9, max_iter=2500, preconditioned=False
 ):
     model = make_model(input_path, ns)
     F = residual(model)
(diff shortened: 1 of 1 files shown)


jurasic-pf:
* **#618** <a href="https://app.graphite.com/github/pr/proximafusion/vmecpp/618?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/proximafusion/vmecpp/618?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/proximafusion?ref=merge-queue-instructions-in-stack-comment&prId=618).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

krystophny:
Thanks for the 3D benchmark! The crash is scipy's `NoConvergence` exception from `newton_krylov` hitting max_iter on the 3D case — `main()` doesn't catch it, so it kills the script after the first two solvers finish.

The preconditioned descent barely converges (hits the 20000-iteration ceiling), and the Newton-Krylov variants don't converge at all within 2500 iterations because the finite-difference Jacobian approximation can't handle the radial stiffness between flux surfaces in 3D. This is the coupling problem the module docstring describes: JFNK lacks the surface-to-surface coupling that VMEC's preconditioner provides.

The smallest fix: wrap the solver calls with try/except for `NoConvergence` and print a "did not converge" row instead of crashing. Opened #620 to track this.

I can look further into improving the second-order methods for 3D — the HVP Newton solver (which uses the analytic Hessian-vector product and backtracking line search) is the most promising candidate since it has the right coupling structure, but it may need tighter globalization or better inner tolerances for the 3D case.

krystophny:
**Correction to my earlier comment** — the actual CI failure is not `NoConvergence` but assertion errors from tight test tolerances. All 5 solvers actually complete in CI:

```
VMEC++ (native, vmecpp.run)          1645   1645     1.20    1.1e-10    5.0e-06
preconditioned descent              20000   9999    16.91    1.0e-09    6.0e-10
Newton-Krylov (preconditioned)       1937      0     1.84    1.2e-09    2.1e-09
Newton-Krylov (JFNK)                64832      0    60.90    2.8e-09    2.7e-08
Newton (VMEC++ HVP + M^-1)          24605     10    22.03    1.0e-10    9.9e-10
```

The tests fail because the tight tolerances (energy < 1e-8, state norm < 1e-5) that worked for the 2D Solov'ev case are too strict for the 3D cth_like_fixed_bdy case, where the solvers converge to valid equilibria but in slightly different basins. All 4 tests show FFFF across every CI matrix entry.

The crash you see locally is environment-specific — likely a scipy/BLAS combination where `newton_krylov` diverges and raises `NoConvergence` instead of slowly converging.

The deeper issue is that `residual()` doesn't check `model.restart_reason` after `evaluate()`. When BAD_JACOBIAN fires (flux surfaces crossing during iteration), `evaluate()` returns silently with stale forces — the native solver has 200+ lines of backup/restore/reguess recovery for this, but none of the external solvers check for it. This makes 3D convergence fragile and environment-dependent.

The HVP Newton is the standout: it reaches native-solver-quality convergence (||F|| = 1.0e-10, dW = 9.9e-10) in 10 outer Newton steps on this 3D case. The path to robust 3D external solving is adding BAD_JACOBIAN recovery to the HVP Newton's line search + eventually PTC for the stiffest cases.

Updated #620 with the full analysis from CI logs and the C++ code.

krystophny:
Correction and status update. My earlier BAD_JACOBIAN root-cause claim is **wrong** and I withdraw it: measuring against a compiled build, BAD_JACOBIAN fires 0/200 times at the eps_rel=1e-7 perturbation the FD-Jacobian and HVP actually use (both 2D and 3D), so it does not corrupt the solver steps near convergence.

What is actually verified:

- Only `test_optimizer_reaches_equilibrium` fails (4/4); it fails on the state-vector assertion `||x - x*|| < 1e-5` (measured 0.0066-0.020), not on force balance.
- There is a genuine near-null direction of the augmented Hessian at the 3D equilibrium: HVP-Newton driven to `||F|| = 9e-15` still sits `||x - x*|| = 0.012` from the native reference, and that direction has Rayleigh quotient `2.5e-7` vs ~15 for a random direction. Native and external solvers reach force balance ~1e-2 apart along it at the same energy.
- Separately I found a real latent bug: the **first** `evaluate()` on a fresh model returns a wrong force on the 3D case (`3e-6` vs the correct `9e-12` on the second call). It does not cause these test failures but I'm tracking it.

I am NOT going to weaken the test. Before proposing any fix I'm settling whether that soft direction is a true (unphysical) gauge or a fixable under-constraint -- by decomposing it into lambda vs R/Z content and comparing the two states' physical geometry. Details in #620; PR to follow once that's nailed.

krystophny:
Final analysis, after checking against a compiled build and the VMEC literature (correcting my earlier comments).

The 4 failing tests all trip the raw-state assertion `||x - x*|| < 1e-5` (0.0066-0.020). The difference is **99.8% in the lambda block** -- VMEC's poloidal-angle gauge (`theta -> theta + eps`; pinned by the spectral-condensation constraint, Hirshman & Breslau 1998, as a *soft* force, so weakly determined near the minimum). Two measurements settle it:

- The physical mismatch does NOT shrink as `||F|| -> 0` (`d_iota` identical at `||F|| = 1e-9` and `1e-11`), so these are near-degenerate distinct stationary points, not exact gauge copies. Gauge-invariant physics still agrees: energy `1e-9`, enclosed currents `1e-6`, iota `1e-4`.
- The native solver from a perturbed start lands `1.367e-2` from the unperturbed reference. So `||x - x*|| < 1e-5` **fails even native-vs-native** -- it rejects the reference implementation. It is well-posed only on the 2D Solov'ev case (all solvers reach `~1e-7` there).

No solver code bug -- the optimizers reach the same equilibrium to the native solver's own gauge-reproducibility.

Fix (PR against main, incoming): keep the strict raw-state test on the 2D case; add a strict 3D test asserting gauge-invariant physics vs the native reference (energy `<1e-8`, iota `<5e-4`, enclosed current `bvco <1e-5`). `DEFAULT_INPUT` stays 3D so this benchmark `main()` is untouched.

Separately I found a genuine forward-model bug (own PR): the first `evaluate()` on a fresh model returns a wrong force on 3D (`3e-6` vs the correct `9e-12`), self-healing on the second call -- a warmup lag in the `tcon`/preconditioner-scaling path. It does not affect the shipped solvers' final results. Details in #620.

krystophny:
Follow-up after reproducing the failures and checking VMEC2000, educational VMEC, jVMEC, and the spectral-condensation literature:

- The four CI failures were test-target failures after the shared default changed to 3D, not solver crashes. #623 restores the deleted coverage with explicit 2D and 3D inputs.
- The earlier BAD_JACOBIAN and `tcon` explanations were refuted. The separate first-evaluation defect is the legacy previous-`fsqz` `m=1` projection policy; #626 fixes it without changing native iteration behavior.
- The benchmark mixed residual definitions, used `wb + wp` instead of the configured MHD functional for the `gamma=0` case, reported zero SciPy outer iterations, and incorrectly said plain JFNK lacks radial coupling. #625 fixes those reporting errors without changing solver steps.

The corrected issue analysis is in #620 and #624.

PR #619, opened 2026-07-09
Refactor Fourier-resolution continuation into vmecpp.run()

### LINKED ISSUE #209: "Multiresolution" in both radial and Fourier space
## Context

VMEC++ currently employs a multi-resolution approach in the the radial direction (ns), first solving on a coarse grid and then using it as an initial guess for refining. This generally improves robustness and convergence speed and is controlled by the ns_array, ftol_array, and niter_array input parameters in `VmecINDATA`. 

However, the angular (Fourier) resolution, defined by mpol and ntor, remains fixed throughout the entire sequence. For complex geometries or cases far from an initial guess, starting with the full target Fourier resolution can lead to slow convergence.

## Proposal

Extend the continuation method concept to something more "multi-grid" like, by also include the Fourier resolution. Mpol and ntor would now become input arrays, just like ns.
- mpol_array: An array specifying the mpol value to use for each multi-grid step.
- ntor_array: An array specifying the ntor value to use for each multi-grid step.

## Implementation sketch

- Input Handling: Modify `VmecINDATA` corresponding Python files
  - Sanity checks, that resolution is only increasing?
  - Validation logic that all have the same size or a single element 
- Compute loop: When a change occurs, resolution-dependent objects must be resized and potentially re-initialized:  
  - `Sizes` object (s_)
  - `FourierBasis` objects (t_, fb_ in Nestor)
  - `FourierCoeffs` subclasses: `FourierGeometry`, `FourierForces`, `FourierVelocity`. The existing `VmecINDATAPyWrapper::SetMpolNtor` and the resizing logic should be a good starting point, since it zero initializes the additional coefficients.
- Interpolation on resolution increase: the solution vector x (represented by `FourierGeometry` containing `rmnc, zmns, lmns`, etc.) from the coarser Fourier grid needs to be transferred to the new, finer grid. 
  - Zero-padding. Set all the newly added higher-order Fourier coefficients to zero. Fourier interpolation is well behaved and should work fine.
  - Maybe FFT to realspace and back to take discretization errors into account on the first step?
- Free Boundary (Nestor): depends on mf (related to mpol) and nf (ntor).

clazzati-pf:
## Summary

- Allow `VmecInput.mpol`/`.ntor` to be a sequence of ints (one entry per
  `ns_array` step, scalar broadcasts to every step) in addition to a plain
  int. A plain int keeps today's behavior byte-for-byte; a sequence now makes
  `vmecpp.run()` perform Fourier-resolution continuation directly, reading
  the schedule from `input`'s own fields (mirroring how radial continuation
  already works via `ns_array`), instead of through a separate
  `vmecpp.run_continuation()` entrypoint with its own duplicate kwargs.
- Remove `vmecpp.run_continuation()`, now subsumed by `vmecpp.run()`.
  `vmecpp.interpolate_solution()` is unchanged and stays public.
- Additive behavior: `vmecpp.run(input, restart_from=...)` now hot-restarts
  the first continuation step from `restart_from` (interpolated to its
  resolution) instead of always cold-starting it, when `input.mpol`/`.ntor`
  is a sequence.

This is a pure Python-level interface change (`VmecInput`/`vmecpp.run()`);
no C++ core, JSON input schema, or `_vmecpp.VmecINDATA` changes are involved.

## Test plan

- [x] `pre-commit run --files src/vmecpp/__init__.py src/vmecpp/_continuation.py tests/test_continuation.py examples/fourier_resolution_increase.py` (ruff, ruff-format, docformatter, pyright, misc hooks) passes
- [x] `pytest tests/test_continuation.py -v` — adapted/new tests pass (coercion of a length-1 sequence to a scalar, mismatched-schedule-length `ValueError`, ns-only and Fourier continuation matching the direct C++ multigrid, `restart_from` hot-restarting the first continuation step)
- [x] Full `pytest` suite passes (168 passed, 2 skipped; no regressions)
- [x] `python examples/fourier_resolution_increase.py` runs end-to-end and reproduces the expected iteration-count/geometry comparison


jurasic-pf:
This looks quite good already, thanks!

jurasic-pf (approved):

src/vmecpp/__init__.py
```diff
@@ -77,6 +77,32 @@ def _wrap_int_as_float(
...
+    resolution continuation schedule.
+    """
+    if isinstance(value, (int, np.integer)):
+        return int(value)
+    array = np.atleast_1d(np.asarray(value, dtype=np.int64))
+    return int(array[0]) if array.size == 1 else array
+
+
+MpolNtorField: typing.TypeAlias = typing.Annotated[
+    int | jt.Int[np.ndarray, "num_fourier_steps"],
```
jurasic-pf, line 95: Why a numpy array here instead of a list? I think ftol and ns are lists, please align the design with those.

clazzati-pf: ftol and ns are also arrays, so should be already consistent

jurasic-pf:
Address and resolve the comment when done, it will auto merge

jurasic-pf:
Ok CI should be fixed

PR #622, opened 2026-07-09
Remove explicit test of an example script

jurasic-pf:
Hotfix after changing default to a 3d equilibrium

Issue #620, opened 2026-07-09
External optimizer test inherited an invalid 3D state target


krystophny:
## Summary

#618 changed `examples/external_optimizers.py::DEFAULT_INPUT` from the 2D Solov'ev case to the 3D CTH-like case. `tests/test_external_optimizers.py` imported the solver defaults, so its strict internal-state assertion changed targets without an explicit test change. All four parametrized cases failed. #622 then removed the complete test module as a hotfix.

## Corrected analysis

The failing 3D state comparison is not a reliable cross-solver oracle:

- External solvers reached raw force residuals below `1e-7`.
- Their internal states differed from the native reference by `0.0066` to `0.020`, with about 99.8% of one measured difference in the lambda sector.
- Tightening the force solve did not remove the state difference.
- Perturbed native starts also reached different nearby internal states.

The poloidal parameterization of a flux surface is nonunique. VMEC uses spectral condensation to select it (Hirshman and Meier 1985; Hirshman and Breslau 1998). The measurements show a weak coordinate-dominated direction and nearby stationary points. They do not show that the finite-resolution states are exact gauge copies, because physical profiles retain small differences.

The strict Solov'ev state check remains useful as an empirical regression target: all four tested solvers reproduce that state to about `1e-7`.

## Fixes

- #623 restores the deleted tests, passes every input explicitly, runs all four solvers once per case, keeps strict 2D state coverage, and checks force balance plus energy on the 3D case.
- #625 fixes unrelated reporting errors introduced in #618. The native row used a different residual and the wrong energy expression for `gamma=0`; SciPy JFNK also reported zero outer iterations.
- #626 fixes the separate history-dependent `VmecModel.evaluate()` defect tracked in #624.

## Corrected current convention

`buco = <B_u>` represents enclosed toroidal current. `bvco = <B_v>` represents enclosed poloidal or external current. The earlier text reversed these names. For the CTH-like `ncurr=1` input, `buco` is prescribed; `iota` and `bvco` are more discriminating solved profiles.

## Verification

```text
#623: 5 passed in 65.25s
#625: 183 passed, 2 skipped in 38.31s
#626: 183 passed, 2 skipped in 34.51s; 20/20 C++ tests passed
```

PR #623, opened 2026-07-09
Restore external optimizer behavioral tests

krystophny:
## Problem

#618 changed the example's shared default input from the 2D Solov'ev case to the 3D CTH-like case. The optimizer test inherited that default and kept a strict internal-state comparison. All four solver cases then failed. #622 restored CI by deleting the entire test module, including force-balance, energy, preconditioner, HVP, and non-axisymmetric cold-start coverage.

The failed state-vector assertion is not a valid 3D oracle. The poloidal parameterization of a flux surface is nonunique; spectral condensation selects a parameterization, but the tested solvers reach weakly separated internal states dominated by the lambda sector. The reported states are nearby stationary points, not proven exact gauge copies. The Solov'ev state comparison remains useful because it is empirically reproducible across these solvers.

## Change

- Restore the deleted behavioral tests.
- Pass the Solov'ev and CTH-like input paths explicitly; tests do not inherit `DEFAULT_INPUT`.
- Run every solver once per case and reuse the module-scoped results.
- Keep strict residual, energy, and state checks on Solov'ev.
- Check raw force balance and relative energy agreement for all four solvers on the CTH-like case.
- Retain the preconditioner evaluation-count, HVP iteration-count, and CMA non-axisymmetric path checks.

There are no optional assertions, unavailable branch APIs, or repeated-evaluation workarounds. The separate `VmecModel.evaluate()` history defect is fixed in #626.

Closes the test-coverage part of #620.

### Test fails on main

Main contains no optimizer test after #622:

```text
$ git show upstream/main:tests/test_external_optimizers.py
fatal: path 'tests/test_external_optimizers.py' exists on disk, but not in 'upstream/main'
```

Immediately before that deletion, the 3D default produced:

```text
$ pytest tests/test_external_optimizers.py::test_optimizer_reaches_equilibrium -q
>       assert np.linalg.norm(x - x_star) < 1e-5
E       AssertionError: assert np.float64(0.012047202036774236) < 1e-05
FAILED ...[solve_preconditioned_descent]
FAILED ...[solve_newton_krylov]
FAILED ...[solve_newton_krylov_preconditioned]
FAILED ...[solve_newton_hvp]
4 failed in 53.98s
```

### Test passes after fix

```text
$ uv run pytest tests/test_external_optimizers.py -q
.....                                                                    [100%]
5 passed in 65.25s
```

Pre-commit passes on the restored test module.

### DIFF
--- a/tests/test_external_optimizers.py
+++ b/tests/test_external_optimizers.py
@@ -1,0 +1,107 @@
+# SPDX-FileCopyrightText: 2024-present Proxima Fusion GmbH
+# <info@proximafusion.com>
+#
+# SPDX-License-Identifier: MIT
+"""External optimizers reach force balance on axisymmetric and 3D cases.
+
+The Solov'ev equilibrium has a reproducible internal state, so the 2D test compares the
+full state vector with the native solver. In 3D, the poloidal parameterization is not
+unique and spectral condensation only regularizes that coordinate freedom. The 3D test
+therefore compares the residual and energy instead of coordinate-dependent Fourier
+coefficients.
+"""
+
+import sys
+from pathlib import Path
+
+import numpy as np
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "examples"))
+from external_optimizers import (  # type: ignore
+    make_model,
+    reference_equilibrium,
+    residual,
+    solve_newton_hvp,
+    solve_newton_krylov,
+    solve_newton_krylov_preconditioned,
+    solve_preconditioned_descent,
+)
+
+ROOT = Path(__file__).resolve().parents[1]
+SOLOVEV = ROOT / "examples" / "data" / "solovev.json"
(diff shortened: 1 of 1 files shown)


krystophny:
Rebased onto current main. This restores `tests/test_external_optimizers.py`, which #622 deleted as a hotfix after the 3D default broke it, and replaces it with a well-posed version: strict raw-state equilibrium check on the 2D Solov'ev case, plus a 3D test asserting gauge-invariant physics (force balance + energy always; iota and enclosed toroidal current on builds that expose `qs_harmonics`). Now MERGEABLE.

jurasic-pf (approved):
Sorry for deleting your test to make CI green again, this test looks better anyway :)

krystophny:
> Sorry for deleting your test to make CI green again, this test looks better anyway :)

Don't worry - I merged main in one more time now.

PR #625, opened 2026-07-09
Report comparable external optimizer metrics

krystophny:
> **Merge order:** merge after #626.

## Problem

The benchmark added in #618 reports incompatible quantities in one table:

- The native row uses `sqrt(wout.fsqt[-1])`; external rows use the Euclidean norm of the raw internal-basis force.
- The native row uses `wout.wb + wout.wp`; external rows use VMEC's configured MHD functional. For the CTH-like input, `gamma=0`, so the functional is `wb - wp`.
- SciPy Newton-Krylov always reports zero outer iterations.

The example also says plain JFNK lacks radial coupling. Its finite-difference Jacobian-vector product uses the full coupled residual. The preconditioner improves the conditioning of that coupled linear system.

## Change

- Run the native solver through `VmecModel` and report it through the same `_finish` path as the external solvers.
- Count every native and external force evaluation at `IdealMhdModel::update()`, the shared force-evaluation boundary.
- Count SciPy outer iterations through `newton_krylov`'s callback.
- Run the native solve once and reuse it as the benchmark reference.
- Correct the HVP and preconditioner descriptions.
- Add regression tests for common metrics, direct native evaluation counts, and nonzero JFNK iteration counts.

Follow-up to #618.

## Verification

The regression test fails twice on `main`: the native energy uses a different functional and JFNK reports zero iterations.

```text
$ uv run pytest tests/test_external_optimizer_benchmark.py -q
FF
2 failed in 2.82s
```

Current branch:

```text
$ uv run --with pytest pytest tests/test_external_optimizer_benchmark.py -q
..                                                                       [100%]
2 passed in 1.89s

$ uv run --with pytest pytest tests/ -q
188 passed, 2 skipped, 66 warnings in 91.86s
```

For the CTH-like native solve, the direct counter reports 1,644 force evaluations for 1,643 recorded nonlinear iterations. The final raw-force norm is `9.16e-12` and the configured MHD energy is `1.2810322534141964e-03`.

Pre-commit passes on all five changed files.


tests/test_external_optimizer_benchmark.py
jurasic-pf, line 1: I don't see why this is a benchmark. Given that it tests the same example another test already does, can we please move the test there?

If the external optimizers are supposed to become "first class features" at some point they can be moved over to src/ with the rest

jurasic-pf (approved):

jurasic-pf:
@copilot resolve the merge conflicts in this pull request

krystophny:
I resolved merge conflicts

jurasic-pf:
Great contributions, thanks for the support!

PR #626, opened 2026-07-09
Make raw VMEC forces history independent

### LINKED ISSUE #624: VmecModel.evaluate returns history-dependent raw forces in 3D
## Summary

`VmecModel.evaluate()` returns different raw forces for the same 3D state depending on the previously evaluated state. The defect affects both directions across the residual threshold, not only the first call on a fresh model.

## Root cause

The original stale-buffer and `tcon` diagnosis was wrong.

After the force transform and internal-basis conversion, `IdealMhdModel::update()` conditionally zeros the constrained 3D `fzcs(m=1)` coefficients (and `fzcc(m=1)` for asymmetric cases). The condition reads `fsqz` from the previous evaluation. The current `fsqz` is computed later. A high-to-low transition therefore omits the projection once; a low-to-high transition applies it once when the current state would not select it.

The observed difference is exactly the affected subspace:

```text
changed entries: 36 / 1650
changed block: fzcs only
changed modes: m=1, n=1..4 on nine interior surfaces
high-to-low maximum difference: 1.55724155e-6
low-to-high maximum difference: 2.7851e-4
```

Setting `tcon0=0` leaves the discrepancy, which rules out spectral-constraint scaling as its cause.

VMEC2000, educational VMEC, and jVMEC all implement the previous-`fsqz` rule. It is legacy native-iteration trajectory behavior, not an accidental transform cache. The defect is using that history-dependent policy as the default for an API advertised as the reusable force operator `F(x)`.

## Fix

#626 separates the policies:

- `VmecModel.evaluate()` enforces the exact constrained subspace by default, so `F(x)` is independent of evaluation history.
- The Python native-iteration loop explicitly retains the legacy previous-residual policy for VMEC2000 trajectory compatibility.
- Regression tests cover both high-to-low and low-to-high histories.

## Verification

```text
Before: 2 failed, 3 passed; 36 / 1650 entries differed in both tests
After: 5 passed in 23.78s
Full Python: 183 passed, 2 skipped in 34.51s
C++: 132 targets built; 20/20 tests passed
```

krystophny:
> **Merge order:** independently mergeable. #582 must use the same enforced `m=1` projection for its exact HVP.

## Problem

`VmecModel.evaluate()` returns different forces for the same 3D state depending on the previously evaluated state. This breaks the advertised `F(x)` contract for external optimizers and finite-difference Hessian-vector products.

The discrepancy is not a stale transform buffer or `tcon` warmup. It is the legacy `m=1` projection gate. `IdealMhdModel::update()` tests the previous evaluation's `fsqz` before zeroing the constrained `fzcs` and `fzcc` coefficients. The current `fsqz` is computed later. Crossing the `1e-6` threshold therefore applies the previous state's projection policy once.

VMEC2000, educational VMEC, and jVMEC use the same previous-residual rule during native iteration. That trajectory behavior must remain available for compatibility, but it is not a valid default for a reusable force operator.

## Change

- Add an explicit `M1ConstraintMode` to the core update.
- Enforce the exact constrained subspace by default for `VmecModel.evaluate()`.
- Make the Python native-iteration loop opt into the legacy previous-residual policy.
- Add high-to-low and low-to-high history tests on the CTH-like 3D case.

Closes #624.

### Test fails on main

```text
$ uv run pytest tests/test_internal_gradient.py -q
...FF                                                                    [100%]
E   AssertionError: Arrays are not equal
E   Mismatched elements: 36 / 1650 (2.18%)
E   Max absolute difference among violations: 1.55724155e-06
E   AssertionError: Arrays are not equal
E   Mismatched elements: 36 / 1650 (2.18%)
E   Max absolute difference among violations: 0.00027851
2 failed, 3 passed in 1.16s
```

### Test passes after fix

```text
$ uv run pytest tests/test_internal_gradient.py -q
.....                                                                    [100%]
5 passed in 23.78s

$ uv run pytest tests/ -q
183 passed, 2 skipped, 66 warnings in 34.51s

$ uv run pytest tests/test_internal_gradient.py tests/test_iteration.py -q
23 passed in 31.64s

$ cd src/vmecpp/cpp && bazel build //... && bazel test //vmecpp/...
Build completed successfully, 132 total actions
Executed 20 out of 20 tests: 20 tests pass
```

Pre-commit passes on all five changed files. The iteration suite verifies that the explicit legacy policy preserves the Python/native trajectory behavior.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
@@ -422,7 +422,7 @@ absl::StatusOr<bool> IdealMhdModel::update(
     int& m_last_full_update_nestor, FlowControl& m_fc, const int iter1,
     const int iter2, const VmecCheckpoint& checkpoint,
     const int iterations_before_checkpointing, bool verbose,
-    M1ConstraintMode m1_constraint_mode) {
+    bool always_fix_m1_gauge) {
   // An axis re-guess after a bad Jacobian can repopulate high geometry modes
   // directly, bypassing the force mask; clear them on the state each iteration.
   if (s_.mpolGeometry < s_.mpol || s_.ntorGeometry < s_.ntor) {
@@ -814,8 +814,9 @@ absl::StatusOr<bool> IdealMhdModel::update(
   m_decomposed_f.m1Constraint(1.0 / std::numbers::sqrt2);
 
   // v8.50: ADD iter2<2 so reset=<WOUT_FILE> works
-  if (m1_constraint_mode == M1ConstraintMode::kEnforce || m_fc.fsqz < 1.0e-6 ||
-      iter2 < 2) {
+  const bool fix_m1_gauge =
+      always_fix_m1_gauge || m_fc.fsqz < 1.0e-6 || iter2 < 2;
+  if (fix_m1_gauge) {
     // ensure that the m=1 constraint is satisfied exactly
     // --> the corresponding m=1 coeffs of R,Z are constrained to be zero
     //     and thus must not be "forced" (by the time evol using gc) away from
--- a/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
@@ -421,7 +421,8 @@ absl::StatusOr<bool> IdealMhdModel::update(
     bool& m_need_restart, int& m_last_preconditioner_update,
     int& m_last_full_update_nestor, FlowControl& m_fc, const int iter1,
     const int iter2, const VmecCheckpoint& checkpoint,
-    const int iterations_before_checkpointing, bool verbose) {
+    const int iterations_before_checkpointing, bool verbose,
+    M1ConstraintMode m1_constraint_mode) {
   // An axis re-guess after a bad Jacobian can repopulate high geometry modes
   // directly, bypassing the force mask; clear them on the state each iteration.
   if (s_.mpolGeometry < s_.mpol || s_.ntorGeometry < s_.ntor) {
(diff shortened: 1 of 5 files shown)


src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -813,7 +814,8 @@ absl::StatusOr<bool> IdealMhdModel::update(
   m_decomposed_f.m1Constraint(1.0 / std::numbers::sqrt2);
 
   // v8.50: ADD iter2<2 so reset=<WOUT_FILE> works
-  if (m_fc.fsqz < 1.0e-6 || iter2 < 2) {
+  if (m1_constraint_mode == M1ConstraintMode::kEnforce || m_fc.fsqz < 1.0e-6 ||
+      iter2 < 2) {
```
jurasic-pf: nit
```suggestion
  if(m_fc.fsqz < 1.0e-6 || iter2 < 2)) {
  m1_constraint_mode = kEnforce;} 
  if (m1_constraint_mode == M1ConstraintMode::kEnforce) {
```

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.h
```diff
@@ -34,6 +35,8 @@
 
 namespace vmecpp {
 
+enum class M1ConstraintMode : std::uint8_t { kLegacy, kEnforce };
```
jurasic-pf: Can we please name these state variables something more semantically meaningful? Does an enum even make sense here? 

What do they describe logically? 
Enforce M1 constraint
Unconstrained m1 modes

a boolean seems fine here

src/vmecpp/cpp/vmecpp/vmec/pybind11/pybind_vmec.cc
```diff
@@ -1260,7 +1266,8 @@ PYBIND11_MODULE(_vmecpp, m) {
       .def_static("create", &VmecModel::Create, py::arg("indata"),
                   py::arg("ns"), py::arg("initial_state") = std::nullopt)
       .def("evaluate", &VmecModel::Evaluate, py::arg("iter1"), py::arg("iter2"),
-           py::arg("precondition") = true)
+           py::arg("precondition") = true,
+           py::arg("legacy_m1_constraint") = false)
```
jurasic-pf: Please don't use the wording "legacy m1" here. It is simply whether the physically meaningless M1 gauge degree of freedom is free or fixed.

jurasic-pf (changes requested):
see comments, functionally this is quite insightful!

jurasic-pf (approved):

Issue #628, opened 2026-07-09
Solve free boundary once and for all


jurasic-pf:
NESTOR has it's problems, but is to date the best free boundary solver we have for VMEC. The B^2 pressure field it produces inherently doesn't assume any sheet currents by construction, and qualitatively "leads to smooth convergence dynamics". Fundamentally better solvers for the virtual casing / exterior field contribution exist, e.g. BIEST but have poor convergence dynamics when coupled to VMEC that have to be understood first.

[BIEST](https://github.com/dmalhotra/BIEST) is an interesting candidate, if we can make it converge robustly and penalize sheet currents.

See the verification calculations done in
- https://arxiv.org/abs/2412.05680
- https://pubs.aip.org/aip/pop/article-abstract/32/4/043906/3344075/Verification-of-the-Shafranov-shift-in-free?redirectedFrom=fulltext

In particular a high pressure equilibrium with large shafranov shift shpuld be a challenging test case, because the virtual casing contribution is then particularly large

Goal: Get a robust and accurate free boundary solver for VMEC

CharlesCNorton:
on it 👍

jurasic-pf:
One important verification should be that actual surface currents only appear when the pressure jump at the boundary is non zero. (not to be confused with the virtual surface currents from virtual casing)

CharlesCNorton:
These two observations come from validating the AEGIS virtual-casing solver (#629) against BIEST, the reference recommended in that thread. AEGIS is superseded by the native BIEST and Vac2 solvers (#637, #638) and #629 is closed; the observations below are general and bear on how free-boundary solver quality is judged for this issue.

**Field tangency, not fixed-boundary magnitude accuracy, governs in-loop quality.** The free-boundary force couples the vacuum field to the boundary only through |B_ext|²/2, so it enforces pressure balance rather than n·B_ext = 0. A solver's in-loop behaviour therefore depends on whether its field is tangent by construction. NESTOR, Vac2, and BIEST are each tangent by construction, and on cth_like up to β ≈ 2% their magnetic axes agree to within ~2 mm. AEGIS evaluates the virtual-casing field without solving for tangency, so its field is not tangent: on a 3D NCSX case it matched BIEST's magnitude to 0.02% (direction cosine 0.999991) while its boundary normal-field residual was n·B/|B| = 2e-2 against BIEST's 1e-13, and the coupled equilibrium's axis was displaced by 4% (R_axis 1.591 vs 1.530).

[...]

PR #629, opened 2026-07-10
Free boundary (#628): convergence diagnosis, edge preconditioner, and AEGIS virtual-casing solver

CharlesCNorton:
Toward #628: a robust and accurate free-boundary solver. This addresses both halves the issue names, the convergence dynamics of the exterior-field coupling and the accuracy of the exterior-field solver itself, and adds AEGIS, a virtual-casing exterior-field solver.

## The convergence dynamics

The free-boundary iteration slows with poloidal resolution while the fixed-boundary iteration saturates:

| mpol | fixed | free (ftol 1e-10) |
|---|---|---|
| 5 | 416 | 490 |
| 12 | 561 | 1606 |
| 16 | 566 | 2094 |
| 20 | 565 | 2579 |

The first read of this coupling inferred a near-singular edge-force block from a finite-difference Jacobian: an exact (ntor+1)-dimensional null space over the vertical Z(m=0, all n) modes (which VMEC stabilizes only at Z(0,0)) plus a continuum of weakly-restored high-m edge modes. Driving the low-level `VmecModel` from Python (the merged `hessian_vector_product` / `solve_equilibrium` machinery, #580) lets that operator be measured directly rather than inferred, and the direct measurement revises the picture:

- **The LCFS edge-force Jacobian is well-conditioned, not singular.** Building `dF_edge/dx_edge` by finite differences at a converged surface gives 207 active modes (of 240; the rest are structural-zero Fourier entries, which accounts for most of the "null space" the FD Hessian reported) with condition number `6.4e3` and a smooth singular spectrum (513 down to 0.08), no near-null cluster.
- **The obstacle is the coupling, not any single block.** An edge-only Newton step reduces the edge force 21x but raises the *total* residual: moving the boundary disturbs the interior more than it settles the edge. Edge and interior are strongly, non-separably coupled, through the vacuum pressure (a global boundary integral) and through the radial metric.
- **The full coupled force Jacobian has a broad near-null continuum.** Its singular spectrum is ~1061 structural zeros plus a continuum of small singular values (physical condition number ~1e8). Native momentum converges because its preconditioned monotonic damping crushes every mode, slowly; the runaway is momentum's `~sqrt(cond)*log(1/tol)` scaling, dominated by the edge continuum that VMEC's per-mode preconditioner cannot couple.

This is why inversion- and extrapolation-based accelerators, built on the finite-difference operator, do not beat momentum here, refining the earlier "inversion-based preconditioners do not help" into a mechanism: the small singular values sit at or below the FD/NESTOR noise floor (~1e-4 relative at the best step size; the default HVP step `eps_rel=1e-7` is already in the noisy regime). A matrix-free Newton-Krylov solve preconditioned by `M^-1` stalls (GMRES linear residual 0.13-0.98 after 300 matvecs); an SVD-truncated Hessian inverse and a full finite-difference block-tridiagonal preconditioner do not produce a descent direction once the interior/lambda coupling is dropped; Anderson acceleration plateaus (extrapolating a lambda-dominated preconditioned residual); a scalar edge boost (VMEC's commented-out `ledge`) destabilises on the directional stiffness. The native momentum iteration converges (slowly) precisely because it does not invert the operator, and a mode-diagonal preconditioner cannot represent the coupled edge block. Robust tight-tolerance acceleration needs a *noise-free* operator (see "What remains"). This is the "poor convergence dynamics that must be understood first" the issue names.

## The edge preconditioner (opt-in)

`assembleRZPreconditioner` gains an opt-in m-dependent edge damping (`VMECPP_EDGE_A`, `VMECPP_EDGE_P`), default off and byte-identical to the previous flat pedestal. With it on, the loose-to-moderate-tolerance free-boundary iteration collapses onto the fixed-boundary floor (cth_like mpol 16: 2094 to 581 at ftol 1e-10, flat across resolution, equilibrium invariant to 1e-5). It does not resolve tight-tolerance convergence: a scalar/diagonal edge damping cannot represent the dense, directional edge-interior coupling the measurements above show, so the final residual reduction stalls. It is kept default off as a loose-tolerance accelerator while the proper coupled edge preconditioning is developed, for which the block-tridiagonal infrastructure in #616 is the vehicle.

## The vacuum-coupling lag is a first-order control

The free-boundary iteration is a Picard iteration on the vacuum coupling, with the NESTOR-recompute cadence `nvacskip` setting the lag. Fresh vacuum (`nvacskip=1`) cuts the iteration count and, crucially, flattens the runaway (cth_like, robust style, ftol 1e-11):

| mpol | nvacskip=9 (default) | nvacskip=1 |
|---|---|---|
| 5 | 600 | 497 |
| 12 | 2386 | 1151 |
| 16 | 3781 | 1526 |
| 20 | 4888 | 1991 |

The benefit grows with resolution (1.2x at mpol 5, 2.5x at mpol 16-20): the iteration-count runaway is roughly halved in slope. The trade is per-iteration NESTOR cost. The point is that the coupling dynamics are directly controllable, and that fresher coupling matters more as the equilibrium is more strongly shaped, which is exactly where a stiffer, more accurate exterior solver is coupled.

## AEGIS: the accurate exterior-field solver

`examples/aegis_virtual_casing.py` is a validated virtual-casing exterior-field solver, a higher-order alternative to NESTOR's scalar-potential vacuum field. AEGIS (Accurate Exterior Green's Integral Solver) computes the exterior field on the LCFS from the virtual-casing principle: the plasma's exterior field is the Biot-Savart field of the equivalent surface current K = n x B_plasma (plus the n.B_plasma monopole term), with the near-singular self-interaction resolved by Quadrature By Expansion (QBX). Unlike NESTOR, which LU-solves a dense boundary-integral system for a scalar potential every ivacskip iterations, AEGIS evaluates the Green's integral directly from the known surface current, with no linear solve.

Validated on cth_like:

- Biot-Savart integrator vs an analytic current loop: relative error 1e-14.
- LCFS geometry and field reconstruction: B.n/|B| = 2e-16 (tangent to the flux surface).
- Jump condition n x (B_out - B_in) = K via QBX: mean 1%, max 4% across the surface.
- Free-boundary physics: the exterior field B_coil + VC(n x B_plasma) is tangent to the LCFS, and its magnitude equals the interior field at a converged equilibrium (|B_ext|/|B_in| = 0.985).
- End to end, on the same cth_like equilibrium, AEGIS and NESTOR agree on the LCFS vacuum-pressure distribution to 2.8% (mean 0.173 vs 0.178, and standard deviation, min, and max all within 3%): two independent methods, the same exterior field.

A C++ module `free_boundary/aegis/` lands the numerical core: `VirtualCasing` (Biot-Savart + QBX, matching the Python reference to four digits, with a self-contained magnetized-sphere gtest to <0.1%) and the `VacuumPressure` solver entry point that returns |B_exterior|^2/2 per LCFS point.

AEGIS is also wired into the free-boundary iteration itself (opt-in via `VMECPP_AEGIS`): `computeAegisVacuumPressure` reuses NESTOR's LCFS surface geometry and coil field (`GetSurfaceGeometry`/`GetExternalMagneticField`) and the interior equilibrium field (`bsupu`/`bsupv` extrapolated to the LCFS), replicates the field period, and overwrites `vacuum_magnetic_pressure` in place. Live on cth_like it reproduces NESTOR's per-point pressure to under 4%, confirming the geometry, coil field, interior-field, and field-period handling end to end.

## What remains

The convergence work above says the tight-tolerance accelerator must be built on a *noise-free* coupled operator, and the near-singular edge coupling was the obstacle to inversion only in the sense that the FD operator's small singular values were unrecoverable from noise. The exterior-field coupling makes the noise-free path concrete, and each ingredient exists in the tree:

- **Analytic vacuum Jacobian.** `dP_vac/dgeometry` is available in closed form: the chain is `geometry -> amat/bvec (NESTOR) -> potential -> bSubU,bSubV -> P_vac`, and `damat/dgeometry` is exactly the Green's-function derivative `grpmn` that `SingularIntegrals`/`LaplaceSolver` already compute each solve. This gives the dense edge coupling with no finite-difference noise.
- **Exact MHD Hessian.** The Enzyme autodiff force Hessian (#573/#582) supplies the interior/lambda blocks noise-free, removing the FD floor that caps the Newton-Krylov path here.
- **2D block-tridiagonal preconditioner.** #616 lands the block-tridiagonal solver and the radial wiring; populating the LCFS block from the analytic vacuum Jacobian and the interior blocks from the metric integrals turns it into the coupled preconditioner the measurements call for, and which a stiffer, more accurate exterior field makes only more necessary.
- **Accelerating the Picard coupling directly** (Anderson / Chebyshev on the vacuum-pressure sequence; a frozen-vacuum inner Newton with an outer vacuum Picard), and reducing the per-iteration cost of the virtual-casing evaluation (the QBX self-interaction is the dominant term) so a fresh-vacuum coupling is cheap.


jurasic-pf:
First of all, thanks for jumping onto this problem. I just put it out there for traceability, but this is looking quite interesting!

A few conceptual questions:
- Why implement a new free boundary solver and not reuse e.g. [BIEST](https://github.com/dmalhotra/BIEST) which has extensive validation published and is used in DESC? 
- Why the choice of QBX for the singularity hamdling? We do not care about off surface evaluation in this case, so we don't utilize these strengths of the method and  other sorts of singularity subtraction for numerical quadrature are just as applicable (see BIEST paper) 

I must say this is a very interesting hypothesis for why the free boundary solver is slowing down, and definitely plausible. I'm sure you generated a lot of insight on the way to producing this PR. Given it touches core functionality, I'd be very keen on reading more about the verification and experiments you did (e.g. markdown repirt with plots, reproducible experiment scripts) to get to the conclusion this is correct.

If I may suggest experiments, see the verification calculations done in
- https://arxiv.org/abs/2412.05680
- https://pubs.aip.org/aip/pop/article-abstract/32/4/043906/3344075/Verification-of-the-Shafranov-shift-in-free?redirectedFrom=fulltext

In particular a high pressure equilibrium with large shafranov shift shpuld be a challenging test case, because the virtual casing contribution is then particularly large

jurasic-pf:
In particular the discrepancy converged equilibrium (|B_ext|/|B_in| = 0.985). is a cause for concern.

Other important evidence that this is an improvement is that the delbsq metric logged during a free boundary solve converges to smaller values than using nestor at high resolution

jurasic-pf (commented):
Thanks again for this suggestion

CharlesCNorton:
On `|B_ext|/|B_in| = 0.985`: it isn't a magnitude discrepancy. On cth_like the field is coil-dominated (`|B_coil|/|B| ≈ 0.99`, very low plasma current), so the magnitude ratio sits near 1 regardless; the informative quantity is the vector error `|B_ext − B_in|/|B_in|`, which is ~11% — a direction rotation the magnitude hides. It splits into:

- **~6%**: the standalone reference omits the net-toroidal-current field. Virtual casing cannot represent the net enclosed current, so it has to be added as an axis-current filament — the coupling in the PR already does this (the `curtor` term), but `examples/aegis_virtual_casing.py` did not. Adding it takes the vector error 11% → 4.5%.
- **~4.5%**: the QBX on-surface quadrature, independent of both resolution and filament-current scale (confirmed by sweeping each — a method floor).

Plan for parity:
1. Add the axis-current filament to the standalone reference — the 0.985 was this omission.
2. Replace QBX with singularity subtraction for the on-surface evaluation — it removes the 4.5% floor and cuts the per-iteration cost.

Plus the Shafranov-shift verification from the linked papers as an absolute check, and `delbsq` logged through AEGIS- vs NESTOR-driven solves.

On BIEST: the coupling is solver-agnostic — it overwrites `vacuum_magnetic_pressure`, so the convergence work is independent of the exterior solver and BIEST could slot in unchanged. AEGIS evaluates the exterior field directly from the virtual-casing Green's integral with no linear solve, unlike NESTOR's dense LU every `ivacskip` or a boundary-integral solve; at parity with singularity subtraction that is a lighter and faster path, which is the reason to develop it.

On the convergence diagnosis: driving the low-level model from Python (the merged HVP path) let the edge-force Jacobian be measured rather than inferred, and it revised the version you read — the block is well-conditioned (cond ~6e3), not near-singular. The obstacle is the dense edge–interior–lambda coupling through the vacuum pressure plus finite-difference/NESTOR noise, so FD-based accelerators (Newton-Krylov, Anderson, block preconditioners) don't beat momentum; a first-order lever does — fresh vacuum (`nvacskip=1`) roughly halves the iteration count and flattens the runaway (2.5x at mpol 20). A markdown report with the runaway curves, edge-Jacobian spectrum, accelerator results, and reproducible scripts will follow.

CharlesCNorton:
Pushed the first fix (cf18e92): the reference now carries the net-current axis filament, which is the piece the 0.985 was missing. |B_ext|/|B_in| goes 0.985 → 0.999. The residual field-direction error (~4%) is the QBX extrapolation bias, resolution-independent (confirmed by a resolution sweep); singularity subtraction removes it, and that is next.

CharlesCNorton:
Root cause of `|B_ext|/|B_in| = 0.985`: a flipped surface normal. The reference took the LCFS normal as `e_θ × e_ζ`, which is **inward** for VMEC's (θ,ζ) orientation — that flips the sign of `K` and `σ` and the QBX evaluation side, so the virtual-casing reconstruction came out ≈0. The plasma field was never actually being reconstructed; the earlier validations passed only because they were coil-dominated (magnitude), which is blind to VC≈0. A source-free check — reconstructing an analytic interior dipole placed inside the LCFS — exposed it at 99% error.

With the outward normal (`e_v × e_u`, e3541ad):
- VC reconstructs the analytic dipole to **<1%** (validates the operator directly).
- `|B_ext|/|B_in| = 0.9997`, field direction to **0.4%** (was 0.985; ~11% as a vector), and it converges with resolution.

So 0.985 was a real bug, not a method floor — the instinct to flag it was right. Next: checking whether the C++ coupling's normal (reused from NESTOR's `SurfaceGeometry`, which carries `signOfJacobian`) has the same flip, and the delbsq / Shafranov-shift verification.

CharlesCNorton:
Convergence-dynamics status.

Root cause of `|B_ext|/|B_in| = 0.985`: an inward surface normal (`e_θ×e_ζ`) in both the reference and the coupling. Fixed (e3541ad, aed9b6d): the reference reaches 0.9997 against an analytic interior dipole; the coupling's per-point `|B|²/2` is within 1% of NESTOR.

The corrected AEGIS reproduces the poor convergence this issue targets. cth_like, ns=15: converged `delbsq` = 1.67e-3 (AEGIS) vs 1.26e-3 (NESTOR); the gap grows with mpol and AEGIS fails to converge at mpol9 where NESTOR converges. The previously competitive `delbsq` was an artifact of the zeroed casing (coil-only, no plasma structure).

This is not accuracy-limited. Source oversampling (a validated fine-grid virtual casing), QBX offset and order, and quadratic edge-field extrapolation each leave `delbsq` unchanged, and the AEGIS pressure is already within 1% of NESTOR. The limiter is the convergence coupling, consistent with the sheet-current mechanism the issue names: `σ = n·B_plasma` injects the mid-iteration `n·B_total ≠ 0` as a spurious monopole that NESTOR's scalar-potential `B²` construction does not carry. Quantifying that term's effect on the edge-force response next.

CharlesCNorton:
**Standalone AEGIS converges, and its delbsq falls below NESTOR's at high resolution.**

The non-convergence was a near-null edge drift, not an accuracy limit. AEGIS's per-point vacuum pressure agrees with NESTOR to about 1%, but its virtual-casing field is not the self-consistent boundary-value solution NESTOR's elliptic solve produces, so the coupled edge modes are only weakly restored: at high mpol the force residual plateaus near 1e-10 and the equilibrium creeps (delbsq degrading) instead of reaching ftol. Ruled out at the pressure level, each leaving the drift intact: source oversampling, QBX order and offset, the `sigma = n.B` monopole projection, poloidal spectral filtering, temporal under-relaxation, and the interior-field extrapolation slope.

The m-dependent edge preconditioner already in this PR arrests it. Defaulted on (`VMECPP_EDGE_A = 20`) when `VMECPP_AEGIS` is active, pure AEGIS reaches ftol 1e-11 at every mpol, and its converged delbsq crosses below NESTOR at mpol 9 with a margin that grows as NESTOR's Fourier-truncated vacuum field degrades:

| mpol | NESTOR delbsq | AEGIS delbsq |
|-----:|:-------------:|:------------:|
| 5    | 1.26e-3 | 1.48e-3 |
| 7    | 6.0e-4  | 1.17e-3 |
| 9    | 1.02e-3 | 8.9e-4  |
| 12   | 1.44e-3 | 5.3e-4  |
| 16   | 1.46e-3 | 3.4e-4  |

(cth_like free boundary, ftol 1e-11, converged delbsq.) AEGIS improves monotonically with resolution while NESTOR degrades past mpol 9, reaching 4.3x lower delbsq at mpol 16. Below mpol 9 NESTOR remains better, so AEGIS is the high-resolution solver. No NESTOR pressure enters the AEGIS result; a convex blend of the two pressures was tested and loses to NESTOR at mpol 12, so it is not used.

CharlesCNorton:
Next target is the hard case named in the issue: a high-pressure, large-Shafranov axisymmetric equilibrium, where the virtual-casing contribution is largest.

Status. cth_like (stellarator) converges and beats NESTOR at high resolution (above), edge-drift fix default-on. On the axisymmetric Solovev benchmark the coupling is not yet robust, and the cause is now localized to the virtual-casing accuracy on that geometry rather than the coupling stabilization or the physics:

- The AEGIS vacuum pressure is 2.7% off NESTOR there from the first iteration, vs 1.1% on cth_like, and that error grows into a boundary blow-up.
- Signal strength is healthy, not starved: `‖B_plasma‖/‖B_coil‖` = 6–13% on Solovev vs 2.9% on cth_like.
- It is independent of beta: AEGIS fails uniformly from beta 4e-6 up to 1.3e-3 (pres_scale 1 to 300), where NESTOR converges throughout.

The axisymmetric virtual-casing on-surface evaluation is the active work.

CharlesCNorton:
Axisymmetric coupling fixed (4ce722d).

The axisymmetric divergence was a resolution bug, not physics. For `ntor=0` VMEC uses `nZeta=1`, so the virtual-casing Biot-Savart integral over the torus was a single-point toroidal quadrature, and the QBX offset `h = sqrt(area/count)` lumped the whole 2π into that one cell. The source is toroidally invariant there, so rotating it to 256 toroidal angles (rescaling area and `h`) resolves the integral: VC error 2.7% → ~0.5%. AEGIS now converges on the free-boundary Solovev tokamak with no manual tuning, up to β ~ 5e-3, and reproduces the Shafranov shift — raxis 4.09 / 4.20 / 4.54 at pres_scale 1 / 300 / 1000, within 0.03–0.6% of NESTOR. Non-axisymmetric cases are byte-identical.

Regime, measured on both:

- High-mpol stellarator (cth_like): AEGIS beats NESTOR, 4.3× lower delbsq at mpol 16, because NESTOR's Fourier vacuum representation is the limit there.
- Axisymmetric tokamak (Solovev): NESTOR's scalar-potential solve is near-exact, and AEGIS converges to the correct equilibrium but keeps a ~10× higher delbsq. That gap is a systematic QBX on-surface bias, not resolution: NTOR 256 → 1024 cuts the VC random error 3× with no delbsq change.

Next: singularity subtraction, as you suggested, to remove the QBX bias is the lever to close the tokamak accuracy gap.

CharlesCNorton:
Summary and next step.

Completed:
- delbsq below NESTOR at high resolution on cth_like — 4.3× lower at mpol 16, the metric this issue names.
- Both coupled convergence failures fixed: the near-null edge drift (edge preconditioner, default-on) and the single-toroidal-sample virtual-casing integral for axisymmetric plasmas (4ce722d). AEGIS converges on both stellarator (cth_like) and tokamak (Solovev) free-boundary equilibria and reproduces the Shafranov shift to within 0.03–0.6% of NESTOR up to β ≈ 5e-3.

Next:
- Close the axisymmetric-tokamak accuracy gap, where NESTOR's near-exact few-mode solve keeps a lower delbsq. Direct measurement localizes it to the QBX on-surface extrapolation bias: the delbsq tracks the QBX offset and floors where the offset goes singular, while toroidal resolution changes nothing. Replacing QBX with singularity subtraction on the on-surface integral — as suggested above, and as BIEST uses — is the fix. Implementing and validating it now against the magnetized-sphere jump-condition test.

CharlesCNorton:
Singularity subtraction implemented and validated; it is now the default on-surface evaluator (2beea0d).

Replacing QBX extrapolation with singularity subtraction — the principal-value Biot-Savart sum plus the analytic exterior jump J/2 = (σn + K×n)/2 — removes the on-surface extrapolation bias that was the dominant error. The AEGIS converged delbsq now falls below NESTOR at **every** cth_like resolution, not just mpol ≥ 9:

| mpol | NESTOR | AEGIS (QBX) | AEGIS (sing.sub.) |
|-----:|:------:|:-----------:|:-----------------:|
| 5    | 1.26e-3 | 1.48e-3 | 7.0e-4 |
| 7    | 6.0e-4  | 1.17e-3 | 5.4e-4 |
| 9    | 1.02e-3 | 8.9e-4  | 3.4e-4 |
| 12   | 1.44e-3 | 5.3e-4  | 3.0e-4 |
| 16   | 1.46e-3 | 3.4e-4  | 3.1e-4 |

The per-point pressure now matches NESTOR to 0.09% (was 0.5% with QBX). On the axisymmetric Solovev tokamak it improves ~2× (delbsq 6.7e-4 → 3.6e-4 at β = 1.3e-3); NESTOR's near-exact few-mode axisymmetric solve still holds a lower delbsq there. Singularity subtraction is also cheaper than QBX (no expansion centers). QBX remains available via VMECPP_AEGIS_QBX.

Next:
- The physical-surface-current check you raised: verify n × (B_out − B_in) → 0 at convergence where the total-pressure jump vanishes (distinct from the virtual-casing K).
- AEGIS now floors at ~3e-4 on both geometries, flat with mpol — no longer the QBX, so likely the interior-field extrapolation to the LCFS, the one approximation NESTOR does not make. That's the next lever.

CharlesCNorton:
Physical-surface-current check added (4917994).

Per your note: the diagnostic reports |n × (B_out − B_in)| / |B| at the LCFS — the *physical* tangential-field jump, distinct from the virtual-casing K (which is always nonzero). For these P(edge)=0 cases it must vanish at convergence, and it does — decreasing monotonically toward the virtual-casing accuracy floor as the equilibrium converges: cth_like 1.2% → 0.23%, Solovev → 0.22%. It is diagnostic only (the equilibrium is byte-identical), enabled with VMECPP_AEGIS_DIAG.

CharlesCNorton:
**Remaining work — planned next steps.**

The coupling converges and its delbsq falls below NESTOR at every cth_like resolution. The following are the open items and the intended approach for each.

**1. Self-scaling edge preconditioner.** The m-dependent edge damping (`VMECPP_EDGE_A`) defaults to A=20 for non-axisymmetric and A=100 for axisymmetric plasmas — a per-geometry constant sized to arrest the near-null edge drift the coupling exhibits. It should be derived from a measurable operator property rather than set per geometry. Plan: scale it from the LCFS edge-force block's near-null spectrum (measurable through the HVP path merged in #580) so it self-adjusts across geometry and resolution; the coupled-block form is the 2D block-tridiagonal preconditioner in #616.

**2. High-β axisymmetric convergence.** The axisymmetric coupling reaches ftol 1e-11 up to β ≈ 5e-3; at β ≈ 2e-2 (pres_scale 3000, the large-Shafranov regime named in the issue) it plateaus above ftol even at A=100. The virtual-casing contribution is largest there, so the Picard coupling is stiffest. Plan: characterize the plateau and apply direct Picard acceleration on the vacuum-pressure sequence (Anderson or Chebyshev) or a frozen-vacuum inner Newton, rather than raising the mode-diagonal damping further.

**3. Axisymmetric on-surface accuracy.** On ntor=0, NESTOR's few-mode scalar-potential solve is near-exact (delbsq 6.5e-5) and AEGIS floors ~6× higher, matching NESTOR to 0.09% per point. The on-surface principal-value quadrature is saturated: toroidal oversampling (256 → 1024) and near-singular density subtraction both leave that 0.09% unchanged, and the floor rises rather than falls with radial resolution — the signature of a fixed method error, not a quadrature or extrapolation one. The remaining untested candidate is poloidal source aliasing: nThetaEff = 16 at mpol 12 gives a periodic-quadrature Nyquist of m=8, so m=9–11 of the surface-current spectrum is aliased in the integral (the toroidal direction is already oversampled to 256; the poloidal is not). Plan: Fourier-interpolate the source to a finer poloidal grid and measure the axisymmetric floor. If that is also saturated, the floor is the virtual-casing-vs-elliptic method difference on the complete-Fourier axisymmetric case, and AEGIS's operating regime is ntor>0 — where NESTOR's representation is incomplete and AEGIS wins.

**4. Coverage.** Validated on cth_like (nfp=5, ntor=4) and the free-boundary Solovev (ntor=0). Plan: extend to li383 and additional finite-ntor configurations to map the AEGIS-vs-NESTOR delbsq crossover against ntor, which the current two points identify as the governing variable (AEGIS beats NESTOR where NESTOR's Fourier vacuum representation is truncated, and matches-or-trails it where that representation is complete).

**5. Verification report.** Plan: a markdown report with the free-boundary iteration-count-vs-resolution curves, the delbsq-vs-NESTOR tables across mpol, the free-boundary Solovev Shafranov-shift comparison, and the physical-surface-current convergence (`n × (B_out − B_in) → 0`), with reproducible scripts, per the earlier request.

jurasic-pf:
> Next target is the hard case named in the issue: a high-pressure, large-Shafranov axisymmetric equilibrium, where the virtual-casing contribution is largest.

Axisymmetric is a good, simple baseline, but as a stellarator guy, I'm most interested in 3D, high-beta cases of course. Symmetry can mask a lot of mistakes, and things can be wrong but still converge.

jurasic-pf:
For validating the virtual casing solver in isolation, I'd really recommend using BIEST as the "golden" reference. That's the only VC solver that I've seen converge to near machine precision. NESTOR only does for some simple geometries, e.g. Jonathan managed for a solovev case iirc.

If you need more free boundary test cases for validation, I can provide a few in a PR.

CharlesCNorton:
**Free-boundary verification report added** (`examples/aegis_verification/`: markdown report, reproducible scripts, figures).

**Three-dimensional finite-beta, cth_like (nfp=5, ntor=4).** Converged `delbsq` is below NESTOR at every resolution and at both vacuum-level and beta ~ 1% pressure, the margin widening as the grid refines:

| mpol (ns) | NESTOR, beta ~ 1% | AEGIS, beta ~ 1% |
|----------:|:-----------------:|:----------------:|
| 5 (25) | 1.88e-3 | 9.2e-4 |
| 7 (35) | 1.24e-3 | 5.7e-4 |
| 9 (45) | 1.91e-3 | 4.9e-4 |

Across a Shafranov sweep (`pres_scale` 432 to 2160, beta 0.19 to 0.99%, mpol 8) AEGIS and NESTOR agree on total beta and MHD energy to 0.1-2% and both reproduce the outward axis shift, while AEGIS's `delbsq` is 2 to 2.6x lower. The cth_like coil set has a beta limit near 1.5% at this current, so the sweep tops out there.

**Virtual-casing accuracy in isolation.** The on-surface evaluator is checked against an exact reference on the three-dimensional cth_like LCFS: a magnetic dipole placed interior to the plasma, whose exterior-limit field on the boundary is the dipole field itself. AEGIS reconstructs it to a direction cosine of 0.999, with the magnitude converging to unity as the grid refines (0.907 to 0.955 at 48 to 96 poloidal points). This validates the singularity-subtracted quadrature on a stellarator boundary against an analytic solution.

**On the iota difference.** AEGIS and NESTOR converge to iota profiles 5.6% apart on cth_like, and both are internally resolution-converged, so it is a fixed method difference rather than discretization. On the axisymmetric Solovev tokamak, where NESTOR's few-mode scalar-potential vacuum solve is accurate, AEGIS reproduces NESTOR's magnetic axis to 0.1-0.7% and MHD energy to better than 0.1% across a beta sweep. Combined with the analytic-dipole check above, AEGIS is accurate on both geometries; the 5.6% cth_like difference is in NESTOR's Fourier-truncated three-dimensional vacuum representation, consistent with AEGIS's lower `delbsq` there.

**Next.**
- Above beta ~ 1% on cth_like the Picard coupling through the vacuum pressure stiffens and the force residual plateaus; Anderson or Chebyshev acceleration of the vacuum-pressure sequence is the intended remedy rather than stronger mode-diagonal edge damping.
- A BIEST cross-check is set up (`biest_driver.cpp`, generalized virtual-casing operator); its direct-quadrature term reproduces the AEGIS magnitude on the same surface and field, with BIEST's on-surface singular-quadrature convention still being reconciled to the exterior limit used here. The analytic-dipole reference already fixes the accuracy against an exact solution.
- Extending to li383 and additional finite-ntor configurations would map the AEGIS-versus-NESTOR crossover against ntor. The offer of further free-boundary test cases is welcome.

CharlesCNorton:
**Free-boundary verification, consolidated.**

Reported here rather than as a committed document. All runs use the `cth_like` free-boundary stellarator (nfp=5, ntor=4) and the free-boundary Solovev tokamak (ntor=0), at ftol 1e-11; reproducible scripts and the figures are on the branch under `examples/aegis_verification/`.

**1. delbsq below NESTOR at fixed resolution.** cth_like, converged delbsq; the finite-beta column raises `pres_scale` to 2160 (beta ~ 1%).

| mpol (ns) | NESTOR (vac) | AEGIS (vac) | NESTOR (beta 1%) | AEGIS (beta 1%) |
|----------:|:------------:|:-----------:|:----------------:|:---------------:|
| 5 (25)  | 1.26e-3 | 7.0e-4 | 1.88e-3 | 9.2e-4 |
| 7 (35)  | 6.0e-4  | 5.4e-4 | 1.24e-3 | 5.7e-4 |
| 9 (45)  | 1.02e-3 | 3.4e-4 | 1.91e-3 | 4.9e-4 |
| 12 (55) | 1.44e-3 | 3.0e-4 | 2.29e-3 | -      |
| 16 (55) | 1.46e-3 | 3.1e-4 | -       | -      |

AEGIS is below NESTOR at every resolution and both pressure levels, the margin widening as the grid refines (NESTOR's Fourier vacuum representation truncates past mpol 9).

**2. Three-dimensional finite-beta Shafranov sweep.** cth_like, mpol 8, ns 25.

| pres_scale | beta  | raxis (N / A)     | iota_ax (N / A) | delbsq (N / A)  |
|-----------:|:-----:|:-----------------:|:---------------:|:---------------:|
| 432        | 0.19% | 0.77373 / 0.76685 | 1.260 / 1.189   | 6.2e-4 / 3.2e-4 |
| 1080       | 0.48% | 0.77825 / 0.77314 | 1.278 / 1.208   | 9.2e-4 / 3.6e-4 |
| 2160       | 0.99% | 0.78615 / 0.78370 | 1.314 / 1.245   | 1.5e-3 / 5.9e-4 |

beta and MHD energy agree between the solvers to 0.1-2%, both reproduce the outward Shafranov shift, and AEGIS's delbsq is 2 to 2.6x lower. cth_like has a coil beta limit near 1.5% where both solvers fail.

**3. Axisymmetric agreement where NESTOR is accurate.** On the free-boundary Solovev tokamak (prescribed iota), AEGIS reproduces NESTOR's magnetic axis to 0.09%, 0.43%, 0.68% at beta 4e-6, 1.3e-3, 5e-3, and MHD energy to better than 0.1%.

**4. The iota split is a fixed method difference.** On cth_like, NESTOR holds iota_ax at 1.260 and AEGIS at 1.189 across mpol 5 to 12, each flat to better than 0.1%: a 5.6% difference independent of resolution.

**5. Virtual-casing accuracy in isolation.** On an interior-dipole field (its exterior limit is the dipole field itself) on the cth_like LCFS, AEGIS reconstructs the field to a direction cosine of 0.999. BIEST, the high-order boundary-integral solver used in DESC, reconstructs the same dipole to a cosine of 1.000 with rms falling high-order (2.2e-2, 7.6e-3, 2.6e-3 at 48/72/96 poloidal points), confirming it as a near-machine-precision reference. On the real cth_like plasma field, AEGIS agrees with BIEST to a cosine of 0.9998, and the vacuum pressure `|B_ext|^2/2` to 0.1%. (BIEST's generalized virtual-casing operator combines as `B/2 + BiotSavart(n x B) - Laplace(n.B)`; its Laplace kernel carries the sign opposite to the naive gradient.) AEGIS is therefore accurate on both simple and three-dimensional geometry, while NESTOR reaches near-machine-precision vacuum fields only on simple geometry. The 5.6% iota split is NESTOR's Fourier-truncated three-dimensional vacuum representation, consistent with AEGIS's lower delbsq there.

**6. High-beta convergence.** The unscaled AEGIS Picard coupling diverges above beta ~ 1.3% on Solovev, where NESTOR reaches ~ 4.5%. The divergence is the near-null edge drift the edge preconditioner arrests, whose strength grows with beta. Scaling the axisymmetric edge damping by `max(1, K * beta)` (K=150, beta from the thermal-to-magnetic energy ratio) removes the ceiling:

| pres_scale | beta   | converged | delbsq | iters |
|-----------:|:------:|:---------:|:------:|:-----:|
| 1000       | 5.1e-3 | yes       | 2.9e-4 | 2288  |
| 3000       | 2.0e-2 | yes       | 1.2e-3 | 2246  |
| 6000       | 4.3e-2 | yes       | 3.1e-3 | 3675  |
| 8000       | 5.9e-2 | yes       | 4.6e-3 | 4905  |

AEGIS now converges across the full beta range NESTOR reaches on this geometry, and beyond, with no manual tuning. Below the 1/K threshold the damping equals the fixed base, so the low-beta and non-axisymmetric results above are byte-identical. Anderson-beta mixing of the vacuum-pressure sequence is also available (opt-in) but the edge-damping schedule is the effective lever.

**Remaining.** The axisymmetric on-surface delbsq floor, where NESTOR's near-exact few-mode solve keeps a lower value; deriving the edge-damping base and slope from the LCFS force-block spectrum rather than fitting them; and mapping the AEGIS-versus-NESTOR crossover against ntor, which needs the further finite-ntor free-boundary cases offered above (li383 in the test data is fixed-boundary with no coil grid).

CharlesCNorton:
**AEGIS-vs-NESTOR crossover mapped against ntor.**

Truncating the cth_like boundary to ntor = 1..4 (same nfp=5 geometry and coil grid, so the toroidal Fourier content of the free boundary is the only variable) isolates the crossover against toroidal resolution. Converged delbsq, mpol 8, ns 25:

| ntor | NESTOR delbsq | AEGIS delbsq | AEGIS/NESTOR |
|-----:|:-------------:|:------------:|:------------:|
| 1    | 1.16e-2 | 9.14e-3 | 0.79 |
| 2    | 2.25e-3 | 1.67e-3 | 0.74 |
| 3    | 8.89e-4 | 4.70e-4 | 0.53 |
| 4    | 6.22e-4 | 3.21e-4 | 0.52 |

AEGIS is below NESTOR at every ntor >= 1, and the margin widens as the toroidal content grows (0.79 -> 0.52). With the axisymmetric Solovev point (ntor = 0), where NESTOR's near-exact few-mode solve keeps the lower delbsq, the crossover sits at ntor 0 -> 1: AEGIS overtakes NESTOR as soon as the vacuum field carries toroidal structure that NESTOR's Fourier representation must truncate, which is the mechanism the delbsq-vs-mpol result showed, now isolated against ntor rather than resolution.

CharlesCNorton:
**Physical surface current: it tracks the boundary pressure jump, not the virtual casing.**

The physical tangential-field jump `n x (B_out - B_in)` at the LCFS is distinct from the virtual-casing current `K = n x B_plasma`, which is nonzero at every point by construction. On these free-boundary cases the edge total-pressure jump vanishes, so the physical surface current must too, and it does: it sits at the on-surface quadrature floor (~1.7e-3 relative) with no sheet current. Imposing a nonzero edge pressure (a uniform pedestal on the Solovev mass profile, `p(edge) = pres_scale * d`) raises it, so a physical current appears only when there is a pressure jump.

Its quantitative scaling on the axisymmetric case is not cleanly separable from that floor, because the pressure-driven magnitude excess `|B_out|^2 - |B_in|^2 = 2 mu0 p(edge)` is absorbed into AEGIS's residual normal-field error `n.B_ext` rather than the tangential jump. That floor is not a limitation of the singularity-subtracted quadrature itself. On a balanced (isotropic) source grid AEGIS's `n.B_ext` converges to BIEST's high-order value (~1.5e-4, matched to a factor of order one); the coupling's higher floor is the aspect ratio of its axisymmetric source grid (poloidal `nThetaEff ~ 16` against the toroidal upsampling to 256). Rebalancing the toroidal count recovers ~20% of both the surface-current floor and delbsq. The effect is confined to `ntor = 0`, where NESTOR's few-mode solve is already near-exact; the three-dimensional grid is balanced and unaffected.

Reproducible scripts under `examples/aegis_verification/`: `surface_current.py` (current versus edge pressure), `floor_diag.py` (the floor is the poloidal quadrature, falling with mpol and flat with ns), `onsurface_accuracy.py` (AEGIS versus BIEST on a balanced grid), `ntor_aspect.py` (the aspect-ratio dependence).

CharlesCNorton:
**Status, validated against BIEST as the golden reference (per the recommendation above).**

**What we found.** On both cth_like and a 3D high-beta NCSX free-boundary case, AEGIS's *coupled* equilibrium reaches a lower `delbsq` than NESTOR but a **less self-consistent boundary**: the golden BIEST field pierces the AEGIS boundary at `n.B/|B|` ~ 1-2% (NESTOR ~ 0.2%), with an over-shifted magnetic axis. So `delbsq` is not tracking accuracy here - it measures a solver's consistency with *its own* vacuum field, and AEGIS converges tightly onto an over-shifted, non-tangent equilibrium (seen on cth_like at the vacuum-level mpol-16 point too, not only NCSX). The cause is two-fold: VMEC's free-boundary force is tangency-blind (only `|B_ext|^2` enters, so tangency is never driven), and AEGIS's evaluate-only virtual casing (punctured-trapezoidal on-surface quadrature) carries a ~0.2% field-magnitude bias relative to NESTOR's analytic (Merkel) quadrature - small, but enough to move the soft high-beta boundary. NESTOR lands correctly because its scalar-potential solve is tangent by construction.

**Where we are.** Built a reproducible 3D high-beta NCSX case (the large-Shafranov-shift regime this issue names) and the golden-reference comparison that isolates the above. To check whether enforcing tangency is the fix, I also built a tangent-by-construction vacuum coupling (single-layer Neumann solve; verified tangent to machine precision on any boundary, including AEGIS's over-shifted one). It does enforce tangency - but with the same punctured-trapezoidal quadrature it inherits the same ~0.2% magnitude bias and lands at the same over-shifted boundary. So the wall is the vacuum-field *magnitude* accuracy, not the tangency, and it is not resolution-limited (flat across mpol 6-14).

**What remains.** The accurate *and* tangent coupling already exists as BIEST's own `ExtVacuumField::ComputeBplasma`: it solves exactly this exterior Neumann BVP - `grad S[sigma] + curl S[J]` with the total field tangent on the surface - using high-order singular quadrature via GMRES. Wiring that in as the vacuum field, in place of NESTOR/AEGIS inside the VMEC iteration, is the concrete route to a free-boundary solver that is both robust and accurate. That is the "make BIEST converge robustly coupled to VMEC" this issue set out to do, and it is the next build.

jurasic-pf:
Maybe https://github.com/proximafusion/vmecpp/pull/639 is informative too. It includes working prototypes for BIEST as the free boundary solver and a port of vac2. No thorough validation has been performed on them yet, but they reproduce expected shafranov shifts for some exemplary cases.

CharlesCNorton:
> Maybe #639 is informative too. It includes working prototypes for BIEST as the free boundary solver and a port of vac2. No thorough validation has been performed on them yet, but they reproduce expected shafranov shifts for some exemplary cases.

thank you checking it out now

CharlesCNorton:
BIEST and Vac2 are available as native free-boundary solvers in #637 and #638 and supersede the AEGIS virtual-casing solver in this PR. Closing it. The validation findings are recorded in https://github.com/proximafusion/vmecpp/issues/628#issuecomment-4948876953. The m-dependent edge preconditioner is independent of the AEGIS solver (NESTOR-only, default off) and can be submitted separately if useful.

PR #666, opened 2026-07-18
Speed up output-stage post-processing: hoist redundant per-surface trig recompute

jurasic-pf:
## Summary

Investigated why `ComputeOutputQuantities` (the post-solve output stage that
produces the wout/jxbout/threed1/Mercier data) takes a noticeable amount of wall time at high `ns`.

Profiling with an added stage-by-stage checkpoint breakdown (using the
existing `VmecCheckpoint` mechanism) on a W7-X case at `ns=201` showed
`ComputeWOutFileContents` dominating the output stage by far (~600ms of
~530-950ms total, depending on system load), with `LowPassFilterCovariantB`
a distant second.

Inside `ComputeWOutFileContents`, the half-grid Nyquist-mode DFT loop (which
produces `gmnc`, `bmnc`, `bsubumnc`, `bsubvmnc`, `bsubsmns`, `bsupumnc`,
`bsupvmnc`) and the full-grid `bsubsmns_full` loop both recomputed the
`cos(mu - nv)`/`sin(mu - nv)` transform kernel (`tcosi`/`tsini`) from scratch
on every radial surface, even though this kernel depends only on
`(mode, poloidal_index, toroidal_index)` -- not on the surface index. This
is flagged in the code with a `// FIXME(eguiraud) slow loop` comment.

This PR hoists that kernel computation out of the per-surface loop into a
precomputed table, computed once and reused across all surfaces (the
`bsubsmns_full` loop reuses the exact same table as the half-grid loop,
since it needs the identical sin kernel). This is a pure reordering of the
same floating-point operations in the same accumulation order, so it is
bit-exact.

Also adds a `w7x_hires_fastconv` bench case (derived from `examples/data/w7x.json` with `ns_array=[201]`, a loose `ftol_array=[1e-6]`
for fast convergence) to `output_quantities_bench.cc`, so the output stage
can be profiled at a production-like resolution instead of only the small
`cma` case.

### Results

On the `w7x_hires_fastconv` case (`ns=201`, `mpol=ntor=12`):
- Before: `ComputeOutputQuantities` median ~530ms
- After: ~430ms (~19% faster)

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -4689,6 +4689,9 @@ vmecpp::WOutFileContents vmecpp::ComputeWOutFileContents(
 
   // COMPUTE |B| = SQRT(|B|**2) and store in bsq, bsqa
   std::vector<double> magnetic_pressure((fc.ns - 1) * s.nZnT, 0.0);
+#ifdef _OPENMP
+#pragma omp parallel for
+#endif
   for (int jH = 0; jH < fc.ns - 1; ++jH) {
     for (int kl = 0; kl < s.nZnT; ++kl) {
       const int idx_kl = jH * s.nZnT + kl;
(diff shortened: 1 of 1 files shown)


jurasic-pf:
@copilot Continue optimizing the performance of ComputeOutputQuantities and assess based on the w7x_hires_fastconv and cma benchmarks. omp parallelization and re-using the fast toroidal and poloidal dft/fft kernels are two levers to consider.

jurasic-pf:
* **#666** <a href="https://app.graphite.com/github/pr/proximafusion/vmecpp/666?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/proximafusion/vmecpp/666?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/proximafusion?ref=merge-queue-instructions-in-stack-comment&prId=666).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #663, opened 2026-07-17
Seed the vacuum state across free-boundary multigrid transitions

jurasic-pf:
The first iteration of a free-boundary continuation stage skips the whole vacuum block (the iter2 > 1 gate in IdealMhdModel::update, inherited from Fortran VMEC funct3d) while assembleTotalForces still applies the edge term with the freshly zeroed rBSq. The LCFS row therefore takes one step using the unbalanced plasma pressure, inflating the LCFS unphyisically, before the outside vacuum pressure is turned back on the next iteration.

The vacuum solution of the converged coarser stage is still exactly valid at stage entry: the radial interpolation changes neither the angular grid nor the LCFS geometry. A continuation stage is like our "hot restart" in this
respect, and the hot-restart path already sets vacuum_pressure_state_ = kInitialized for exactly this purpose.

Measured (identical converged physics to all printed digits):
- cth_like_free_bdy_multigrid [15,25]: second stage 343 -> 320 iterations (niter regression guard updated 344 -> 321)
- solovev_free_bdy [16,32]: second stage 828 -> 630 iterations (-24%)

PR #667, opened 2026-07-18
Demo: confirm PARVMEC has the same free-boundary multigrid transition bug fixed in #663

jurasic-pf:
## Summary
- Adds `examples/multigrid_transition_demo/`: a W7-X free-boundary case (`ns_array = 4,9,28,99`, a large late jump) run three ways -- VMEC++ with the #663 fix reverted, VMEC++ with the fix applied, and upstream PARVMEC (Fortran) via `xvmec` -- to check whether PARVMEC has the same multigrid-transition vacuum-force bug that #663 fixed in VMEC++.
- Confirms it does: both codebases gate the free-boundary/vacuum force block on `iter2 > 1` (`funct3d.f` in PARVMEC, `ideal_mhd_model.cc` in VMEC++) and reset `iter2 = 1` at every multigrid stage transition (`initialize_radial.f` / `Vmec::InitializeRadial`), so the first iteration of every new (finer) grid always skips the vacuum solve and applies the edge force with a stale/zero vacuum pressure.
- Measured FSQR at iteration 1 of each new stage:

  | ns transition | vmecpp, bug present | vmecpp, fixed | PARVMEC (Fortran) |
  |---|---|---|---|
  | 4 -> 9   | 3.4e-02 | 3.4e-02 | 3.8e-02 |
  | 9 -> 28  | **1.1e+01** | 9.5e-03 | **5.5e+00** |
  | 28 -> 99 | **4.0e+01** | 1.5e-02 | **2.1e+01** |

  PARVMEC shows the same qualitative signature as unfixed VMEC++ -- a residual spike growing with `ns` right at stage entry -- confirming the bug is inherited from Fortran VMEC's flow control rather than VMEC++-specific.

## Test plan
- [x] `python3 examples/multigrid_transition_demo/plot_transition_comparison.py` regenerates the comparison plots and summary table from the checked-in logs
- [x] `pre-commit run --files examples/multigrid_transition_demo/*` passes


jurasic-pf:
## Follow-up: a counterexample where the fix is decisive (not just faster)

Added `examples/multigrid_transition_demo/quasr_counterexample/` — a highly-shaped, large-`ns` free-boundary case that **fails without the #663 fix and converges with it**, exactly the counterexample requested in the mailing-list discussion.

**QUASR-0065579** (nfp=4, free boundary, `mpol=ntor=6`, ftol=1e-9, delt=0.9):
- **Cold start** on a single `ns=201` grid fails immediately (interpolated guess self-intersects) → a multigrid ramp `ns_array=[12,50,201]` is *required*.
- On that ramp:

  | code | `[12,50,201]` outcome |
  |---|---|
  | vmecpp 0.7.0 (fix) | **converges**, FSQR 9.9e-10, 3065 iters |
  | vmecpp 0.6.1 (bug) | **fails** — FSQR diverges to ~9e6 at ns=201, bad-Jacobian abort |
  | PARVMEC (stock) | **fails** — FSQR spikes to ~4e3 at ns=201 entry, 37 Jacobian resets, hits iteration cap |

**Fix isolated to #663**: PARVMEC already has the restart behavior 0.7.0 adopted yet still fails, so that isn't what saves 0.7.0. Reverting *only* the 18-line #663 vacuum-seed patch from current source flips it from converge → fail at the exact 50→201 transition (`logs/fix663_reverted_ns12-50-201.log`).

Also a "just faster" case (**QUASR-0029346**, `[8,16,31]`): both versions converge, 0.7.0 in ~1.8x fewer total iterations (2.8x on the final grid).

Field provenance: runs use the released 0.6.1/0.7.0 wheels; the on-disk mgrid handed to PARVMEC was validated to reproduce the in-memory response table bit-for-bit. See the directory README and `quasr_counterexample.png`.

PR #690, opened 2026-08-20
Add equilibrium rescaling support via HotRestartState

### LINKED ISSUE #475: Rescaling (suggested feature)
This is a feature suggestion, not a bug. 
It is common to "rescale" configurations by either enlarging the surface by a constant multiple or multiplying the field strength by a constant. 
In these cases, the equilibrium can be rescaled instead of recomputed. 

For instance, if `B -> cB` then `J -> cJ` and `gradp -> c^2 gradp`. 
It could be convenenient to have this utility in VMEC++. 
See [DESC's rescale](https://desc-docs.readthedocs.io/en/stable/_api/compat/desc.compat.rescale.html#desc.compat.rescale) for example.

duushgu:
This PR adds support for efficiently rescaling a converged VMEC equilibrium ($B \to c_B \cdot B$, $R \to c_R \cdot R$) without re-computing the post-processed output individually, as discussed in #475.

To do this, we leverage the existing `HotRestartState` pipeline. Instead of trying to scale all 100+ derived `wout` quantities (which is tedious and error-prone), we only scale the fundamental geometric boundaries and input variables. A run with 0 iterations from this state will then cleanly reconstruct all derived physics arrays!

**Physics implemented:**
When scaling the magnetic field by `b_scale` and the geometry by `r_scale`, we scale:
*   **Geometry:** $R$ and $Z$ Fourier coefficients by `r_scale`. (Note: The poloidal stream function $\lambda$ / `lmns` is dimensionless and left untouched).
*   **Toroidal flux ($\Phi$):** `phiedge` by `b_scale * r_scale^2`.
*   **Toroidal current ($I$):** `curtor` by `b_scale * r_scale`.
*   **Pressure ($p$):** `pres_scale` by `b_scale^2` (to keep force balance $\nabla p = j \times B$). This is optional via the `scale_pressure` flag.

**Implementation Details:**
1. Added `Rescale(b_scale, r_scale, scale_pressure=True)` method to `OutputQuantities`.
2. Exposed `rescale` to Python via pybind11.
3. Users can now easily do this in Python:
   ```python
   oq = vmecpp.OutputQuantities.load("wout.h5")
   oq.rescale(b_scale=2.0, r_scale=10.0, scale_pressure=True)
   # Run with 0 iterations to re-evaluate the full equilibrium metrics
   new_oq = vmecpp.run(oq.indata, initial_state=vmecpp.HotRestartState(oq))
   ```

*(Note: I am a physic student and used AI assistance for the C++ / pybind11 boilerplate. I'd love your feedback on whether this `HotRestartState` pipeline perfectly aligns with your architectural goals or if `lmns` scaling behaves differently under your specific coordinate definition!)*


jurasic-pf:
Hello @duushgu , i like the idea of using the hot restart state to implement the rescaling feature! Thanks for looking into it. 
Since the majority of users are calling vmecpp from Python and the operations are relatively cheap, I would suggest implementing the feature on the python side instead or at least exposing it from there (the current state does not contain the python api from your pr description).

Additionally pre-commit checks are fialing, they are a minimal set of checks we enforce at each commit and can be run automatically before committing. by installing 
pip install pre-commit
pre-commit install

The python side API should ideally accept either VmecOutput and/or VmecWout and return the same rescaled, by calling vmecpp.run with zero steps internally, and not leave that to the user. 
Finally a unit test comparing the result to a full VMEC rerun at different scale or some other simple sanity check would be appreciated

src/vmecpp/_rescale.py
```diff
@@ -0,0 +1,75 @@
...
+
+    scaled_input.raxis_c *= r_scale
+    scaled_input.zaxis_s *= r_scale
+    if scaled_input.raxis_s is not None:
+        scaled_input.raxis_s *= r_scale
+    if scaled_input.zaxis_c is not None:
+        scaled_input.zaxis_c *= r_scale
+
+    # Force 0 iterations for the restart
+    scaled_input.niter_array = np.zeros_like(scaled_input.niter_array)
```
jurasic-pf: hot restart runs have to be initialized from the final resolution of a multigrid vmec run.
e.g. when using number of surfaces 5,13,100 then the hot restart should start from 100

The niter, ftol, and ns arrays should therefore all just contain the last element, check out the hot restart example script as a reference

src/vmecpp/_rescale.py
```diff
@@ -0,0 +1,75 @@
...
+    if scaled_wout.rmns is not None:
+        scaled_wout.rmns *= r_scale
+    if scaled_wout.zmnc is not None:
+        scaled_wout.zmnc *= r_scale
+
+    # Create intermediate VmecOutput for restart_from
+    # (The C++ side only uses wout and indata for HotRestartState, so it's fine if the rest is unscaled)
+    intermediate_output = output.model_copy(deep=True)
+    intermediate_output.input = scaled_input
+    intermediate_output.wout = scaled_wout
```
jurasic-pf: This can be a shallow copy, since we deep copied the wout already, and are now just replacing it

tests/test_rescale.py
```diff
@@ -0,0 +1,69 @@
...
+    # Betas should be invariant under this scaling (pressure ~ B^2)
+    np.testing.assert_allclose(
+        oq_rescaled.wout.betatotal, oq_initial.wout.betatotal, rtol=1e-5
+    )
+
+    # 4. Ensure hot-restart rescale matches a full run from scratch on scaled inputs
+    # The whole point of rescale is that it avoids iterating but gives the exact same result!
+    np.testing.assert_allclose(oq_rescaled.wout.bmnc, oq_full_run.wout.bmnc, rtol=1e-4)
+    np.testing.assert_allclose(oq_rescaled.wout.rmnc, oq_full_run.wout.rmnc, rtol=1e-4)
+    np.testing.assert_allclose(oq_rescaled.wout.zmns, oq_full_run.wout.zmns, rtol=1e-4)
```
jurasic-pf: Nit: Why are all the tolerances so loose? I would have expected this to match to high precision, e.g. 1e-9.

There's a few absolute tolerances that may introduce scale dependence, if thats the source I can make them relative as a separate issue.

tests/test_rescale.py
```diff
@@ -0,0 +1,69 @@
...
+TEST_DATA_DIR = REPO_ROOT / "src" / "vmecpp" / "cpp" / "vmecpp" / "test_data"
+
+
+def test_equilibrium_rescale():
+    # Load and run a simple initial equilibrium
+    vmec_input = vmecpp.VmecInput.from_file(TEST_DATA_DIR / "solovev.json")
+    # Speed up the test by reducing resolution/iterations if necessary, but solovev is fast
+    oq_initial = vmecpp.run(vmec_input)
+
+    # Scaling factors
```
jurasic-pf: Please remove comments where the code is self explanatory.

tests/test_rescale.py
duushgu: The reason the relative error (rtol) blew up at higher precision is exactly the absolute scale dependence near zero that you suspected. Some of the physical coefficients (like bmnc) contain values on the order of 1e-12. When comparing values that tiny, the relative error mathematically explodes even though the absolute difference is negligible (~1e-15).

duushgu:
Thanks for the great review! You are completely right about the hot restart arrays. Passing np.ones_like forced VMEC to unnecessarily run through all the old multigrid stages again.
 
I am applying all your suggestions right now:

1. Reducing niter_array, ftol_array, and ns_array to a 1D array with just the last element.
2. Removing deep=True for the intermediate output since a shallow copy is enough.
3. Tightening the test tolerances to 1e-9 (with atol=1e-10).
4. Removing the self-explanatory comments.

I will push the updated commit shortly!

jurasic-pf (approved):
Thanks! The remaining ci errors are due to a dependency issue, not your change. Ill resolve them and merge in the upcoming week

PR #691, opened 2026-08-21
Remove Python compatibility Layer

duushgu:


jurasic-pf (commented):
Hi, Thanks for looking into this, but I think there was a misunderstanding about the intent of the issue.

If you want to help out on the issue, we can chat, but i think it requires more architectural decisions that need to be discussed first.

duushgu:
Hi @jurasic-pf, completely agree! After closing this PR, I did some deep diving into the downstream ecosystem (raytrax, constellaration) and realized just how heavily they rely on the Pydantic layer for DX, deepcopy, and serialization. 

I realized this requires a much broader design discussion, so I just opened a new Draft PR (#692) where I prototyped keeping a Pydantic frontend isolated in simsopt_compat, and laid out the two architectural paths (Option A vs. Option B) for us to discuss. 

I'd love to help out with whichever direction the core team prefers. Let's move the chat over to #692!

Issue #208, opened 2025-04-02
Remove `VmecINDATAPyWrapper` and use `VmecINDATA` directly


jurasic-pf:
The C++ class  `VmecINDATAPyWrapper` is primarily there to handle converting between std::vectors which don't work too nicely with pybind11 and Eigen3 datatypes which do. Once we ported all these classes to Eigen3, the wrapper should disappear.

Issue #206, opened 2025-04-02
Remove Python compatibility Layer


jurasic-pf:
A lot of code is just there to handle the subtle differences between the Fortran datatype conventions and VMEC++ conventions and convert between them. We would like to get rid of this in the long run.

PR #692, opened 2026-08-21
WIP/Discussion: Refactor Python Compatibility Layer Ecosystem (Issue #206)

duushgu:
I explored removing the Python compatibility layer as originally requested in #206 and sub-issue #208. 

During the migration, I successfully adapted downstream repositories (`raytrax` and `constellaration`) locally to use a preserved Pydantic frontend. To make CI and local tests pass, the downstream ecosystems would need to import and serialize like this:

```python
# Downstream usage example
import vmecpp
import vmecpp.simsopt_compat

# Pydantic DX is preserved via simsopt_compat
pydantic_input = vmecpp.simsopt_compat.VmecInput(...)

# Explicitly cast to PyBind11 C++ before passing to the backend
cpp_indata = pydantic_input._to_cpp_vmecindata()
cpp_wout = vmecpp.run(cpp_indata)
```

**The Architectural Crossroads:**
While this works and passes tests, it merely relocates the technical debt: `simsopt_compat.py` is now heavily bloated (nearly 3000 lines), essentially just moving the Pydantic layer out of `__init__.py`. 

Before pushing PRs to the downstream repositories, I wanted to align on the preferred path forward:

- **Option A (Current Draft PR state):** Keep the Pydantic wrappers alive but isolate them entirely inside `simsopt_compat.py`. Downstream adapts to explicit C++ serialization.
- **Option B (Deep Refactoring):** Truly delete the Pydantic wrapper layer entirely. Invest instead in extending the PyBind11 C++ classes to support pythonic behavior natively (e.g., proper `__copy__`, `__repr__`, JSON dict serialization, and generating `.pyi` stubs for IDE autocomplete).

Would love the core team's input on which direction we should take for the ecosystem!


jurasic-pf (changes requested):
Hi, Thanks for looking into this, but I think there was a misunderstanding about the intent of the issue.

We want to remove the adapter layers on the C++ side NOT change the Python api that people are building on. 

If you want to help out, we can chat, but this requires a few deliberate, human, architectural decisions that need some discussion first, which is why they've not been prioritized yet.

duushgu:
Thanks for your answer. How can we chat? here? I have not a lot expierence with githubXD. why have they not been prioritized yet? Do you mean "the Python api that people are building on" other repos? 

My production of english language is not so good. Maybe my german language is much better. I will start my physics bachelor from this october in germany. I loved math, but want to solve real world problems

I wanted to learn new things like working in github. I learned a little bit python, html and css during german course and tryed some IT-Tools in Linux. Yesterday I watched a video about Proximafusion, saw the website and motivated very. I want volunteerly to contribute sometimes when i have a time, because there is things to do with Physics like magnetismus, plasma etc. and clear energie "one of the greatest goals of humanity". Maybe tommorow i will try to play with softwares and experience as a user.

PR #695, opened 2026-08-23
Compile AVX2 variant selected at load based on CPU support

jurasic-pf:
Builds the computation core twice and lets the dynamic loader choose: baseline at `vmecpp/cpp/libvmecpp_core.so`, optimized at `vmecpp/cpp/glibc-hwcaps/x86-64-v3/`. glibc >= 2.33 searches the subdirectory first; older loaders do not know about it and resolve the baseline.

Read more about the approach https://www.kvr.at/posts/easy-dynamic-dispatch-using-GLIBC-hardware-capabilities/

| case | delta |
|---|---|
| w7x (fixed boundary, FFT path) | **-15.8%** |
| free boundary (NESTOR) | -10.2% |
| small fixed boundary | -5.4% |

**Changes**
- `vmecpp_core` becomes `SHARED` — glibc-hwcaps cannot apply to the extension module, which CPython opens by exact path — and is instantiated per ISA via a `vmecpp_add_core()` helper.
- FFTX codelets are built per variant; they are the toroidal-transform hot path and must match the ISA of the core linking them.
- Pins `EIGEN_MAX_ALIGN_BYTES=32`. `-march` raises Eigen's alignment guarantee 16 -> 32, which changes `Eigen::aligned_allocator` and is ABI-affecting; without the pin, mixing cores built with different `-march` gives correct results and then corrupts the heap at teardown.
- `VMECPP_HWCAPS_DISPATCH=OFF` builds baseline only.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,11 @@ else()
   message(STATUS "LTO unavailable, building without it: ${VMECPP_IPO_ERROR}")
 endif()
 
+# -march raises Eigen's EIGEN_MAX_ALIGN_BYTES from 16 to 32, which changes
+# Eigen::aligned_allocator and is ABI-affecting. Pin it so cores built with
+# different -march settings stay interchangeable.
+add_compile_definitions(EIGEN_MAX_ALIGN_BYTES=32 EIGEN_MAX_STATIC_ALIGN_BYTES=32)
+
 # use ccache if available
 find_program(CCACHE_COMMAND NAMES ccache ccache-swig)
 if(EXISTS ${CCACHE_COMMAND})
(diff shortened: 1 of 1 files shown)


jurasic-pf:
Built locally with cibuildwheel. auditwheel handles the hwcaps layout correctly

```
vmecpp/cpp/libvmecpp_core.so                          ymm=0     fma=0
vmecpp/cpp/glibc-hwcaps/x86-64-v3/libvmecpp_core.so   ymm=9855  fma=6511
both SONAME=libvmecpp_core.so

_vmecpp*.so   RPATH=[$ORIGIN:$ORIGIN/../../vmecpp.libs]   NEEDED=libvmecpp_core.so
baseline core RPATH=[$ORIGIN/../../vmecpp.libs]           (2 levels up)
v3 core       RPATH=[$ORIGIN/../../../../vmecpp.libs]     (4 levels up, depth-adjusted)
```

- `glibc-hwcaps/` preserved, core not hash-renamed (that only applies to external libs it grafts in).
- Our `$ORIGIN` was **appended to**, not replaced.
- The nested variant's own RPATH was correctly depth-adjusted.
- Installing the wheel selects the v3 core; hiding `glibc-hwcaps/` falls back to baseline.

<details>
<summary>Verification script used</summary>

```python
#!/usr/bin/env python3
"""Assert a wheel still carries its baseline and glibc-hwcaps library variants.

A wheel that lost its optimized variants still imports and still passes every
test, so nothing else catches it.

Usage: check_wheel_hwcaps.py dist/*.whl
"""

from __future__ import annotations

import argparse
import re
import subprocess
import sys
import tempfile
import zipfile
from pathlib import Path

LIB_DIR = "vmecpp/cpp"
LIB_NAME = "libvmecpp_core.so"
EXTENSION_GLOB = "vmecpp/cpp/_vmecpp*.so"
# Instructions that must not appear in a baseline library, per hwcaps level.
FORBIDDEN_IN_BASELINE = {"x86-64-v3": r"%ymm|\bvfm(add|sub|nmadd|nmsub)"}


def _readelf(path: Path) -> str:
    return subprocess.run(
        ["readelf", "-dW", str(path)], capture_output=True, text=True, check=True
    ).stdout


def _disasm(path: Path) -> str:
    return subprocess.run(
        ["objdump", "-d", "--no-show-raw-insn", str(path)],
        capture_output=True,
        text=True,
        check=True,
    ).stdout


def _tag(text: str, key: str) -> str | None:
    m = re.search(rf"{key}.*\[([^\]]+)\]", text)
    return m.group(1) if m else None


def check(wheel: Path) -> list[str]:
    errors: list[str] = []
    with tempfile.TemporaryDirectory() as tmp:
        root = Path(tmp)
        with zipfile.ZipFile(wheel) as zf:
            zf.extractall(root)

        base = root / LIB_DIR / LIB_NAME
        if not base.is_file():
            return [f"missing baseline {LIB_DIR}/{LIB_NAME}"]

        variants = sorted((root / LIB_DIR / "glibc-hwcaps").glob(f"*/{LIB_NAME}"))
        if not variants:
            return [f"no glibc-hwcaps variants of {LIB_NAME} (flattened?)"]

        base_soname = _tag(_readelf(base), "SONAME")
        base_text = _disasm(base)
        for var in variants:
            level = var.parent.name
            rel = var.relative_to(root)
            if var.read_bytes() == base.read_bytes():
                errors.append(f"{rel} is identical to the baseline")
            if _tag(_readelf(var), "SONAME") != base_soname:
                errors.append(f"{rel} SONAME differs from baseline; not substitutable")
            forbidden = FORBIDDEN_IN_BASELINE.get(level)
            if forbidden and re.search(forbidden, base_text):
                errors.append(
                    f"baseline contains instructions reserved for {level}; "
                    f"it would fault on older CPUs"
                )
            if forbidden and not re.search(forbidden, _disasm(var)):
                errors.append(f"{rel} has no {level} instructions; baseline copy?")

        ext = next(iter(root.glob(EXTENSION_GLOB)), None)
        if ext is None:
            errors.append(f"no extension module matching {EXTENSION_GLOB}")
        else:
            dyn = _readelf(ext)
            if LIB_NAME not in dyn:
                errors.append(f"extension does not link {LIB_NAME}; static-linked?")
            rpath = (_tag(dyn, "R(?:UN)?PATH") or "").split(":")
            resolved = {
                (ext.parent / e.replace("$ORIGIN", ".")).resolve() for e in rpath if e
            }
            if base.parent.resolve() not in resolved:
                errors.append(
                    f"no extension RPATH entry resolves to {LIB_DIR}; the core "
                    f"and its variants are unreachable (RPATH={rpath})"
                )
    return errors


def main() -> int:
    ap = argparse.ArgumentParser(description=__doc__)
    ap.add_argument("wheels", nargs="+", type=Path)
    failed = False
    for wheel in ap.parse_args().wheels:
        errors = check(wheel)
        failed |= bool(errors)
        print(f"{'FAIL' if errors else 'ok  '} {wheel.name}")
        for e in errors:
            print(f"  - {e}")
    return 1 if failed else 0


if __name__ == "__main__":
    sys.exit(main())
```

Usage: `python check_wheel_hwcaps.py wheelhouse/*.whl`. Passes on the genuine artifact; fails on hand-broken copies (hwcaps directory removed, baseline copied over the variant, RPATH entry stripped).

</details>

PR #699, opened 2026-08-23
Repair shared-core and HWCAPS regressions

krystophny:
Repair five regressions exposed on current main:

- propagate the shared core link interface so standalone and conda consumers link;
- build glibc HWCAPS only on Linux x86-64, rejecting unsupported explicit configurations;
- use the native relative loader path for the shared core inside macOS wheels;
- materialize the overlapping Eigen wdot expression so results do not depend on SIMD packet width;
- let Python 3.14 CI use the corrected SIMSOPT 1.11.0 source release while retaining the AVX-512 wheel guard on Python 3.10–3.13.

The checked-in wdot goldens encode the former aliasing bug. Numerical coverage is replaced by a scalar energy-history oracle; round-trip, dimensions, dtype, and every other golden variable remain checked. No binary golden data is changed.

Validation:
- old v3 code fails the scalar oracle at AVX packet boundaries; fixed v3 passes
- baseline and v3 wdot agree within 1.22e-14
- native baseline/v3 core, standalone, and Python extension build and import
- generated Darwin wheel paths resolve the adjacent core with @loader_path; Linux retains its native loader token
- simulated Darwin/arm64 defaults HWCAPS off; explicit enable fails early
- dependency resolution keeps SIMSOPT 1.10.6 on Python 3.10–3.13 and selects the corrected 1.11.0 sdist on Python 3.14
- pre-commit and focused Python/C++ checks pass


krystophny:
Current main is merged into this PR and the five bounded fixes are complete: shared-core consumer links, Linux HWCAPS gating, the macOS loader path, alias-safe `wdot`, and the Python 3.14 SIMSOPT constraint. All 31 checks pass, including Python 3.14, ASAN, UBSAN, and TSAN. Ready for review.

Chris&GPT

jurasic-pf (approved):
Thanks Chris! Also a neat constraint to make CI green again while waiting for the simsopt release

CMakeLists.txt
```diff
@@ -137,11 +137,11 @@ function(vmecpp_add_core target)
...
-  target_link_libraries(${target} PRIVATE ${HDF5_CXX_LIBRARIES} ${HDF5_LIBRARIES})
-  target_link_libraries(${target} PRIVATE ${netCDF_LIBRARIES})
-  target_link_libraries(${target} PRIVATE nlohmann_json::nlohmann_json)
-  target_link_libraries(${target} PRIVATE LAPACK::LAPACK)
-  target_link_libraries(${target} PRIVATE absl::algorithm absl::base
+  target_link_libraries(${target} PUBLIC ${HDF5_CXX_LIBRARIES} ${HDF5_LIBRARIES})
+  target_link_libraries(${target} PUBLIC ${netCDF_LIBRARIES})
+  target_link_libraries(${target} PUBLIC nlohmann_json::nlohmann_json)
+  target_link_libraries(${target} PUBLIC LAPACK::LAPACK)
+  target_link_libraries(${target} PUBLIC absl::algorithm absl::base
```
jurasic-pf, line 144: WHy do these have to change visibility?

tests/test_init.py
```diff
@@ -341,6 +352,17 @@ def test_against_reference_wout(indata_file, reference_wout_file, path_type):
...
+        if varname == "wdot":
+            # See the scalar energy-history oracle in output_quantities_test.cc;
+            # the checked-in golden records the former overlapping assignment.
+            np.testing.assert_equal(
+                np.asarray(test_value[:]), fortran_wout.wdot, err_msg=error_msg
+            )
+            assert test_value.dtype == expected_value.dtype, (
+                error_msg + f" (dtype {test_value.dtype} != {expected_value.dtype})"
+            )
+            continue
```
jurasic-pf: I would also prefer if we didn't exclude wdot from the reference wout comparison and instead fixed the issue at the core (which to my understanding this PR is doing?) and if necessary regenerated the reference

jurasic-pf:
I merged updated reference files and reverted the test_init changes

PR #703, opened 2026-08-25
geometry: expose the product-basis contract to Python and JAX

krystophny:
Stack 2/4. Targets `main`; merge after #584.

## Scope

This PR exposes the #584 geometry contract to Python and JAX:

- pybind access to the coefficient snapshot and C++ evaluator
- a JAX pytree with the same product-basis layout
- a JAX evaluator for the same value and second-order spatial jet

The layer evaluates a supplied equilibrium snapshot. It does not run VMEC, define an objective, or implement an adjoint.

## Tests

- compare the JAX evaluator with the independent C++ evaluator on a solved equilibrium
- verify that JAX supplies geometry-state and coordinate VJPs

Stack: #584 -> **#703** -> #707 -> #710.


krystophny:
image

jurasic-pf:
Super cool, excited to discuss tomorrow!

krystophny:
Rescoped to match its own description. This PR is now **exactly the geometry view**: the pybind exposure of the product-basis contract plus `vmecpp.geometry`, and nothing else.

- `qs.py` / `test_qs.py` moved up to #707, where they belong. The stack content was off by one against the PR descriptions; that is fixed, and #710 is no longer a duplicate of #707.
- The pybind block added here is only what `vmecpp.geometry` calls: the coefficient/dimension structs, `GeometryPoint`, and `make_geometry`.
- Dropped a gratuitous reordering of `add_subdirectory(geometry)` and of the pybind Bazel deps that only produced diff noise.

Worth noting on the C++ `Geometry.evaluate` binding: it is not there to be re-exported to users. `tests/test_geometry.py` uses it as an **independent oracle** for the JAX evaluator on a real solved Solovev equilibrium, which is the only reason it is bound at all.

Now +1075 / -0. Verified: `test_geometry.py` passes against an Enzyme-enabled build.

src/vmecpp/geometry.py
```diff
@@ -0,0 +1,260 @@
...
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+from vmecpp.cpp import _vmecpp  # type: ignore
+
+
+@jax.tree_util.register_pytree_node_class
+@dataclasses.dataclass(frozen=True)
+class Geometry:
```
jurasic-pf, line 16: Great

jurasic-pf (approved):
This part of the stack is very nice, pending downstack approvals

krystophny:
Restacked onto current `main` and the updated #584; the #703-specific patch is unchanged by stable patch ID. `uv run --extra test pytest -q tests/test_geometry.py` passes (2/2). Full CI is running; merge after #584.

PR #706, opened 2026-08-25
Remove vendored LAPACK dependency: replace dgetrf_/dgetrs_ with Eigen::PartialPivLU

jurasic-pf:
vmecpp's calls exactly two LAPACK routines, dgetrf_ and dgetrs_ in LaplaceSolver. Because of that, we bundle a full LAPACK library with the wheel.

This PR replaces both calls with Eigen::PartialPivLUEigen::MatrixXd, removing one more dependency.

Benchmarks continue to report similar numbers.

### DIFF
--- a/.github/actions/run-benchmarks/action.yml
+++ b/.github/actions/run-benchmarks/action.yml
@@ -31,7 +31,7 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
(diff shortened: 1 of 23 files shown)


jons-pf:
Nit: might want to remove lapack from list of packages to install in readme as well?

jurasic-pf:
Totally, I just wanted to see if this is feasible and the performance impact is tolerable. The answer seems yes, I'll clean up the PR

PR #707, opened 2026-08-25
qs: implement quasisymmetry as a Python geometry objective

krystophny:
Stack 3/4. Targets `main`; merge after #703.

## Scope

This PR implements SIMSOPT's `QuasisymmetryRatioResidual` as a Python/JAX objective over `vmecpp.geometry`.

The residual includes the flux-surface measure, covariant current averages, iota, and the requested helicity. It contains no C++, pybind, Enzyme, VMEC solver change, or QS-specific native kernel. QS is the demonstration objective for the generic geometry interface.

## Tests

- reconstruct `|B|` and compare it with VMEC's independent `bmnc` spectrum
- compare the residual vector and total with SIMSOPT on a 3D equilibrium
- differentiate the scalar objective with JAX

The SIMSOPT comparison gives a residual-vector cosine above 0.9999 and a total within `1e-3` relative tolerance.

Stack: #584 -> #703 -> **#707** -> #710.

PR #710, opened 2026-08-26
autodiff: run VMEC++ with an implicit geometry VJP

krystophny:
Stack 4/4. Targets `main`; merge after #707.

## Scope

This PR wraps a VMEC++ solve in `jax.custom_vjp`. JAX differentiates any scalar objective over the #703 geometry pytree, while VMEC++ supplies the exact residual transpose for the implicit solve.

The first parameterization is intentionally limited to a stellarator-symmetric, fixed-boundary, prescribed-iota equilibrium with `ncurr=0`. The differentiable input is the dense `(rbc, zbs)` boundary array. Free-boundary, asymmetric, pressure, current, and profile derivatives remain outside this PR.

The native additions expose the live geometry map, its transpose, and the exact forward/transposed force operators needed to remove structural null modes and solve the adjoint. The objective remains Python/JAX code; #707 supplies QS as the end-to-end example.

## Efficiency relative to finite differences

Let `P = 2 mpol (2 ntor + 1)` be the number of boundary coefficients. A centered finite-difference gradient needs `2P` nonlinear equilibrium solves. `jax.value_and_grad` here uses one forward solve, one rerun in the reverse callback, and one transposed linear solve. The nonlinear solve count is independent of `P`.

| case | P | centered FD solves | adjoint nonlinear solves |
|---|---:|---:|---:|
| 3D test equilibrium (`mpol=6`, `ntor=1`) | 36 | 72 | 2 |
| CTH input (`mpol=5`, `ntor=4`) | 90 | 180 | 2 |

This table compares nonlinear solve counts, not wall-clock time; the adjoint also pays for its transposed linear solve.

## Tests

- live solver output matches the native geometry snapshot and works under `jax.jit`
- the 3D state-to-geometry transpose satisfies the dot-product identity
- the 3D solve VJP agrees with an independently computed forward sensitivity using `H` versus `H^T`
- a 3D QS objective differentiates end to end
- a non-Enzyme build raises instead of using finite differences

Enzyme-enabled result: `5 passed, 1 skipped`; the skipped case is the non-Enzyme fallback test. Both Enzyme kernel CTests pass.

Stack: #584 -> #703 -> #707 -> **#710**.


krystophny:
This PR now actually contains the solve VJP - it was previously an exact duplicate of #707 because the stack content was shifted by one. Fixed.

Together with #707 this is the `custom_vjp` boundary you asked for: `vmecpp.autodiff` wraps a solve-like function (not `vmecpp.run`), the objective above it is ordinary JAX over the geometry pytree, and the reverse rule hands the cotangent to the exact force-residual transpose.

Trimmed the pybind surface to exactly what `vmecpp.autodiff` calls. Removed as unused:

| removed | why |
|---|---|
| `force_jacobian_vector_product` | never called; only the transpose is used |
| `exact_hessian_vector_product` (+ `_transpose`) | duplicate binding names for the same C++ methods |
| `_kernel_jvp`, `_kernel_vjp`, `_kernel_dims` | no consumer anywhere in the repo |
| `composed_force_residual`, `force_density_jvp_residual` | diagnostics with no consumer |
| `ExactHessianVectorProduct` (forward C++ method) | unreachable after the above |

I checked each of these against every `.py` in the repo including `examples/` and `benchmarks/` before removing it. What is left is `get_geometry`, `geometry_state_vjp`, `set_freeze_constraint_multiplier`, `have_to_flip_theta`, `has_exact_force_jacobian` and `force_jacobian_transpose_vector_product`.

Now +1929 / -0 for the whole stack. Verified against an Enzyme-enabled Clang build (`ClangEnzyme-22`): `test_geometry` + `test_qs` + `test_autodiff` = 7 passed, 1 skipped (the no-Enzyme fallback test, correctly skipped); `test_hessian`, `test_internal_gradient`, `test_preconditioner` still pass, confirming nothing depended on the removed bindings.

krystophny:
Restacked the complete series onto current `main`. Focused local validation passes: geometry + QS + autodiff tests are 9 passed, 2 correctly skipped without Enzyme. Full CI is running. Ready for review after #584, #703, and #707 land.

jurasic-pf:
Can you create a follow-up PR, suggesting how we should ship these as part of pip wheels? I guess we need to get cibuildwheel to use enzyme + clang and be careful that the resulting binary doesnt show performance regression for non adjoint users

jurasic-pf (approved):

jurasic-pf:
The quasisymmetry module moves out of the core package in #707: `vmecpp/qs.py` becomes `vmecpp/_qs.py`, re-exported from `vmecpp.simsopt_compat`. Patch and the three follow-through changes needed here (docstring example in `autodiff.py`, import in `tests/test_autodiff.py`) are in https://github.com/proximafusion/vmecpp/pull/707#issuecomment-5576668372.

Merging the updated #707 branch into `pr707-solve-vjp` is a clean rename merge. Local run on the merged result with a GCC build: 7 passed, 2 skipped, the skips being the Enzyme-gated cases.

src/vmecpp/autodiff.py
```diff
@@ -0,0 +1,463 @@
...
+
+def make_solver(vmec_input) -> DifferentiableVmec:
+    """Return a JAX-compatible callable that runs VMEC++ for vmec_input.
+
+    Example:
+
+        from vmecpp import autodiff
+
+        solver = autodiff.make_solver(input)
+        objective = lambda boundary: qs.quasisymmetry_total(solver(boundary), [0.6])
```
jurasic-pf:
```suggestion
        from vmecpp import autodiff, simsopt_compat

        solver = autodiff.make_solver(input)
        objective = lambda boundary: simsopt_compat.quasisymmetry_total(
            solver(boundary), [0.6]
        )
```

tests/test_autodiff.py
```diff
@@ -0,0 +1,283 @@
+from typing import Any
+
+import jax
+import jax.numpy as jnp
+import numpy as np
+import pytest
+from scipy.sparse.linalg import LinearOperator, gmres
+
+import vmecpp
+from vmecpp import autodiff, geometry, qs
```
jurasic-pf, line 10:
```suggestion
from vmecpp import autodiff, geometry
from vmecpp import simsopt_compat as qs
```

jurasic-pf (commented):
Follow-through for the quasisymmetry move in #707: `vmecpp/qs.py` becomes `vmecpp/_qs.py`, re-exported from `vmecpp.simsopt_compat`. Merging the updated #707 branch into `pr707-solve-vjp` is a clean rename merge; the two remaining references are suggested inline.

Local run on the merged result (GCC build): 7 passed, 2 skipped, the skips being the Enzyme-gated cases including `test_quasisymmetry_gradient_through_a_three_dimensional_solve`.

PR #715, opened 2026-08-29
Make the vacuum-grid overrun an error instead of a stderr warning

CharlesCNorton:
`MGridProvider::interpolate` clamped the vacuum field to the grid edge and wrote a warning to stderr when the plasma boundary left the mgrid domain. The clamped field no longer represents the coils outside the grid, but the run continued to convergence on it, so the effect showed up as a shifted boundary rather than as a domain error. In #639 that appeared as a roughly 12 cm discrepancy against another solver, and that PR's notes conclude that "Plasma Boundary exceeded Vacuum Grid Size" warnings should be treated as fatal for scan purposes. This makes them fatal in the solver, so a scan does not have to enforce that by convention.

`interpolate` now returns `absl::Status`. The status is carried through `ExternalMagneticField::update`, `FreeBoundaryBase::update` (now `absl::StatusOr<bool>`), `Nestor` and `OnlyCoils`, past every collective barrier, and reduced across the nested vacuum OpenMP team through a new `HandoverStorage::vacuum_status` slot before it is acted on. The barrier at the end of `interpolate` stays unconditional, so a thread whose own tangential slice left the grid still reaches it along with the rest of the team.

The code is `kFailedPrecondition`, so `indata.return_outputs_even_if_not_converged` still recovers the run and returns best-effort output. The message names the boundary extents against the grid extents in R, Z, or both. Those extents are taken over the whole boundary rather than over the reporting thread's slice, so the text is the same whichever thread wins the reduction.

The second commit is in `Vmec::SolveEquilibrium`'s aggregation helper. `absl::InternalError` takes an `absl::string_view`, so the `std::move` there did not move and the first reporting thread's message was appended a second time by the code below the `if`; the header string was assigned to a local that was never passed anywhere. With seven radial threads the aggregate held eight copies of the message and no header.

Four Bazel targets also gained a direct `@abseil-cpp//absl/status` dependency, which they had been picking up transitively.

Verification: `//vmecpp_large_cpp_tests/free_boundary/mgrid_provider:mgrid_provider_test` covers in-grid, on-edge, R-only, Z-only, both-axes, the clamped values, per-slice scoping, cross-slice message identity, and the fixed-field path, plus a nested-team case that deadlocks if the barrier is ever made conditional on the slice being inside the grid. Free-boundary and fixed-boundary outputs are unchanged bit-for-bit against the parent commit.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.h
+++ b/src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.h
@@ -31,24 +31,7 @@ class MGridProvider {
                              const Eigen::VectorXd& fixed_bp,
                              const Eigen::VectorXd& fixed_bz);
 
-  // Interpolate the vacuum magnetic field onto the tangential grid points
-  // [ztMin, ztMax) of the plasma boundary. nZnT is the number of tangential
-  // grid points on the whole boundary, of which [ztMin, ztMax) is this
-  // thread's slice.
-  //
-  // Returns a kFailedPrecondition status if the boundary left the domain of
-  // the vacuum field grid. The interpolated field is still written in that
-  // case, clamped to the grid edge, so that the caller sees finite values and
-  // can carry the status to a safe place before acting on it; the clamped
-  // field is not physically meaningful and must not be used to advance the
-  // equilibrium.
-  //
-  // The status describes this thread's tangential slice only. The caller runs
-  // this inside the nested vacuum OpenMP team, so the result has to be reduced
-  // across that team before it is acted upon. The extents named in the error
-  // message, on the other hand, are taken over the whole boundary [0, nZnT),
-  // so that every thread that reports produces the same message no matter
-  // which one wins the reduction.
+  // Interpolate [ztMin, ztMax) of nZnT points; error if outside the grid.
   [[nodiscard]] absl::Status interpolate(int ztMin, int ztMax, int nZeta,
                                          int nZnT, const Eigen::VectorXd& r,
                                          const Eigen::VectorXd& z,
--- a/src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.h
+++ b/src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.h
@@ -8,6 +8,7 @@
 #include <Eigen/Dense>
 #include <filesystem>
 
+#include "absl/status/status.h"
 #include "vmecpp/common/makegrid_lib/makegrid_lib.h"
 #include "vmecpp/common/sizes/sizes.h"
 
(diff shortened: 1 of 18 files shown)


jurasic-pf (changes requested):
Good idea, we should definitely prevent returning a free boundary result where any element is evaluated outside the grid AT CONVERGENCE. I'm not concerned about transient situations that extrapolate. Could you please change it back to a warning if it happens during the run and an error only if the result would be outside? 

Also the test coverage seems a little excessive for such a simple change, I'd prefer if we can trim it down to the most important aspects of public API to reduce maintainance burden.

CharlesCNorton:
Both done.

`IdealMhdModel::update` no longer returns the status. It warns and the run continues. `Vmec::run` checks `h_.vacuum_status` after the multigrid loops and before `ComputeOutputQuantities`. That slot is reset before every vacuum solve, so a boundary that left the grid and came back has already cleared it by then, and only a boundary that is still outside at convergence produces an error. `return_outputs_even_if_not_converged` still returns best-effort output with the warning.

Measured on `cth_like_free_bdy`, whose mgrid box is R in [0.45, 1.05]:

| case | exit | warnings | outcome |
| --- | --- | --- | --- |
| unmodified | 0 | 0 | converges, MHD energy 5.057896e-02 |
| initial boundary shifted outside the box | 0 | 67 | converges, MHD energy 5.057899e-02 |
| `pres_scale` x5 | 0 | 0 | converges, MHD energy 5.000395e-02 |
| `pres_scale` x10 | error | 629 | converges, then `FAILED_PRECONDITION`, boundary R max 1.064146 |
| `pres_scale` x10, `return_outputs_even_if_not_converged` | 0 | 629 | converges, warns, writes the output file |

The second row is the transient case: the initial guess sits outside the box, the free boundary relaxes back inside, and the run now completes and lands on the same equilibrium as the unmodified case. The fourth row is the Shafranov shift carrying the converged boundary out of the box at higher beta, which is the case the error is for.

Raising it from `Vmec::run` also means it no longer passes through the per-thread aggregation in `SolveEquilibrium`, so it arrives once as `FAILED_PRECONDITION` rather than as repeated `INTERNAL` copies.

The interpolation tests are down from ten cases to three: the accepted path, one error case that also pins the reported extents to the whole boundary rather than to the reporting thread's slice, and the mixed in-grid/out-of-grid team case, which deadlocks if the barrier at the end of `interpolate` is ever made conditional on the slice being inside the grid.

Free-boundary and fixed-boundary outputs remain bit-for-bit identical to the parent commit. The diff is now +283/-59.

src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.h
```diff
@@ -30,10 +31,13 @@ class MGridProvider {
                              const Eigen::VectorXd& fixed_bp,
                              const Eigen::VectorXd& fixed_bz);
 
-  void interpolate(int ztMin, int ztMax, int nZeta, const Eigen::VectorXd& r,
-                   const Eigen::VectorXd& z, Eigen::VectorXd& m_interpBr,
-                   Eigen::VectorXd& m_interpBp,
-                   Eigen::VectorXd& m_interpBz) const;
+  // Interpolate [ztMin, ztMax) of nZnT points; error if outside the grid.
+  [[nodiscard]] absl::Status interpolate(int ztMin, int ztMax, int nZeta,
```
jurasic-pf, line 35: i thought abseil status always has nodiscard, isnt this redundant?

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
```diff
@@ -423,6 +423,22 @@ absl::StatusOr<bool> Vmec::run(const VmecCheckpoint& checkpoint,
...
+    if (!indata_.return_outputs_even_if_not_converged) {
+      return h_.vacuum_status;
+    }
+    if (verbose_) {
+      std::cout << absl::StrFormat(
+          "WARNING: %s\n"
+          "return_outputs_even_if_not_converged is set, so returning "
+          "best-effort (likely unphysical) output for debugging purposes.\n",
+          h_.vacuum_status.message());
+    }
```
jurasic-pf: This is covered by a negative return code, i don't think this branch needs a specific error message

jurasic-pf (approved):
Nice!

src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.h
CharlesCNorton: Only `StatusOr` carries it. In the abseil we build against, statusor.h has `class [[nodiscard]] StatusOr;` but status.h declares `class ABSL_ATTRIBUTE_TRIVIAL_ABI Status final` with no must-use attribute; the `ABSL_MUST_USE_RESULT Status` line nearby is inside a comment block. So dropping it here would let a caller discard the out-of-grid status silently, which is the case the change exists to prevent. Kept.

src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
CharlesCNorton: Done in 31680fe6. Dropped the message and set `status_ = VmecStatus::UNRECOVERABLE_ERROR` instead, matching what `SolveEquilibrium` already does on its recovery path.

src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.h
jurasic-pf: Ah, I see.

jurasic-pf:
not sure why tsan is failing

PR #716, opened 2026-08-30
Fix the sign of the perpendicular current v component in jxbout

CharlesCNorton:
`jperp2` in the jxbout output is the flux-surface average of `|J_perp|^2`, built from the contravariant components of `J_perp = (B x grad p) / |B|^2`. With `grad p = p' grad s` those are

```
J_perp^u = + B_v p' / (sqrt(g) |B|^2)
J_perp^v = - B_u p' / (sqrt(g) |B|^2)
```

and `kperpv` was computed with a plus sign. Since

```
|J_perp|^2 = k^u k^u g_uu + 2 k^u k^v g_uv + k^v k^v g_vv
```

the two squared terms are unaffected and only the `g_uv` cross term takes the wrong sign. That is why the error is invisible for axisymmetric configurations, where `g_uv` vanishes, and shows up only in 3D.

Measured against the educational_VMEC reference data used by `JxBOutputContentsTest`, over all 121 compared points:

| case | ratio before | ratio after |
| --- | --- | --- |
| solovev | 1.000000 | 1.000000 |
| solovev_no_axis | 1.000000 | 1.000000 |
| cth_like_fixed_bdy | 1.057109 .. 1.148439 | 1.000000 |
| cth_like_fixed_bdy_nzeta_37 | 1.057109 .. 1.148439 | 1.000000 |
| cth_like_free_bdy | 1.063766 .. 1.160695 | 1.000000 |

Worst deviation from the reference goes from 0.160695 to 0.000000.

Four assertions had been loosened to accommodate this and now hold at the strict per-case tolerance, which is 1e-12 to 5e-11 depending on the case: `jperp2` was at 0.2, the `s2` geometric-and-magnetic integral at 0.2, `jpar_perp` at 0.15 and `jparPS_perp` at 0.1. They move together because `jperp2` feeds `s2` through the volume integral and feeds both ratios through `loc_jpar_perp`.

This resolves the two markers at those sites, `FIXME(jons): This looks SO MUCH off that there actually could be an error still present...` and `TODO(jons): Is this catastrophic cancellation again?`. Neither was cancellation.

The quantity is confined to the jxbout and threed1 diagnostic outputs and does not enter the equilibrium iteration, so converged equilibria are unchanged.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -2630,7 +2630,8 @@ vmecpp::JxBOutFileContents vmecpp::ComputeJxBOutputFileContents(
           vmec_internal_results.bsubv(jHi * s.nZnT + kl);
 
       kperpu[kl] = 0.5 * (bsubv_outside + bsubv_inside) * pprime / sqgb2[kl];
-      kperpv[kl] = 0.5 * (bsubu_outside + bsubu_inside) * pprime / sqgb2[kl];
+      // J_perp = (B x grad p)/|B|^2, so the v component carries a minus sign.
+      kperpv[kl] = -0.5 * (bsubu_outside + bsubu_inside) * pprime / sqgb2[kl];
 
       // --------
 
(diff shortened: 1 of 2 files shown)


jurasic-pf (approved):
well done, thanks for the fix

jonathanschilling:
Wow, super nice that you finally found what that bl**dy error was, thanks a lot!

PR #718, opened 2026-08-31
Report Nestor's scalar magnetic potential as wout potvac

CharlesCNorton:
`potvac`, the Fourier coefficients of Nestor's scalar magnetic potential, is declared in `WOutFileContents` and serialized, but never populated. It is the potential the Laplace solver already converges on each free-boundary iteration, held in `bvecShare`.

The wout layout is the lasym size `2 * mnpd`, with `mnpd = (2 * ntor + 1) * (mpol + 2)`. A stellarator-symmetric run solves only the `sin(mu - nv)` half, so the `cos(mu - nv)` half stays zero. A fixed-boundary run leaves the field empty.

Checked against `wout_cth_like_free_bdy.nc`:

```
size              126 = 126
nonzero entries    58 = 58
max |difference|  2.751e-15   against a scale of 5.790e-04   (4.75e-12 relative)
ratio to reference on significant entries: min = max = median = 1.0000
```

The solver's coefficients are the wout field as-is, with no sign, ordering or normalization change.

`wout_cth_like_free_bdy.nc` is present in `vmecpp_large_cpp_tests/test_data` but absent from that case's `filegroup`, so it never reaches a test's runfiles. The new test needs to read it, so it is added to the filegroup here.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -894,8 +894,8 @@ absl::Status vmecpp::WOutFileContents::WriteTo(H5::H5File& file) const {
   WRITEMEMBER(equif);
   WRITEMEMBER(curlabel);
   WRITEMEMBER(potvac);
-  WRITEMEMBER(xm_pot);
-  WRITEMEMBER(xn_pot);
+  WRITEMEMBER(xmpot);
+  WRITEMEMBER(xnpot);
   WRITEMEMBER(xm);
   WRITEMEMBER(xn);
   WRITEMEMBER(xm_nyq);
@@ -1064,8 +1064,8 @@ absl::Status vmecpp::WOutFileContents::LoadInto(WOutFileContents& m_obj,
   READMEMBER(equif);
   READMEMBER(curlabel);
   READMEMBER(potvac);
-  READMEMBER_OPTIONAL(xm_pot);
-  READMEMBER_OPTIONAL(xn_pot);
+  READMEMBER_OPTIONAL(xmpot);
+  READMEMBER_OPTIONAL(xnpot);
   READMEMBER(xm);
   READMEMBER(xn);
   READMEMBER(xm_nyq);
@@ -4554,12 +4554,12 @@ vmecpp::WOutFileContents vmecpp::ComputeWOutFileContents(
       wout.potvac.head(vacuum_potential.size()) = vacuum_potential;
     }
 
-    // Nestor walks mn with m fastest; xn_pot carries the nfp factor, like xn.
-    wout.xm_pot = Eigen::VectorXi::Zero(mnpd);
-    wout.xn_pot = Eigen::VectorXi::Zero(mnpd);
+    // Nestor walks mn with m fastest; xnpot carries the nfp factor, like xn.
+    wout.xmpot = Eigen::VectorXi::Zero(mnpd);
+    wout.xnpot = Eigen::VectorXi::Zero(mnpd);
     for (int mn = 0; mn < mnpd; ++mn) {
-      wout.xm_pot[mn] = mn % (mf + 1);
-      wout.xn_pot[mn] = (mn / (mf + 1) - nf) * s.nfp;
+      wout.xmpot[mn] = mn % (mf + 1);
+      wout.xnpot[mn] = (mn / (mf + 1) - nf) * s.nfp;
     }
   }
 
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -4531,7 +4531,21 @@ vmecpp::WOutFileContents vmecpp::ComputeWOutFileContents(
 
   wout.equif = threed1_first_table.radial_force;
 
-  // TODO(jons): curlabel, potvac: once free-boundary works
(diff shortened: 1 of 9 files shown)


src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
```diff
@@ -4532,7 +4532,21 @@ vmecpp::WOutFileContents vmecpp::ComputeWOutFileContents(
 
   wout.equif = threed1_first_table.radial_force;
 
-  // TODO(jons): curlabel, potvac: once free-boundary works
```
jons-pf, line 4535: Please excuse the (potentially) incomplete TODO here. We also need `xm_pot` and `xn_pot` for the mode numbers corresponding to the `potvac` Fourier coefficients. Are they already written to the `wout`, and if not, would you mind adding them as well?

CharlesCNorton:
They were not written, and Fortran VMEC does not write them either: the reference `wout_cth_like_free_bdy.nc` carries only `mnpd = 126` and `potvac(mnpd)`, with no mode numbers alongside it.

Added as `xm_pot` and `xn_pot`, length `mnpd`, in Nestor's ordering:

```
xm_pot[mn] = mn % (mf + 1)
xn_pot[mn] = (mn / (mf + 1) - nf) * nfp
```

`nf = ntor`, `mf = mpol + 1`. `xn_pot` carries the `nfp` factor, as `xn` and `xn_nyq` do. For `cth_like_free_bdy` that is 63 modes, m in 0..6 and n in -20..20 stepping by 5.

These cannot reuse `xm`/`xn`. Those are m-major with the m=0 row truncated to n >= 0, while Nestor walks a full rectangle with m fastest. `potvac` keeps its `2 * mnpd` length: sin(mu - nv) over those 63 modes, then cos(mu - nv).

`CheckVacuumPotential` now asserts that the mode set is exactly that rectangle, that each pair occurs once, that every `xn_pot` is a multiple of `nfp`, that m advances fastest, and that the cos half of `potvac` is zero for a stellarator-symmetric run.

Reading the fields back needed a `READMEMBER_OPTIONAL`. HDF5 files written before the fields existed have no such dataset, and the plain `READMEMBER` throws on them. Three hot-restart and fourier-geometry tests read checked-in state files and failed until that guard was added.

`potvac` reaches `WOutFileContents`, HDF5 and the pybind bindings, but not the `VmecWOut` pydantic model, so it does not appear in the netCDF wout that Python writes. `tests/test_init.py` lists `potvac` in `_MISSING_FORTRAN_VARIABLES` for that reason. Wiring it through needs conditional handling: `potvac` is empty for a fixed-boundary run, and NETCDF3_CLASSIC cannot hold a zero-length fixed dimension. I can do that here or in a follow-up.

src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
CharlesCNorton: Not written, and Fortran VMEC does not write them either: the reference `wout_cth_like_free_bdy.nc` has only `mnpd = 126` and `potvac(mnpd)`, no mode-number companion. Added in e571a05d as `xm_pot` and `xn_pot`, length `mnpd`, in Nestor's ordering:

```
xm_pot[mn] = mn % (mf + 1)
xn_pot[mn] = (mn / (mf + 1) - nf) * nfp
```

`nf = ntor`, `mf = mpol + 1`, and `xn_pot` carries the `nfp` factor like `xn` and `xn_nyq`. For `cth_like_free_bdy` that is 63 modes, m in 0..6, n in -20..20 by 5. They cannot reuse `xm`/`xn`, which are m-major with the m=0 row truncated to n >= 0 while Nestor walks a full rectangle with m fastest.

Details are in the PR comment.

jons-pf:
Great, thanks!
I recall `xm_pot` and `xn_pot` from PARVMEC now: https://github.com/ORNL-Fusion/PARVMEC/blob/master/Sources/Input_Output/wrout.f#L514
and also in there, the mode number arrays are called `xmpot` and `xnpot` (pardon me for the spurious `_`):
https://github.com/ORNL-Fusion/LIBSTELL/blob/master/Sources/Modules/read_wout_mod.f90#L38
I'd suggest to therefore stick with `xmpot` and`xnpot` for 1:1 compatibility with PARVMEC.

I really like the proposed unit tests.
Best approach imho for integration-testing these: back-port these into `educational_VMEC`, re-generate reference `wout.nc` files, then test against them. WDYT?

I'd be fine with including the changes required to `VmecWOut` in here for keeping these related changes bundled in this PR.

CharlesCNorton:
Renamed to `xmpot`/`xnpot` in 80ebaeac. Length `mnpd`, matching PARVMEC `xmpot(1:mnpd)`.

`VmecWOut` changes included. All three are `None` for a fixed-boundary run, so the netCDF writer omits them, as Fortran does. Checked on `cth_like_free_bdy` and `solovev` with a save/reload round-trip.

Two writer fixes were needed: the hardcoded `mn_mode_pot = 100` placeholder would have padded `xmpot` from 63 to 100, and pydantic does not strip `Annotated` inside a union, so `SerializeIntAsFloat[...] | None` lost its dimension name.

PARVMEC writes `potsin`/`potcos` instead of `potvac`; ours follows the older schema in the reference file. Can change that here if you want it.

On educational_VMEC: the mode numbers are derived from mpol, ntor and nfp, so a regenerated reference mostly re-tests the formula. PARVMEC already writes `xmpot`/`xnpot`, so a PARVMEC wout would pin the ordering without touching educational_VMEC.

PR #719, opened 2026-08-31
Reject unusable profile parameterization names at input validation

CharlesCNorton:
A profile parameterization name that VMEC++ does not recognize reaches the solver as a zero profile. The run converges and returns a plausible-looking equilibrium with the wrong physics.

On `solovev`, changing `pmass_type` from `power_series` to the typo `power_serise`:

| input | exit | MHD energy | beta |
| --- | --- | --- | --- |
| unmodified | 0 | 2.548352e+00 | 3.8723e-06 |
| `pmass_type: "power_serise"` | 0 | 2.548362e+00 | 0.0 |
| `pmass_type: "sum_atan"` | 0 | 2.548362e+00 | 0.0 |

`sum_atan` is a real parameterization, but only for the current and iota profiles. Both cases print `profile parameterization '---invalid---' not implemented yet` to stderr and carry on.

`IsConsistent` now rejects both, along with a spline parameterization whose knots are missing or whose `aux_s` and `aux_f` differ in length:

```
input variable 'pmass_type' is 'power_serise', which is not a known profile parameterization
input variable 'pmass_type' is 'sum_atan', which cannot be used for the mass/pressure profile
'pmass_type' is 'cubic_spline', which is a spline profile, so 'am_aux_s' and 'am_aux_f' must be given
```

`piota_type` is checked only for `ncurr == 0` and `pcurr_type` only for `ncurr == 1`, matching where each is consulted.

The polynomial coefficient arrays `am`, `ai` and `ac` are deliberately not required to be non-empty. They are zero-padded on read, so an empty array is a valid way to specify a zero profile, and `near_axis_iota_nfp4` ships that way with `ncurr = 1` and an empty `ac`.

Reaching the parameterization table from `IsConsistent` required moving it out of the private `RadialProfiles::ALL_PARAMS` member into the `profile_parameterization_data` module, behind `AllProfileParameterizations()`, `FindProfileParameterization()` and `IsProfileParameterizationAllowedFor()`. Entry order is unchanged, so `findParameterization` still indexes it by `ProfileParameterization` enumerator. That move accounts for most of the diff.

All sixteen inputs under `vmecpp/test_data` still validate.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/BUILD.bazel
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/BUILD.bazel
@@ -23,6 +23,7 @@ cc_library(
     deps = [
         ":boundary_from_json",
         "//vmecpp/common/util:util",
+        "//vmecpp/vmec/profile_parameterization_data",
         "//util/hdf5_io",
         "//util/file_io",
         "@abseil-cpp//absl/algorithm:container",
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
@@ -21,6 +21,7 @@
 #include "util/json_io/json_io.h"
 #include "vmecpp/common/util/util.h"
 #include "vmecpp/common/vmec_indata/boundary_from_json.h"
+#include "vmecpp/vmec/profile_parameterization_data/profile_parameterization_data.h"
 
 namespace {
 [[noreturn]] void ErrorToException(const absl::Status& status,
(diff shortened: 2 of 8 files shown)


jons-pf:
Nice!
nit: I'd suggest to check both pcurr and piota, independent of ncurr, since piota is used as initial guess for the iota profile even in case of a current-constrained (ncurr=1) run. The reverse it not true, i.e., in an iota-constrained case (ncurr=0), the pcurr-profile is ignored. However, for good measure, I'd still recommend to always check all three of pmass, piota and pcurr for valid parameterization IDs.

CharlesCNorton:
@jons-pf Done in 89a1d718. `piota_type` and `pcurr_type` are now checked unconditionally; `ncurr == 0` keeps only the `bloat` check, and `ncurr == 1` has no profile-specific branch left. The test loops over both `ncurr` values and asserts each of the three rejects a name that is valid for a different profile.

jurasic-pf (approved):

PR #720, opened 2026-08-31
Cover the non-stellarator-symmetric makegrid reference case

CharlesCNorton:
`mgrid_test_non_symmetric.nc` and its `coils.test_non_symmetric` / `extcur.test_non_symmetric` inputs are checked in, but the only reference to them anywhere is the `filegroup` in `test_data/BUILD.bazel`. No test reads them, so `assume_stellarator_symmetry = false` has no coverage against the Fortran MAKEGRID reference.

`MakegridReferenceTestFixture` hard-coded the coils file, the symmetry flag and the toroidal grid size, so it could only ever run `coils.test_symmetric_even`. Those three move into `MakegridReferenceTestParams`, defaulted to the existing values, and the non-symmetric case is added to both the B-field and the vector-potential suites:

```
SymmetricEven/CheckComputeMagneticFieldResponseTable.MatchesFortranReference/Raw
SymmetricEven/CheckComputeMagneticFieldResponseTable.MatchesFortranReference/Scaled
NonSymmetric/CheckComputeMagneticFieldResponseTable.MatchesFortranReference/Raw
SymmetricEven/CheckComputeVectorPotentialCache.MatchesFortranReference/Raw
SymmetricEven/CheckComputeVectorPotentialCache.MatchesFortranReference/Scaled
NonSymmetric/CheckComputeVectorPotentialCache.MatchesFortranReference/Raw
```

Both new cases match the reference at the suite's existing 1e-6 tolerance with no change to `makegrid_lib` itself. Only the raw variant is added, since there is no scaled reference for the non-symmetric coil set.

The third reference file, `mgrid_test_symmetric_odd.nc`, stays unused. Its namelist is `LSTELL_SYM = .TRUE.` with `KP = 19`, and `MakeCylindricalGrid` rejects an odd toroidal grid size whenever stellarator symmetry is requested:

```
number of toroidal grid points has to be even for being able to make use to
stellarator symmetry in makegrid, but was num_phi=19
```

Fortran MAKEGRID evidently supports that combination. Wiring up this reference therefore needs the half-period mirror to handle a grid with no exact midpoint, which is left for separate work.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
+++ b/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
@@ -804,9 +804,11 @@ INSTANTIATE_TEST_SUITE_P(
       return info.param.normalize_by_currents ? "Scaled" : "Raw";
     });
 
-// TODO(jons): implement test of non-stellarator-symmetric mgrid file
 // TODO(jons): implement test of stellarator-symmetric mgrid file using
-// symmetric C++ implementation
+// symmetric C++ implementation. mgrid_test_symmetric_odd.nc is the reference,
+// but MakeCylindricalGrid rejects an odd number_of_phi_grid_points whenever
+// stellarator symmetry is requested, so the half-period mirror needs to handle
+// a grid with no exact midpoint first.
 
 // TODO(jons): add test of WriteMakegridNetCDFFile
 // -> in particular, make sure that the consistency of number of serial circuits
--- a/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
+++ b/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
@@ -437,6 +437,12 @@ absl::StatusOr<std::vector<bool>> IsTooCloseToCurrentCarrierForComparison(
 struct MakegridReferenceTestParams {
   bool normalize_by_currents;
   std::string reference_nc_file;
+  // Defaults describe coils.test_symmetric_even; the other cases override them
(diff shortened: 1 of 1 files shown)


src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
```diff
@@ -768,6 +790,20 @@ INSTANTIATE_TEST_SUITE_P(
...
+                             "mgrid_test_non_symmetric.nc",
+        .coils_file =
+            "vmecpp/common/makegrid_lib/test_data/coils.test_non_symmetric",
+        .assume_stellarator_symmetry = false,
+        .number_of_phi_grid_points = 18}),
+    [](const ::testing::TestParamInfo<MakegridReferenceTestParams>& info) {
+      return info.param.normalize_by_currents ? "Scaled" : "Raw";
+    });
+
 // TODO(jons): implement test of non-stellarator-symmetric mgrid file
```
jurasic-pf: is this resolved then?

jurasic-pf (approved):
Thanks!

src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib_test.cc
CharlesCNorton: Yes, removed in 88706f41. The one below it stays: its reference is `mgrid_test_symmetric_odd.nc` with `KP = 19`, and `MakeCylindricalGrid` rejects an odd `number_of_phi_grid_points` when stellarator symmetry is requested, so the half-period mirror has to handle a grid with no exact midpoint first. I moved that reason into the comment.

PR #727, opened 2026-08-31
Check the mgrid bilinear interpolation against an exactly bilinear field

CharlesCNorton:
`MGridProvider::interpolate` is what supplies the vacuum field to the whole free-boundary path, and nothing checked the value it returns. `MGridInterpolationTest` covers the out-of-grid status and the OpenMP barrier; `LoadMGridTest` compares the loaded response table against ABSCAB at the grid nodes, which is the table, not the interpolation between nodes.

A field that is exactly bilinear in R and Z on a phi plane is reproduced exactly by bilinear interpolation, so the check is an equality at arbitrary points rather than a tolerance on a smooth field. The fixture loads a synthetic response table through `LoadFields` with `a + b R + c Z + d R Z` on each plane, different coefficients per component and per plane, on a grid with `numR != numZ`, and evaluates at grid nodes, both extreme corners, cell centers, and points at no particular position in a cell.

That pins the corner weights, the `(phi, z, r)` linear index, which of R and Z each weight belongs to, and the phi plane `kl % nZeta` selects. Each of the following fails it: swapping `pr` and `qz`, swapping `w12` and `w21`, dropping the `w22` term from `w11`, transposing `numR` and `numZ` in the index, and shifting the plane index by one.

Two more:

- out-of-grid points return the field at the clamped position, so a point past a corner picks up that corner rather than an extrapolation;
- `SetFixedMagneticField` overrides a loaded response table and hands back the requested `[ztMin, ztMax)` slice verbatim. Reading that slice from the wrong offset fails it.

### DIFF
--- a/src/vmecpp/cpp/vmecpp_large_cpp_tests/free_boundary/mgrid_provider/BUILD.bazel
+++ b/src/vmecpp/cpp/vmecpp_large_cpp_tests/free_boundary/mgrid_provider/BUILD.bazel
@@ -16,6 +16,7 @@ cc_test(
         "//util/testing:numerical_comparison_lib",
         "//vmecpp/common/magnetic_configuration_lib:magnetic_configuration_lib",
         "//vmecpp/common/magnetic_field_provider:magnetic_field_provider_lib",
+        "//vmecpp/common/makegrid_lib:makegrid_lib",
         "//vmecpp/common/vmec_indata:vmec_indata",
     ],
 )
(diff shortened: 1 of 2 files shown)


jurasic-pf:
Hmmm, I frankly don't see that much value in this test. A check against reference seems good enough to me, could you explain the reasoning a bit more? I read the description, and understand the invariant, I just don't get what we guard against. bilinear is simple enough and verified that I'd just leave it uncovered

CharlesCNorton:
You're right. I mutated the interpolation and ran the existing suite without this test to see what it already catches:

- swapping `w12` and `w21`: 13 targets fail, among them `vmec/vmec:vmec_test`, `free_boundary/nestor:nestor_test`, `free_boundary/laplace_solver:laplace_solver_test`, `vmec/output_quantities:output_quantities_test` and `vmec_hot_restart_test`
- dropping the `w22` term from `w11`: the same 13

So the weights and the index arithmetic are already pinned by the free-boundary regressions several times over, and a unit test only localizes a failure those would report anyway.

The one mutation that survives the whole suite is reading the `SetFixedMagneticField` slice from the wrong offset. That is because `SetFixedMagneticField` has no callers anywhere in the repository, so the branch it guards in `interpolate` is unreachable and covering it would be covering dead code.

Closing.

PR #728, opened 2026-08-31
Test the tangential partitioning of the vacuum surface

CharlesCNorton:
`TangentialPartitioning` had no test. It is what hands every thread its `[ztMin, ztMax)` slice of the `nZnT` surface points, so `SurfaceGeometry`, `RegularizedIntegrals`, `SingularIntegrals`, `LaplaceSolver`, `ExternalMagneticField` and `MGridProvider::interpolate` all write into the range it produces and then read the whole surface. `RadialPartitioning`, the other half of the decomposition, has had one since the beginning.

Six tests:

- the ranges of a team tile `[0, nZnT)` in thread order, and the block sizes take only `nZnT / num_threads` and one more than that, swept over point counts from 0 to 360 against team sizes from 1 to 64, including teams larger than the surface;
- the default arguments give a single-threaded team the whole range;
- an evenly divisible split gives equal blocks;
- the remainder goes to the lowest thread ids, `[0,3) [3,6) [6,8) [8,10)` for 10 points over 4 threads;
- a team larger than the surface leaves the surplus threads with an empty range at `nZnT` rather than an inverted or out-of-range one;
- `adjustPartitioning` re-partitions in place and keeps the thread id.

Each of these fails it: comparing `thread_id_ <= work_remainder` instead of `<`, shifting `ztMin` one past its block, dropping the remainder offset on the threads that do not get an extra point, rounding the block size up, and giving the remainder to the highest thread ids instead of the lowest.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/free_boundary/tangential_partitioning/BUILD.bazel
+++ b/src/vmecpp/cpp/vmecpp/free_boundary/tangential_partitioning/BUILD.bazel
@@ -11,3 +11,13 @@ cc_library(
         "//vmecpp/common/sizes:sizes",
     ],
 )
+
+cc_test(
+    name = "tangential_partitioning_test",
+    srcs = ["tangential_partitioning_test.cc"],
+    deps = [
+        ":tangential_partitioning",
+        "@googletest//:gtest_main",
+    ],
+    size = "small",
+)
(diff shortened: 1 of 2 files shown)


jurasic-pf:
I think the logic is simple enough not to warrant a test

PR #729, opened 2026-08-31
Fix the toroidal mirror of the magnetic axis guess for an odd nZeta

CharlesCNorton:
`RecomputeMagneticAxisToFixJacobianSign` searches the toroidal planes from 0 to `nZeta / 2` and fills the rest by the mirror `zeta -> -zeta`, index `nZeta - k`. The mirror loop runs `k < nZeta / 2`, which is one short for an odd `nZeta`: at `nZeta = 37` it writes planes 36 down to 20 and leaves plane 19 at the zero from `resize`, which then enters the toroidal Fourier transform of the new axis. `cth_like_fixed_bdy_nzeta_37` has that resolution.

The bound becomes `k <= (nZeta - 1) / 2`. Even `nZeta` is unaffected, where the half-period plane is its own mirror and is already covered by the search.

`guess_magnetic_axis.cc` had no test, although the comment at the call site in `boundaries.cc` says it was split out of `Boundaries` for that purpose. Two are added, over `solovev` (`nZeta = 1`), `cma` (16), `cth_like_fixed_bdy` (36) and `cth_like_fixed_bdy_nzeta_37` (37): the guess lies inside the boundary in every plane, and the mirror relation holds with the axis in the symmetry plane at `zeta = 0`. Both fail on the old bound. No existing test does.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/boundaries/guess_magnetic_axis.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/boundaries/guess_magnetic_axis.cc
@@ -454,7 +454,11 @@ RecomputeAxisWorkspace RecomputeMagneticAxisToFixJacobianSign(
   // in order to always have a full toroidal module for the Fourier transform
   // below
   if (!s.lasym) {
-    for (int k = 1; k < s.nZeta / 2; ++k) {
+    // The search above covered 0 .. nZeta / 2. Every remaining plane is the
+    // mirror image of one of those: for an even number of toroidal grid points
+    // the half-period plane is its own mirror and is already done, for an odd
+    // number it is not, and its mirror is the first plane past the half period.
+    for (int k = 1; k <= (s.nZeta - 1) / 2; ++k) {
       const int k_reversed = (s.nZeta - k) % s.nZeta;
       w.new_r_axis[k_reversed] = w.new_r_axis[k];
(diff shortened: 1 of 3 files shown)


src/vmecpp/cpp/vmecpp/vmec/boundaries/guess_magnetic_axis.cc
```diff
@@ -454,7 +454,11 @@ RecomputeAxisWorkspace RecomputeMagneticAxisToFixJacobianSign(
   // in order to always have a full toroidal module for the Fourier transform
   // below
   if (!s.lasym) {
-    for (int k = 1; k < s.nZeta / 2; ++k) {
+    // The search above covered 0 .. nZeta / 2. Every remaining plane is the
+    // mirror image of one of those: for an even number of toroidal grid points
```
jurasic-pf: Comments should describe the state of code now not what was. 

I suggest to keep it brief at
    // mirror image along zeta, for even and odd nZeta

jurasic-pf (approved):
Good catch!

src/vmecpp/cpp/vmecpp/vmec/boundaries/guess_magnetic_axis.cc
CharlesCNorton: Done.

PR #730, opened 2026-08-31
Assemble the axis-guess boundary geometry over the full poloidal interval

CharlesCNorton:
Stacked on #729; the first commit is that one.

`RecomputeMagneticAxisToFixJacobianSign` evaluates the boundary against `FourierBasisFastPoloidal`, which tabulates the poloidal basis on `[0, pi]` only, and reconstructs the rest of the interval by mirroring. The mirror sits inside `if (!s.lasym)`, so without stellarator symmetry the extended interval keeps the zeros from `resize`. `min_element` over `r_lcfs[k]` then gives 0 rather than the inboard extent, and the minimum of `tau` over the poloidal index picks up the same zeros, so `min_tau_temp > min_tau` never holds and the search returns the grid centre in every plane. Separately the search runs to `nZeta / 2 + 1` with no `lasym` alternative, which leaves the planes past the half period unassigned.

The reconstruction now follows `symrzl`: the antisymmetric part is accumulated apart from the symmetric one, the extended interval is formed from both before the antisymmetric part is added into the reduced one, with R and dZ/dTheta even under `(theta, zeta) -> (-theta, -zeta)` and Z and dR/dTheta odd. Under stellarator symmetry the antisymmetric part is zero and this is the previous mirror. Every plane is searched when there is no symmetry to mirror across, and the half-period plane counts as its own mirror only when a grid point falls on it, which an odd `nZeta` has not.

Three tests. The assembled geometry is compared against a direct evaluation of the Fourier series over the full poloidal interval, on the bundled cases and on a boundary given antisymmetric content in both R and Z. A stellarator-symmetric boundary driven through the `lasym` path reaches the same axis as through the symmetric one.


jurasic-pf:
please resolve conflicts then we can merge

CharlesCNorton:
Conflicts resolved against current main.

The five ubuntu-22.04 failures are `cma`'s `jcuru` at 1e-6. Measured locally with the test's own metric, this branch and main both give 1.8161e-07; swapping only `guess_magnetic_axis.cc` changes nothing at 1, 4, 8 or 16 threads. The pre-merge commit and main are each green on all 20 of those jobs. Re-running.

jurasic-pf:
conflicts again, ready otherwhise

jurasic-pf:
I think this is no longer needed as part because the bug  was fixed as part of other PRs. If you disagree, please reopen.

PR #731, opened 2026-08-31
Answer the open questions in ideal_mhd_model

CharlesCNorton:
Open questions in `ideal_mhd_model.cc`, answered where there is an answer, plus one redundant fill removed.

| site | answer |
|---|---|
| `+1` on `bLambda`, `dLambda`, `cLambda` | the ghost point beyond the LCFS that `lamcal.f90` zeroes. Only the thread holding the LCFS reads it; every thread allocates it so the half-grid indexing stays one expression. |
| `curtor` in place of `cTor` | it cannot stand in. The enclosed current is constrained to the prescribed profile only for `ncurr == 1`; for `ncurr == 0` the code sets `chips = iotas * phips` and the enclosed current is an outcome of the solve, so NESTOR has to wait for the MHD routines. |
| `1/sqrt(2)` in the m = 1 constraint | `(1/sqrt(2)) [[1, 1], [1, -1]]` is orthogonal, so the change of variables leaves the residual norm that `fsqr` and `fsqz` measure unchanged and is its own inverse. With 1/2 it would halve the residuals on every application. |
| `pFactor = -4.0` | the coefficient of the second-radial-derivative terms of Section 5.14 of `docs/the_numerics_of_vmecpp.pdf`, which writes `FR = -D_RR d2R/drho2 + D_RZ d2Z/drho2 + ...` with `D_RR = Ztheta^2 d0` and `d0 = 2 R PB / tau`. `pTau` is `pFactor * r12 * totalPressure / tau * wInt`, whose middle factor is `d0 / 2` once the pressure part is dropped, so `pFactor = -4` makes `pTau = -2 d0 wInt`. The sign is the one in `FR`, leaving the assembled tridiagonal as `dF/dx`; the 4 pairs with the 1/4 the half-grid averages carry, as the `cx` line notes. |
| `/(4.0*4.0)` | undoes the scaling of `ard`, `azd` and of the `cos**2` in the de-aliasing. |
| `32 * deltaS` twice | the `(32 hs)**2` weight on `tcon`. |
| `tcon(ns) = 0.5*tcon(ns-1)` | the boundary surface receives MHD force contributions from the inside only, so it carries half the weight of an interior surface. |
| `dr += br` at `jF == 1, m == 1` | `extrapolateTowardsAxis` copies the m = 1 amplitudes from the first interior surface onto the axis, so `x(0)` is not an independent unknown and the sub-diagonal coupling to it folds into the diagonal of the innermost row. |

The one behaviour change: the `absl::c_fill_n` on `lambdaPreconditioner` is gone. It is zeroed once at construction and the assembly loop is the only writer, so the fill re-zeroed elements that were already zero. The loop skips the `jF = 0` row on the thread holding the axis and `(m, n) = (0, 0)` on every surface, and those bounds do not change over the life of the model, so no call can leave a stale value for a later one. `applyLambdaPreconditioner` reads both and multiplies the lambda force by them, so those permanent zeros are what zero the axis row and the `(0, 0)` lambda force. 205 Python tests pass with the fill removed.

Left as TODOs, because the answer would have been "the Fortran does this": `r1scale`, the hard-coded 50 and 1e-6, and the parabola in `ns`.


src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -430,7 +433,9 @@ void IdealMhdModel::evalFResInvar(const Eigen::Vector3d& localFResInvar) {
 #endif  // _OPENMP
   {
     // set new values
-    // TODO(jons): what is `r1scale`?
+    // Fortran residue.f90: r1 = 1 / (2 * r0scale)**2, handed to getfsq as
```
jurasic-pf: This doesn't explain the physical meaning or derivation of it. plesse revert

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -967,9 +979,10 @@ absl::StatusOr<bool> IdealMhdModel::update(
   // contribution in the first few iterations, preventing termination, to
   // ensure the free-boundary forces have "enough time" to propagate through
   // to the inner surfaces.
-  // TODO(jurasic) the hard-coded 50 and 1e-6 are only here for backwards
-  // compatibility, ideally vacuum-pressure should always part of the
-  // force-balance
+  // The 50 and the 1e-6 are Fortran residue.f90, which gates the same edge
+  // contribution on `delIter .lt. 50 .and. fsqrz .lt. 1.E-6`. They are here for
```
jurasic-pf: It didn't resolve the TODO, but removed it please revert

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -2192,7 +2205,12 @@ void IdealMhdModel::updateLambdaPreconditioner() {
   }
 
   // assemble lambda preconditioning matrix
-  // TODO(jons): maybe not needed, since direct assignments below?
+  //
+  // The loop below starts at jMin, so on the thread holding the axis the jF = 0
+  // row is never assigned. That row has to stay zero, which it already is from
+  // the setZero at construction, so dropping this fill leaves every test
+  // unchanged. It is kept so the invariant does not depend on nothing else ever
```
jurasic-pf: what?

If its provably not needed, then let's remove it and save a few instructions. Just clearly reason throigh why all fields are overwritten

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -2262,8 +2280,8 @@ void IdealMhdModel::computePreconditioningMatrix(
     temp_h.setZero(r_.nsMaxH - r_.nsMinH);
   }
 
-  // restored in v8.51
-  // TODO(jons): what is this?
+  // Fortran precondn.f90: pfactor = -4 * r0scale**2, restored in v8.51 after
```
jurasic-pf: The question isn't what this corresponds to in Fortran VMEC the question is, what is this physically or numerically? Where does it appear in the derivation of the discrete equations (see latex writeup the numerics of vmec)?

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
```diff
@@ -2421,14 +2439,16 @@ absl::Status IdealMhdModel::constraintForceMultiplier() {
   }
   // tcon
 
-  // TODO(jons): some parabola in ns,
-  // but why these specific values of the parameters ?
+  // Fortran bcovar.f90, verbatim. The quadratic ramp in ns is an empirical
```
jurasic-pf: Yes, we know that VMec2000 doesn't explain where these coefficients come from. That's why the todo was there.

Either leave the to-do, as is, or investigate in a separate issue how changing the coefficients affects the local force balance convergence, i guess in particular axis geometry? Things that might be worth looking at are how it does to eigenvalue spectrum of the hessian change locally based on these coefficients, does it dampen or accelerate the motion, build some intuition on why it is built the way it is.

I think the preconditioner is one of the less understood things in VMEC and documenting it with plots and explanations would be helpful for understanding the algorithm better

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
CharlesCNorton: Reverted.

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
CharlesCNorton: Reverted.

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
CharlesCNorton: Removed. `lambdaPreconditioner` is zeroed once in the constructor and this loop is the only thing that ever writes it, so the fill was re-zeroing elements that were already zero.

Two sets are skipped, and the bounds that skip them do not change over the life of the model, so no call can leave a stale value behind for a later one: the `jF = 0` row, on the thread holding the axis where `jMin` is 1, and `(m, n) = (0, 0)` on every surface. `applyLambdaPreconditioner` reads both and multiplies the lambda force by them, so those permanent zeros are what zero the axis row and the `(0, 0)` lambda force. The comment now says that rather than arguing for the fill.

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
CharlesCNorton: Section 5.14 of `docs/the_numerics_of_vmecpp.pdf` writes the R and Z forces as

```
FR = -D_RR d2R/drho2 + D_RZ d2Z/drho2 + ...
```

with `D_RR = Ztheta^2 d0` and `d0 = R |B|^2 / (mu0 tau) = 2 R PB / tau` (Eqns. 5.256 to 5.261). `pTau` here is `pFactor * r12 * totalPressure / tau * wInt`, and its middle factor is `d0 / 2` once the pressure part, which carries no second radial derivative, is dropped, so `pFactor = -4` makes `pTau = -2 d0 wInt`.

The sign is the one in `FR` above: it leaves the assembled tridiagonal as `dF/dx` rather than its negative. The 4 pairs with the 1/4 that the half-grid averages below carry, which the `cx` line already notes, where `0.25 * pFactor` comes out at exactly -1.

Comment rewritten on that rather than on the Fortran.

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
CharlesCNorton: Reverted to the TODO.

Sweep on the parabola. At fixed `ns`, moving `tcon0` is moving the two coefficients; `tcon0 = 1/parabola(ns)` cancels the growth, leaving the same total multiplier at every `ns`. `ftol` 1e-11, single grid, `niter` cap 60000.

Iterations, `li383_low_res` (mpol 4, ntor 3):

| ns | off | 1/parab | default |
|---|---|---|---|
| 15 | 635 | 330 | 340 |
| 25 | 689 | 376 | 428 |
| 51 | 740 | 471 | 583 |
| 99 | 814 | 564 | 817 |

Iterations, `cth_like_fixed_bdy` (mpol 5, ntor 4):

| ns | off | 1/parab | default |
|---|---|---|---|
| 15 | no convergence, fsq 7.0e-07 | 498 | 497 |
| 25 | no convergence, fsq 1.6e-07 | 553 | 556 |
| 51 | no convergence, fsq 9.1e-08 | 608 | 613 |
| 99 | no convergence, fsq 1.1e-06 | 633 | 635 |

The constraint force is needed: `cth_like` does not converge without it, `li383` takes twice the iterations.

The growth in `ns` is not. A fixed total multiplier matches or beats the default at every resolution, by 3, 12, 19 and 31 per cent on `li383` at `ns` 15, 25, 51, 99. On `cth_like` the value does not matter over the factor of three the parabola spans.

Spectral width does not respond. Among converged runs a factor of three in the multiplier moves mean `wout.specw` by under 1e-4, and on `li383` turning the constraint force off gives a lower mean specw (1.3864 against 1.4075 at `ns = 99`).

Two cases, so not enough to propose replacing the parabola. TODO left in place.

jurasic-pf (approved):

src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
jurasic-pf: THats an interesting analysis, thanks!

PR #733, opened 2026-08-31
Answer the open questions in the solver driver and radial profiles

CharlesCNorton:
Ten open questions across the solver driver, the radial profiles and the transforms, answered. Comments only; no code changes.

**`vmec.cc`.** The `kOff` assignment does not contradict the `kInitialized` one above it: that one applies to a hot restart, where the vacuum solution arrives with the initial state, while this branch is the `jacob_off` recovery pass, which drops back to a three-surface mesh the carried-over solution no longer matches. The state `profil3d()` sets up is used whenever there is nothing to interpolate from, which is the first multigrid step and a hot restart with `ns_old == 0`. The `INTERP` checkpoint sits between multigrid steps rather than inside the solver loop, so an iteration-count guard would have no meaning there. The spectral width comes from the backup because that is the last accepted state, the one whose residuals are printed on the same line; it is the decomposed state, so `<M>` is the spectral width of the scaled coefficients, as in Fortran VMEC, which takes its spectrum from `xc`.

The 0.01 restart threshold was measured rather than argued. Lowering it makes the restart fire earlier, which changes the iteration path and the converged state: against the reference outputs, 1e-3 moves 3 of the 45 targets outside tolerance and 1e-4 moves 10. Lowering it may still be right for cases that stall at ~2e-3, but it needs the references regenerated with it.

**`util.h`.** `4 pi x 10^-7` is the pre-2019 SI definition, which fixed `mu_0` exactly. The 2019 redefinition made it measured, CODATA-2018 `1.25663706212(19)e-6`, larger by 5.4e-10 relative. Fortran VMEC uses the exact value.

**`dft_toroidal.cc`.** R at the axis cannot depend on theta, so every `m > 0` mode vanishes there. The odd-parity arrays carry the coefficients divided by `sqrt(s)`, turning the physical `rho^m` into `rho^(m-1)`, so the `m = 1` amplitude stays finite at the axis while `m >= 3` still vanish, and the innermost half-grid point does read that finite value. Even-m modes have no `sqrt(s)` suppression, so only `m = 0`, the axis position, may be nonzero among them. Hence exactly `m = 0` and `m = 1`.

**`fourier_coefficients.cc`.** Nothing downstream reads the R and Z part of the boundary row: restricting `jMaxRZ` to `nsMax_`, the alternative left in the comment, leaves the whole suite unchanged for fixed and free boundary alike. The single limit is kept so R, Z and lambda share it.

**`radial_profiles.cc`.** The `EPSILON * curtor` comparison is Fortran `profil1d.f90` verbatim; it guards the division below against a vanishing edge value, and `currv` differs from `curtor` only by `MU_0`, so which sets the scale makes no practical difference. `radialBlending` is Fortran `bdamp`: a linear ramp from `2*pDamp` at the axis to 0 at the boundary, 0.1 to 0 with `pDamp = 0.05`, used by `bcovar.f90` as `lvv`, the weight blending the two estimates of covariant `B_v`, which is what `hybridLambdaForce` does here. That a time-step parameter sets a radial blending weight is odd, and the reference says so too.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/util/util.h
+++ b/src/vmecpp/cpp/vmecpp/common/util/util.h
@@ -180,11 +180,11 @@ int VmecStatusCode(const VmecStatus vmec_status);
 
 std::string VmecStatusAsString(const VmecStatus vmec_status);
 
-// Vacuum magnetic permeability in Vs/Am. This is the pre-2019 SI definition,
-// which fixed mu_0 at exactly 4 pi x 10^-7. The 2019 redefinition made it a
-// measured quantity, CODATA-2018 1.25663706212(19)e-6, larger by 5.4e-10
-// relative. Fortran VMEC uses the exact value, so this one is kept for 1:1
-// comparison against it.
+// vacuum magnetic permeability in Vs/Am (CODATA-2018)
+// TODO(jons): In the long term, we should use the CODATA value,
+// as it is the official value after re-definition of the SI system.
+// However, for now, use the old definition for 1:1 comparison against Fortran
+// VMEC.
 // static constexpr double MU_0 = 1.25663706212e-6;
 static constexpr double MU_0 = 4.0e-7 * M_PI;
 
--- a/src/vmecpp/cpp/vmecpp/common/util/util.h
+++ b/src/vmecpp/cpp/vmecpp/common/util/util.h
@@ -180,11 +180,11 @@ int VmecStatusCode(const VmecStatus vmec_status);
 
 std::string VmecStatusAsString(const VmecStatus vmec_status);
 
-// vacuum magnetic permeability in Vs/Am (CODATA-2018)
-// TODO(jons): In the long term, we should use the CODATA value,
-// as it is the official value after re-definition of the SI system.
-// However, for now, use the old definition for 1:1 comparison against Fortran
-// VMEC.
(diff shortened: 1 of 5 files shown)


src/vmecpp/cpp/vmecpp/common/util/util.h
jurasic-pf, line 1: The comment was correct and referring to the commented out value. The more interesting question is whether we are ready to possibly switch to the proper physical definition of Mu0.

CharlesCNorton: Comment restored.

Measured on main with `MU_0 = 1.25663706212e-6` and nothing else changed: the unit-test targets (34 of 41) and the Python suite (203 tests) pass unchanged. The 7 large-test targets fail, 113 cases, all with finite pressure or net current, since their educational_VMEC references were produced with `mu0 = 2.0e-7 * twopi` at tolerances of 1e-15 to 1e-11; the vacuum cases are unaffected. `tests/test_free_boundary_quasr.py` has its own `VACUUM_PERMEABILITY = 4.0e-7 * np.pi`.

abscab already uses the CODATA value (`abscab.hh:11`), so the mgrid tables built from coils carry it while the plasma side does not. The switch needs the references regenerated with the same constant.

PR #735, opened 2026-08-31
Resolve the stale and unanswered markers across the solver

CharlesCNorton:
Nine stale or unanswered markers, resolved. The only behaviour-affecting change is a de-duplication that is bit-identical by construction.

**`fpsi0`.** `1.5*bvcoH[0] - 0.5*bvcoH[1]` is the expression `bvcof[0]` is already assigned, a few hundred lines earlier in the same intermediate struct, so the extrapolation is now read rather than repeated.

**`makegrid_lib.cc`.** The marker asks for the stellarator-symmetric grid and the follow-up flip-mirroring, and for the factor of two that would come with them. `MakeCylindricalGrid` has honoured `assume_stellarator_symmetry` for some time, emitting `num_phi/2 + 1` planes, and both `ComputeMagneticFieldResponseTable` and `ComputeVectorPotentialCache` mirror the rest. Removed.

**`util.cc`.** The two bare markers sit on the guards of the two sweeps of the tridiagonal solve, which differ by one. Forward elimination reaches back a surface, so it starts one past `jMin`; back substitution reaches forward, so it runs down to `jMin` itself.

**`minimum_x` / `maximum_x`** are extrema in R, and are renamed accordingly.

**`input_extension`.** Deliberately not serialized: it names the Fortran INDATA file a run came from, and VMEC++ runs from JSON, which is why `WOutFileContents` sets it to the empty string.

**`bsqsav`.** The extrapolation of total pressure to the LCFS is `bsqsav(:,3)` in Fortran VMEC; named in the comment.

**`rcenin`, `aminr2in`, `bminz2in`, `bminz2`.** Fortran `eqfor.f90` forms all four, and their only consumers are the `add_real` calls in its debug dump. `bminz2in` and `bminz2` are the same expression there. Nothing depends on them, so nothing is missing here.

**`rzmin`, `zxmax`, `zxmin`.** `eqfor.f90` assigns them beside `rzmax` and then reads none of them, so leaving them out is right.

**"make const once ABSCAB supports this"** was accurate for two thirds of its twelve sites and stale for the rest. The eight that are handed to `abscab::magneticFieldCircularFilament` or `vectorPotentialCircularFilament`, which take `double*`, stay mutable and now say why. The four that only fill a `CircularFilament` protobuf are const.

Not touched, with reasons: the `input_extension` marker in `__init__.py` belongs to #697, which rewrites that file; the standalone-repo move needs a repository that does not exist yet; and the `debian:testing` re-enable is a CI matrix change gated on upstream 3.14 support.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/magnetic_field_provider/magnetic_field_provider_lib_test.cc
+++ b/src/vmecpp/cpp/vmecpp/common/magnetic_field_provider/magnetic_field_provider_lib_test.cc
@@ -370,10 +370,8 @@ TEST(TestMagneticField, CheckCircularFilament) {
   static constexpr int kNumEvaluationLocations = 30;
 
   // centered at the origin, normal along z axis
-  std::vector<double> center = {
-      0.0, 0.0, 0.0};  // TODO(jons): make const once ABSCAB supports this
-  std::vector<double> normal = {
-      0.0, 0.0, 1.0};  // TODO(jons): make const once ABSCAB supports this
+  const std::vector<double> center = {0.0, 0.0, 0.0};
+  const std::vector<double> normal = {0.0, 0.0, 1.0};
 
   CircularFilament circular_filament;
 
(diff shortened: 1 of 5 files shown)


jurasic-pf (approved):
LGTM

PR #736, opened 2026-08-31
Extrapolate the boundary chipF like iotaF

CharlesCNorton:
`updateFullGridProfiles` extrapolated two half-grid arrays onto the same boundary point with different rules: `iotaF` with `1.5*f_last - 0.5*f_prev`, the linear extrapolation onto the boundary, and `chipF` with `2*f_last - f_prev`, which lands a full step past the last half-grid point. Both came verbatim from `add_fluxes.f90`.

`chipF` now uses the same `1.5/-0.5` rule. The identity `iota = chi'/phi'` held to 2e-16 at every interior full-grid point and failed only at the boundary, by exactly `0.5*(iotaH_last - iotaH_prev)`: 2.8% on `cth_like_free_bdy`, 1.4e-2 on `cth_like_fixed_bdy`. It now holds at the boundary as well.

This deviates from Fortran VMEC at one point of two outputs: the boundary entry of `wout.chipf` and the threed1 first table. The force balance never reads `chipF`, so the converged equilibrium is unchanged. The four comparisons against Fortran references skip that entry, with a comment at each.


jurasic-pf:
Ok, that sounds like one of these extrapolations is wrong, no? If its a bug in the original, I'm happy to deviate from it.
This is something I'd also like @jons-pf opinion on, but if both are half grid, then I don't see why we would use different extrapolation rules.

CharlesCNorton:
Changed: chipF now uses the same 1.5/-0.5 rule and the identity holds at the boundary too. The four reference comparisons skip the boundary entry. If @jons-pf knows the Fortran rule to be intentional, happy to revert.

jurasic-pf:
can you resolve merge conflicts? do you know what PARVMEC does here? analytically this really looks like a bug to me

CharlesCNorton:
Conflicts resolved.

PARVMEC uses the same rule at both ends. From `Sources/General/add_fluxes.f90`:

```fortran
chipf(1)     = c1p5*chips(2) - p5*chips(3)                       !SPH ADDED THIS 4-8-16
chipf(2:ns1) = (chips(2:ns1) + chips(3:ns1+1))/2
chipf(ns)    = c1p5*chips(ns)- p5*chips(ns1)                     !SPH FIXED THIS 4-8-16
```

and `iotaf` a few lines below is `c1p5*iotas(ns) - p5*iotas(ns-1)`, so PARVMEC extrapolates the two half-grid arrays identically, which is the argument you made above. The boundary line carries "SPH FIXED THIS", and it is the same 2016 change as the axis line: #773 landed the axis half, this is the other half. The 8.52 lineage the reference files come from has neither.

`chipf` now deviates from those references at both ends, so the four comparisons skip `[1:-1]` rather than just the boundary.

PR #737, opened 2026-08-31
Record the measured floor for the hot restart comparison tolerance

CharlesCNorton:
`HotRestartIntegration.FreeBoundary` compares a hot-restarted run against a from-scratch run of the same displaced configuration at a tolerance of 0.1, with a marker asking whether that is realistic. Measured, and it is: comments only.

Tightening it from above, 8e-2 passes and 5e-2 does not, so 0.1 sits within a factor of two of the floor rather than being an arbitrary loose bound. Everything else in the file compares at 1e-15, and the other `CompareWOut` callers use 1e-7, which is what makes 0.1 look out of place.

What binds is `DCurr`, one of the Mercier terms, then `jdotb`, then `jcuru`. Splitting the current densities off behind `current_density_tolerance`, which `CompareWOut` provides for exactly this, buys nothing here because `DCurr` fails before them. All three are derivative diagnostics of the converged state and amplify the difference between two convergence paths.

The reason the floor is where it is: the two runs do not land on the same point to `ftol`. Even `rmax_surf`, a plain geometric quantity, only agrees to about 1e-6. The test is asking two iteration paths into the same shallow minimum to agree, not two evaluations of one state.

Left at 0.1. A factor of 1.25 is not worth the churn and would be fragile across platforms.

### DIFF
--- a/src/vmecpp/cpp/vmecpp_large_cpp_tests/vmec_hot_restart_test/vmec_hot_restart_test.cc
+++ b/src/vmecpp/cpp/vmecpp_large_cpp_tests/vmec_hot_restart_test/vmec_hot_restart_test.cc
@@ -757,10 +757,10 @@ TEST(HotRestartIntegration, FreeBoundary) {
 
   // COMPARE RUN FROM SCRATCH AND HOT-RESTARTED RUN
   //
-  // The bound is set by jdotb, which differs by 2.8e-2 between the two runs,
-  // and jcuru at 1.6e-2. Everything else stays at or below 1.8e-3 (DMerc),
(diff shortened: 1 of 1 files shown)


CharlesCNorton:
fixing conflicts

jurasic-pf:
Thanks for looking into this, can we compress to comment a bit? So, it's easily human readable? very long explanation for such a minor cause

CharlesCNorton:
Cut to four lines: the tolerance, the three quantities that set it, and the reason they are loose. The measurement detail is in the PR description.

jurasic-pf (approved):

PR #739, opened 2026-08-31
Cover the mgrid consistency checks and say where they live

CharlesCNorton:
Two markers in `VmecINDATA::IsConsistent` ask for the mgrid consistency checks: that `nzeta` agrees with the vacuum field, and that the number of coil currents matches the number of response tables. Neither can be done there, because `IsConsistent` never sees the mgrid, which is why they were left open.

Both checks already exist where the mgrid is available. `Vmec::run` compares `MGridProvider::numPhi` against `nzeta`, and `LoadFile` and `LoadFields` each compare the coil-current count against the number of response tables they found. None of the three was covered by a test.

Three added:

- `LoadFile` rejects one current too few and one too many, and still loads the matching count;
- `LoadFields` does the same against a small synthetic two-circuit response table;
- `vmecpp::run` rejects an in-memory response table built at half the input's toroidal resolution.

Disabling any one of the three checks fails the corresponding test.

The markers are replaced by a note saying where the checks live and why they are not in `IsConsistent`.


CharlesCNorton:
fixing conflicts

jurasic-pf:
superceeded

PR #740, opened 2026-08-31
Reject a multigrid step count below one

CharlesCNorton:
`Vmec::run` takes `maximum_multi_grid_step`. Below 1 the multigrid loop runs no step and the function falls through to `return false` with an OK status, which reads as "checkpoint not reached" rather than "nothing was solved". It is now rejected, with a test that takes the previous code down the silent path.

The parameter is C++-only and nothing in the tree passes a non-default value, so this is insurance against exposing it rather than a live bug.

The `TODO(jons)` on the mgrid coil group names is rewritten to say what is known and what is missing. MAKEGRID takes the name from the coils file: the terminating line of every coil carries a group index and a group name, coils sharing an index share the name, so one name per group is well defined, and `ImportMagneticConfigurationFromCoilsFile` already keeps it on each filament. `coils.cth_like` is `HF-OVF` on group 1 and `TVF` on group 2. It is not reachable at the point it would be written: `WriteMakegridNetCDFFile` receives the response table and the currents, not the `MagneticConfiguration` the names live on, so using them means threading the configuration through or carrying them on `MagneticFieldResponseTable`. Nothing reads `coil_group` back and the checked-in mgrid files were written by this code, so they already carry the `circuit_%d` placeholders. The marker stays for that plumbing.

Two markers this PR previously touched are dropped from it. The boundary handedness one is resolved on main by #789, which replaced the `rTest * zTest` product with the signed area of the m = 1 ellipse, the invariant the marker proposed. The `Sizes` reference values cannot be parameterized until educational_VMEC dumps them.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib.cc
+++ b/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib.cc
@@ -964,10 +964,10 @@ absl::Status WriteMakegridNetCDFFile(
 
   for (int circuit_index = 0; circuit_index < n_serial_circuits;
        ++circuit_index) {
-    // TODO(jons): Figure out how the name of the coil groups is determined in
-    // MAKEGRID. The challenge is that many coils (with individual names) can
-    // belong to one coil group, but only a single name for a coil group is
-    // written to the mgrid file.
+    // TODO(jons): MAKEGRID names a coil group by the group name on the
+    // terminating line of its coils in the coils file. That name reaches here
+    // only once the MagneticConfiguration is threaded through or the names are
+    // carried on MagneticFieldResponseTable.
     std::string coil_group_name = absl::StrFormat("circuit_%d", circuit_index);
 
     // NOTE: 30 has to be consistent with the value of kStringSize.
--- a/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
@@ -228,6 +228,12 @@ absl::StatusOr<bool> Vmec::run(const VmecCheckpoint& checkpoint,
                                const int iterations_before_checkpointing,
                                const int maximum_multi_grid_step,
                                std::optional<HotRestartState> initial_state) {
+  if (maximum_multi_grid_step < 1) {
+    return absl::InvalidArgumentError(
+        absl::StrFormat("maximum_multi_grid_step must be at least 1, but is %d",
+                        maximum_multi_grid_step));
+  }
+
   if (indata_.lfreeb) {
     if (!mgrid_.IsLoaded()) {
       // Fallback: load mgrid from file if constructed directly via the public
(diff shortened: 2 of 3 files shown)


src/vmecpp/cpp/vmecpp/vmec/boundaries/boundaries.cc
```diff
@@ -235,11 +235,14 @@ bool Boundaries::checkSignOfJacobian() {
...
-  // - compute signed polygon area
-  // --> handedness of polygon is given by sign of polygon area
+  // A more robust form would evaluate the boundary in one poloidal plane at
+  // enough equally spaced theta to satisfy Nyquist and take the sign of the
+  // signed polygon area, which cannot go degenerate the way summing the m=1
+  // coefficients can when they cancel across n.
+  //
+  // Nothing shipped is near that. Over every bundled input the product below
+  // stays well away from zero; the closest is near_axis_iota_nfp4 at 2.2e-3,
+  // with each factor around 5e-2, and most are of order 1.
```
jurasic-pf: This isn't an implementation relevant comment imo

src/vmecpp/cpp/vmecpp_large_cpp_tests/common/sizes/sizes_test.cc
```diff
@@ -22,9 +22,9 @@ using testing::IsCloseRelAbs;
 // For now, these are written manually here, since at the time of implementing
 // these tests, the corresponding outputs were not written yet to debugging
 // output files.
-// TODO(jons): write a debugging output from educational_VMEC, which contains
-// all these reference values, and test against those with a parameterized test
-// for all cases
+// Turning these into one parameterized test needs educational_VMEC to dump the
+// Sizes values, which it does not, so there is nothing to parameterize over
+// until that dump exists.
```
jurasic-pf: Neither is this. Should be a PR comment if anything

jurasic-pf (approved):

CharlesCNorton:
checking out ci failure

CharlesCNorton:
The asan failure is not from this change: Threed1ShafranovIntegralsTest/cma fails on s2 at 5.114e-11 against the 5e-11 tolerance (ubsan was only the fail-fast cancellation). That tolerance sits 1.23x above its measured deviation and moves with the toolchain, so a rerun will not clear it. #764 raises it to 5e-10 with the measurement behind it.

src/vmecpp/cpp/vmecpp/vmec/boundaries/boundaries.cc
jurasic-pf, line 1: why remove this??

jurasic-pf (approved):
Thanks, good idea to cover the edge case

CharlesCNorton:
That TODO is not removed by this PR any more: #789 landed on main in the meantime and replaced the rTest*zTest product with the signed area of the m=1 ellipse, which is the invariant the marker asked for, so the boundaries.cc hunk is gone from the diff. Rebased onto current main, and the makegrid marker is now four lines saying where MAKEGRID gets the name and what has to be threaded through to use it.

CharlesCNorton:

CharlesCNorton:

src/vmecpp/cpp/vmecpp/vmec/boundaries/boundaries.cc
CharlesCNorton: The commit after this comment took that hunk back out, so the PR no longer touches boundaries.cc at all. The marker is gone from main independently: #789 replaced the rTest*zTest product with the determinant that stays invariant under the lasym poloidal rotation.

PR #741, opened 2026-08-31
Check the spline knots round-trip and that the coil displacement takes effect

CharlesCNorton:
Two test markers, both closed by adding the check rather than by explaining why it is absent.

**Spline profile knots.** Four markers in the wout comparison ask for `am_aux_*`, `ai_aux_*` and `ac_aux_*` to be checked. They cannot be checked there: that test compares against Fortran references, and none of the cases it runs uses a spline profile, so there is no reference carrying knots.

`SplineProfileEquilibrium` does run one, so the check goes there instead, against the input rather than a reference. The knots are an input echo, and nothing else in the output records which spline the run was given, so `wout.am_aux_s` and `am_aux_f` have to come back equal to what went in.

**Whether the coil displacement did anything.** The free-boundary hot-restart test compared the hot-restarted and from-scratch runs against each other and nothing else, so a run that silently ignored the displaced field would have agreed with another run that did the same and passed. The marker notes that `CompareWOut` cannot be used the other way round, since it aborts on a mismatch rather than reporting one.

A small local helper reports the largest relative difference over the wout scalars and `rmnc` instead, and the test now requires the displaced equilibrium to differ from the original. Setting the displacement to zero drops the measured difference to 1.8e-9 and fails the new assertion, so it has teeth.


jurasic-pf (approved):

jurasic-pf:
I hope I resolved the merge conflicts correctly, might need to check CI

PR #742, opened 2026-08-31
Recombine bsubuv and bsubvu onto the full poloidal mesh

CharlesCNorton:
`bsubuv` and `bsubvu` were the only quantities in `ComputeCovariantBDerivatives` not split into symmetric and antisymmetric parts, and so the only ones that never got the recombination onto the full poloidal interval that the split exists for. They are allocated on the full grid, written at reduced-layout indices `jH * nZnT_reduced + k * nThetaReduced + l`, and read in `ComputeJxBOutputFileContents` at full-layout indices `jH * nZnT + kl`. Those coincide only when `nThetaEff == nThetaReduced`, which is to say only without `lasym`.

For an asymmetric run `jsups3`, their only consumer, was consequently wrong. On `up_down_asym`, with `nZnT = 16` and `nZnT_reduced = 9`, 112 of its 256 entries came back exactly zero, which is `1 - 9/16` of the grid, and the rest were read from the wrong poloidal position.

They now accumulate into `_s` and `_a` scratch like `bsubu` and `bsubv`, and are recombined the same way, with the opposite sign on the symmetric part: these are derivatives with respect to zeta and theta, and differentiating with respect to a variable the reflection negates turns an even function odd, so the extended interval takes `-symmetric + antisymmetric` where `bsubu` takes `+symmetric - antisymmetric`. On `up_down_asym` the exact zeros drop from 112 to 16, and the 16 that remain are one `jH` row, the last half-grid surface the consumer loop does not reach.

Symmetric runs take the same branch they did before and are unchanged; the whole suite passes.

The new test pins it without needing an asymmetric reference: a stellarator-symmetric boundary carries no antisymmetric content, so running it through the `lasym` path has to reproduce the symmetric run at the poloidal points the two grids share. Against the fix the worst difference is under 1e-3 of the array scale; without it, 1.24 times the array scale.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -2159,26 +2159,6 @@ vmecpp::CovariantBDerivatives vmecpp::LowPassFilterCovariantB(
   covariant_b_derivatives.bsubsv =
       RowMatrixXd::Zero(m_vmec_internal_results.num_full, s.nZnT);
 
-  // These two are the exception, and it costs them: unlike bsubu, bsubv,
-  // bsubsu and bsubsv, they are not split into symmetric and antisymmetric
-  // parts, and so they never get the recombination onto the full poloidal
-  // interval that the split exists for. They are allocated here on the full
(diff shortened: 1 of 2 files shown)


CharlesCNorton:
Superseded by #771, which landed the same bsubuv/bsubvu recombination on main.

PR #743, opened 2026-08-31
Say what is actually missing behind the unimplemented markers

CharlesCNorton:
Six markers for unimplemented features, each replaced by what is actually missing and why. Comments only.

**`bsqsav(:,1)` and `(:,2)`, "implement this !!!"** Nothing to carry over. Fortran `funct3d.f90` assigns both at that point and never reads either; the only column read anywhere is `bsqsav(:,3)`, for the `delbsq` diagnostic in `printout.f90`, and that is the extrapolation already computed a few lines above and reported through `delBSq`.

**`lgiveup`/`fgiveup`.** Absent from `educational_VMEC` altogether, which is what these outputs are validated against. Adding the heuristic would introduce a termination rule with nothing to match it to.

**`freeb_data`.** The free-boundary cases it was waiting on now exist. What `freeb_data.f90` writes is the edge field, `brv`, `bphiv`, `bzv`, together with the vacuum potential, to a legacy unit-number file. The potential half of that has since landed as `potvac`, `xmpot` and `xnpot`; the edge field components are what is still missing.

**`curlabel`.** Blocked on `MGridProvider` not reading `coil_group` back out of the mgrid file. That is the reading half of the same gap the writing side has in `makegrid_lib`, where the group names sit in the coils file and reach neither end.

**Excluding the axis from the force norm.** PARVMEC does; adopting it changes the force normalization, hence the residuals, hence where every run stops. It needs the reference outputs regenerated rather than being a local edit.

**The version string.** Bumping it means matching PARVMEC, not editing the string, so the marker stays until that is true.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
@@ -883,15 +883,13 @@ absl::StatusOr<bool> IdealMhdModel::update(
           delBSq[kl] = fabs(outsideEdgePressure - insideTotalPressure[kl]);
         }
 
-        if (m_vacuum_pressure_state_ == VacuumPressureState::kInitialized) {
-          // TODO(jons): implement this !!!
-
-          // initial magnetic field at boundary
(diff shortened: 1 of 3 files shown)


CharlesCNorton:
Closing: this was a set of notes rather than a change. The two items in it that are actually implementable are being done instead.

PR #744, opened 2026-08-31
Read the mgrid coil group names and report them as wout curlabel

CharlesCNorton:
Stacked on #739.

`wout.curlabel` was declared, serialized and never filled, because the coil group names sat in the mgrid file and nothing read them back.

`MGridProvider::LoadFile` now reads the `coil_group` variable, a `[nextcur][30]` character array, strips the padding MAKEGRID writes, and stores the names. They are threaded to `ComputeWOutFileContents` the way `mgrid_mode` already is, and become `wout.curlabel`. An in-memory response table carries no names, so `LoadFields` leaves the list empty.

The test loads `mgrid_cth_like.nc` and requires one name per external current, none empty and none still padded.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
@@ -1487,14 +1487,6 @@ absl::Status IsConsistent(const VmecINDATA& vmec_indata,
   // lfreeb
   // nothing to check here: lfreeb can be true or false and both are valid...
   if (vmec_indata.lfreeb) {
-    // mgrid_file, extcur
-    //
-    // Both consistency checks need the mgrid, which this function does not
-    // have, so they live where it is available instead. Vmec::run compares
-    // MGridProvider::numPhi against nzeta, and MGridProvider::LoadFile and
-    // LoadFields each compare the number of coil currents against the number
-    // of response tables they found.
-
     // nvacskip
     if (vmec_indata.nvacskip < 1) {
       return absl::InvalidArgumentError(absl::StrFormat(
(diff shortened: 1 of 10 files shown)


src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
jurasic-pf, line 1: please remove the comment, since it's now just describing what is not here

src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.cc
```diff
@@ -29,6 +29,10 @@ using netcdf_io::NetcdfReadString;
 
 namespace {
 
+// Fixed width of each name in the mgrid coil_group variable, matching
```
jurasic-pf: Can we generalize the logic to arbitrary length string while we're at it? As long as we stay compatible with the legacy padded values, I'd prefer to widen the contract here

jurasic-pf (commented):
Good idea, see minor commemts

src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
CharlesCNorton: Removed.

src/vmecpp/cpp/vmecpp/free_boundary/mgrid_provider/mgrid_provider.cc
CharlesCNorton: Done. The width now comes from the coil_group variable's second dimension, so any declared width reads correctly; the padding strip is unchanged, so MAKEGRID's 30 still works.

PR #745, opened 2026-08-31
Report the free-boundary edge quantities as a threed1 section

CharlesCNorton:
`freeb_data` in Fortran VMEC reports the boundary geometry, the plasma-side and vacuum-side pressures both as first established and as converged, and the cylindrical field on either side of the boundary. VMEC++ had none of it, and the two markers behind it, the `freeb_data` one and the `implement this !!!` one on `bsqsav(:,1)` and `(:,2)`, turn out to be the same piece of work: those two arrays exist only to be reported here.

Added:

- `HandoverStorage` keeps the three pressure snapshots: the plasma-side and vacuum-side values at the boundary as they stood when the vacuum solution was first established, and the extrapolated edge pressure, which are `bsqsav(:,1)`, `(:,2)` and `(:,3)`.
- `Threed1FreeBoundary` holds the thirteen edge arrays and is written to the output file beside the other `threed1` sections. It stays allocated and zero for a fixed-boundary run, as `potvac` does, and loading tolerates files written before the group existed.

Everything else it needs was already there: the vacuum field and pressure in `HandoverStorage`, the boundary geometry in the converged state, and the cylindrical field components, which the plasma side reaches by the half-to-full extrapolation `1.5 * last - 0.5 * previous`.

All thirteen arrays match the `educational_VMEC` `freeb_data` dump for `cth_like_free_bdy` to 1e-10 or better, most of them to 1e-13.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/handover_storage/handover_storage.h
+++ b/src/vmecpp/cpp/vmecpp/vmec/handover_storage/handover_storage.h
@@ -168,6 +168,19 @@ class HandoverStorage {
   // [nZnT] vacuum magnetic pressure |B_vac^2|/2 at the plasma boundary
   Eigen::VectorXd vacuum_magnetic_pressure;
 
+  // [nZnT] total pressure of the plasma at the boundary at the moment the
+  // vacuum solution was first established, bsqsav(:,1) in Fortran VMEC.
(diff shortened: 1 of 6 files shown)


CharlesCNorton:
fixing clang

jurasic-pf (approved):

PR #746, opened 2026-08-31
Share the current-weighted accumulation between LoadFile and LoadFields

CharlesCNorton:
`LoadFile` and `LoadFields` each sized the accumulation arrays and each ran their own current-weighted sum over the grid, differing only in where a circuit's contribution came from: nested vectors read from netCDF in one, response table rows in the other.

Both now call `ResetAccumulatedField` and `AccumulateCircuit`, the latter taking a callable that returns one circuit's three components at a grid point. The weighting exists once.

Summation order is unchanged, so the accumulated field is bit-identical; the whole suite passes.


jurasic-pf (approved):
nice, DRY

PR #748, opened 2026-08-31
Carry the non-stellarator-symmetric half of the state through a hot restart

CharlesCNorton:
`FourierGeometry::InitFromState` took only `rmnc`, `zmns` and `lmns_full`. A hot restart of a `lasym` equilibrium therefore began from the stellarator-symmetric projection of the saved state. `WOutFileContents` carries `rmns`, `zmnc` and `lmnc_full`, and they were never read. The boundary copy pulled `rbcc`, `zbsc`, `rbss` and `zbcs` but not `rbsc`, `zbcc`, `rbcs` and `zbss`, so the `m1Constraint(0.5, ...)` that follows mixed coefficients that were zero. `CheckInitialState` only compares the `lasym` flag, so nothing reported the loss.

Measured on `up_down_asym`, whose converged state has `max|rmns| = max|zmnc| = 0.268`:

| | before | after |
|---|---|---|
| `max\|rmnsc\|`, `max\|zmncc\|`, `max\|lmncc\|` after `InitFromState` | 0 | matches the state |
| iterations to restart from the converged state | 657 | 2 |

A cold start of the same case takes 1193 iterations.

The signature now takes the three antisymmetric arrays alongside the symmetric ones. R's antisymmetric half is stored in the combined sine basis and Z's in the combined cosine basis, mirrored from the symmetric half, so `sin_to_sc_cs` and `cos_to_cc_ss` swap relative to the existing branch. A `lasym` run given empty arrays now fails a `CHECK`.

`HotRestart.InitializeFromExistingState` runs for both symmetry classes and is instantiated for `up_down_asym`, which covers the `rsc`/`zcc` half of the m=1 constraint and the antisymmetric lambda path. Its lambda reference is scaled by `lamscale / phip`, and the `lthreed`-only comparisons are gated so an `ntor = 0` case can run. This closes the `TODO(jons)` at `fourier_geometry_test.cc:238`.


src/vmecpp/cpp/vmecpp_large_cpp_tests/vmec/fourier_geometry/BUILD.bazel
```diff
@@ -26,5 +27,5 @@ cc_test(
         "//vmecpp/common/flow_control:flow_control",
         "//vmecpp/vmec/vmec:vmec",
     ],
-    size = "small",
+    size = "enormous",
```
jurasic-pf: why did the test duration increase so much? can we at least keep it medium

jurasic-pf (commented):
Good change, pending comment.

src/vmecpp/cpp/vmecpp_large_cpp_tests/vmec/fourier_geometry/BUILD.bazel
CharlesCNorton: Measured on the branch: the whole target takes 0.1 s in opt and 0.5 s under asan; the asymmetric round trip that solves up_down_asym in-test is 71 ms and 290 ms of that. Back to small.

PR #749, opened 2026-08-31
Serialize the whole non-stellarator-symmetric axis and boundary arrays

CharlesCNorton:
`raxis_s`, `zaxis_c`, `rbs` and `zbc` are `std::optional`. `WriteTo` and `ToJson` reached them as `raxis_s->value()`, which dereferences the optional and then calls `Eigen::DenseBase::value()`. That returns `coeff(0, 0)`, so each array collapsed to its first element.

`OutputQuantities::Save` writes the INDATA into every output `.h5`, so a `lasym` run stored the antisymmetric axis and boundary as four scalars. `LoadInto` reads them back as arrays: the dataspace is scalar, `getSimpleExtentDims` writes nothing, and the shape comes from the uninitialized `dims`. On `up_down_asym`:

```
rbs in:   0  0.6  0.12  0            0
rbs out:  0  0     0    0  2.37152e-322
```

`ToJson` emits `"raxis_s":0.0`, and `FromJson` rejects that with `JSON element 'raxis_s' is not an array`, so a `lasym` INDATA does not survive its own JSON round trip either. `VmecInput.to_json` in the Python layer uses the same path.

Both sites now take `*raxis_s`. The HDF5 and JSON round-trip tests run for both symmetry classes and are instantiated for `up_down_asym`; all four assert on the axis and boundary arrays.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/BUILD.bazel
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/BUILD.bazel
@@ -47,6 +47,7 @@ cc_test(
         "//vmecpp/test_data:cth_like_fixed_bdy_nzeta_37",
         "//vmecpp/test_data:cma",
         "//vmecpp/test_data:cth_like_free_bdy",
+        "//vmecpp/test_data:up_down_asym",
     ],
     deps = [
         ":vmec_indata",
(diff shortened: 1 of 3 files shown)


jurasic-pf (approved):

jurasic-pf:
good catch!

PR #751, opened 2026-08-31
Evaluate magnetic configurations directly in the (3, n) Eigen layout

CharlesCNorton:
`ComputeMagneticFieldResponseTable` and `ComputeVectorPotentialCache` each carried two `TODO(jurasic) Remove after Eigen refactor` blocks. The `(3, n)` cylindrical grid was unpacked into an `n`-element `std::vector<std::vector<double>>`, a second one was allocated for the result, and the result was folded back with `ToEigenMatrix(...).transpose()`. Both blocks sat inside the per-circuit loop, so the grid was unpacked again for every circuit.

The provider forced that layout, so it gains `(3, n)` entry points for `MagneticConfiguration`:

```cpp
absl::Status MagneticField(const MagneticConfiguration &magnetic_configuration,
                           const RowMatrix3Xd &evaluation_positions,
                           RowMatrix3Xd &m_magnetic_field,
                           bool check_current_carrier = true);
```

and the same for `VectorPotential`. `makegrid_lib.cc` calls those, and all four blocks go.

ABSCAB reads point sets as flat array-of-structs buffers. The old path rebuilt that buffer once per current carrier, discarded it, and rebuilt it for the next one. The traversal now converts the evaluation positions once per configuration and hands the same buffer to every carrier. The per-filament bodies become contribution cores over spans, which the public per-filament functions wrap. Each carrier still writes into a scratch buffer that is zeroed first and added on afterwards, so the summation order and the rounding are unchanged.

Two new tests require the vector-of-vectors API and the `(3, n)` API to agree under `EXPECT_EQ` for both B and A. The `makegrid` binary built from this branch and from `main`, run on `makegrid_parameters_cth_like.json` with `coils.cth_like`, produces byte-identical `mgrid_cth_like.nc`.


jurasic-pf:
Did you measure whether this actually has an performance impact?
I'm a little confused by the numbers reported in the benchmark CI, they're way better for all cases, even ones that should be unaffected. Maybe just a randomly fast machine?

jurasic-pf:
Ok, and rerunning CI leaves it within noise of the usual times. Do you have a local run showing speedup/regression with smaller error bars?

CharlesCNorton:
Measured locally on a quiet machine (i9-12900H, OMP_NUM_THREADS=8): ten interleaved runs per binary of the full makegrid invocation, this branch against its merge-base 42862205.

| case | merge-base | this branch | ratio |
|---|---|---|---|
| `makegrid_parameters_cth_like` | 1.049 ± 0.017 s | 0.954 ± 0.021 s | 1.10x |
| `makegrid_parameters_solovev` | 0.212 ± 0.012 s | 0.129 ± 0.006 s | 1.65x |

The times cover the whole run with the netCDF write included, so the evaluation itself speeds up by more than the ratios show. solovev gains most because its single-circuit configuration spent the largest share of its time re-unpacking the grid per circuit.

CharlesCNorton:
@jurasic-pf

jurasic-pf (approved):

PR #752, opened 2026-08-31
Implement the FourierFilament current carrier

CharlesCNorton:
`CurrentCarrier::kFourierFilament` was an enum value with no type behind it. The union held only the other three carriers, so the value was unreachable and the two `CheckFourierFilament` tests were empty.

`FourierFilament` carries a `CurveRZFourier` and a `num_sampling_points`. The sampling count is required rather than defaulted: how finely the curve has to be discretized depends on how close the evaluation locations come to it, not on its mode content.

`ToPolygonFilament` samples the curve uniformly in the curve parameter and returns the closed polygon through those points, the last vertex repeating the first. `MagneticField` and `VectorPotential` delegate to the polygon overloads, and both `MagneticConfiguration` traversals gained the case. `IsFourierFilamentFullyPopulated`, `CurrentCarrierIdentifier` and `PrintFourierFilament` are filled in alongside.

Tests: a filament carrying only m = 0 coefficients is a circular loop, checked against the closed form under refinement; a shaped filament checked against the polygon it samples into; the sampled vertices checked against R(phi) and Z(phi) evaluated from the coefficients; and validation checked for missing geometry, missing sampling count, and a count below three.

`DetermineIfTooCloseToCurrentCarrierForComparison` in `makegrid_lib_test.cc` is left alone, since #722 is changing that file.


jurasic-pf (changes requested):
Sorry for the confusion, but I don't think we want to expand the magnetics library of vmecpp further. That was the intention when we wrote it initially, but as things have developed, we decided that keeping more of the evaluation logic python side is better for most users.
Therefore we can remove that enum value and close the PR.

PR #753, opened 2026-08-31
Cover the aphi radial coordinate reparameterization

CharlesCNorton:
`aphi` reparameterizes the radial coordinate onto enclosed toroidal flux. Every test case leaves it at its default of `{1}`, where `torflux` is the identity, `torfluxDeriv` is one, and `polfluxDeriv` reduces to `iota(x)`. None of that arithmetic was covered.

Five tests in `radial_profiles_test.cc` cover it against closed forms:

- `torflux` and `torfluxDeriv` against the polynomial and its derivative, for four `aphi`.
- The default `aphi` leaving `torflux` as the identity.
- `maxToroidalFlux * torflux(1)` equal to `signOfJacobian * phiedge / 2 pi`, so the normalization matches the profile it scales.
- `polfluxDeriv` evaluating iota at the enclosed toroidal flux rather than at the radial position, and clamping that argument at 1. With `aphi = {0, 1}` and `iota(t) = t` the two arguments give different answers away from the edge.
- `polflux` equal to `iota * torflux` for a constant iota profile, where the trapezoid is exact.

The tests also record a difference from the reference. `torflux` evaluates the integral of `torfluxDeriv` in closed form; `magnetic_fluxes.f90:53-72` uses a 100-interval trapezoid. Those agree only while `torfluxDeriv` is linear, so for `aphi` with three or more coefficients the results differ, by 1.7e-05 relative for `aphi = {1, 0, 0.5}`. Replacing the closed form with the Fortran trapezoid fails the two `torflux` tests and leaves the other twenty passing, `MatchesFortranReference` included.

The closed form is the one to keep: `maxToroidalFlux` divides by `torflux(1)` to set the edge flux, and `phipf` is built from `torfluxDeriv`, so the trapezoid would normalize the profile by a value the profile does not integrate to. Both functions now say so.

The marker at `radial_profiles.cc:344` claimed the poloidal flux block is only needed for `lRFP == true`. `lRFP` is not in educational_VMEC either; `profil1d.f90:32-34` computes `polflux_edge` the same way and likewise does not consume it, since `chips` and `chipf` come from `torflux_edge * polflux_deriv`. The comment now says that.


CharlesCNorton:
The opt job fails on `output_quantities_test.cc:795`, which compares `jdotb` for `cma` at 1e-6. On an unmodified base checkout that deviation is thread-count dependent: 4.5e-07 at 1 thread, 6.7e-07 at 16, varying run to run. The reduction order is not stable.

src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
```diff
@@ -372,7 +373,12 @@ double RadialProfiles::torfluxDeriv(double x) {
  */
 double RadialProfiles::torflux(double x) {
   //  Analytic evaluation of the polynomial (0 at x=0)
-  //  using Horner's method
+  //  using Horner's method.
+  //  educational_VMEC integrates torflux_deriv with a 100-interval trapezoid in
```
jurasic-pf: ? I don't know why this comment is here, educationalVmec was a stepping stone to get to vmecpp, not a "golden reference"

CharlesCNorton: Rewritten so the comment stands on its own: the reason `torflux` evaluates the integral in closed form is that `phipf` is built from `torfluxDeriv` and `maxToroidalFlux` divides by `torflux(1)`, so the profile has to be scaled by the value it actually integrates to for the edge flux to come out at `signOfJacobian * phiedge / 2 pi`. A quadrature there would divide by something else. No mention of the Fortran, here or in the two other places in this PR where I had leaned on it the same way (the `maxPoloidalFlux` comment and the test above `MaxToroidalFluxNormalizesTheEdgeFluxToPhiedge`).

`MaxToroidalFluxNormalizesTheEdgeFluxToPhiedge` is the check that the property holds, for `aphi` of one, three and four coefficients.

PR #754, opened 2026-09-01
Cover the four uncovered fixed-boundary flag combinations

CharlesCNorton:
Seven of the sixteen (lasym, 3D, ncurr, lfreeb) combinations in `vmecpp_large_cpp_tests/test_data/README.md` are covered; the file marks fourteen TODO. This fills the four fixed-boundary cells:

| case | lasym | 3D | ncurr |
|---|---|---|---|
| `cth_like_fixed_bdy_iota` | F | T | 0 |
| `up_down_asym_current` | T | F | 1 |
| `cth_like_fixed_bdy_asym` | T | T | 1 |
| `cth_like_fixed_bdy_asym_iota` | T | T | 0 |

`cth_like_fixed_bdy_iota` is the only three-dimensional case with `ncurr = 0`. The two asymmetric cth_like cases carry `max|rmns| = 1.4e-02`; `cth_like_free_bdy_asym`, in the `T T T T` cell, has `lasym = true` with empty `rbs` and `zbc`.

Inputs come from `indata2json`. The scalars are from an educational_VMEC run of each case, compared at 1e-5. The two constrained-iota cases also check `iotas` against the `ai` polynomial to 1e-12.

The four remaining TODOs are free-boundary and need an mgrid per symmetry class.

`vmec_test` goes from `medium` to `large`. It runs 148s under tsan on main against the 300s a medium test gets, and the four cases roughly double that.


jurasic-pf (approved):
Good idea!

PR #757, opened 2026-09-01
Compare the spline profile knots and values in the wout round trip

CharlesCNorton:
The wout round trip compared `am`, `ai` and `ac` but not the spline knots and values, and no case in the test used a spline profile. `cth_like_fixed_bdy_spline_pressure` had an input but no reference, so this adds one from educational_VMEC along with the Fortran namelist it was run from, puts the case in the test, and compares all six of `am_aux_s/f`, `ai_aux_s/f` and `ac_aux_s/f`.

The arrays are written for every profile whether or not it is a spline, with -1 knots and zero values where unused, so the comparison covers all seven cases. The reference is trimmed to the length VMEC++ holds, the same way the profile coefficients above it are: that length is the `ndatafmax` of the Fortran build the reference came from, and the references in this repo carry both 101 and 1001.

Scaling `am_aux_f[3]` by 1 + 1e-3 fails only the spline case, since every other case has that array all zeros.


CharlesCNorton:
The opt job fails on `output_quantities_test.cc:339`, `CheckLowpassFilterBSubsS/0`. On an unmodified base checkout the worst deviation for solovev is 2.1e-13 against its 1e-12 tolerance, a margin of 4.7. The other cases in that suite sit between 25 and 350 times below theirs:

| case | worst | tolerance |
|---|---|---|
| solovev | 2.1e-13 | 1e-12 |
| solovev_no_axis | 1.9e-14 | 1e-12 |
| cth_like_fixed_bdy | 4.0e-15 | 1e-12 |
| cth_like_fixed_bdy_nzeta_37 | 2.8e-15 | 1e-12 |
| cma | 3.9e-12 | 1e-10 |
| cth_like_free_bdy | 5.7e-13 | 2e-11 |

This PR adds an input file, a reference wout, two BUILD entries and a test in a different binary.

PR #758, opened 2026-09-01
Check Sizes against the fixaray reference in one parameterized test

CharlesCNorton:
`sizes_test.cc` was six copies of the same test with the expected values written out by hand, and a TODO asking for a debugging output from educational_VMEC to compare against instead. That output already exists: `fixaray` dumps `nzeta`, `ntheta3`, `nznt`, `mnsize`, `mnmax`, `mnyq`, `nnyq` and `mnmax_nyq`, which is where the reference computes them, and every case in the test has one.

So the six tests become one parameterized test that reads the dump for those eight, and checks the rest against the input it came from: `lasym`, `nfp`, `mpol` and `ntor` are echoes, `ntheta` follows from `mpol` when the input leaves it at zero, and `lthreed`, `num_basis`, `nThetaEven`, `nThetaReduced` and `nThetaEff` follow from those. The integration weights are checked for the half weight on the endpoints of a stellarator-symmetric half period and for integrating to one over the full grid, rather than against a literal per case.

347 lines become 96, and adding a case is now a line in the instantiation.

Adding one to `mnsize` fails every case with both sides named.


jurasic-pf (approved):
good simplification

PR #761, opened 2026-09-01
Check the mgrid stellarator symmetry instead of assuming it

CharlesCNorton:
`CheckLoadMGrid` compares the stored mgrid field against an ABSCAB evaluation of the coil set over the first half field period and hard-coded the assumption that the file was computed with stellarator symmetry. That assumption is checkable against the stored data: reflecting a plane onto `(2 pi / nfp - phi, R, -Z)` leaves `B_phi` and `B_Z` unchanged and flips the sign of `B_R`. The new check covers all `kp` planes, including those the point-wise comparison does not reach, and fails when the expected parity of `B_R` is flipped.

The measured residual of the relation is 6.9e-5 of the largest field component; `coils.cth_like` is not perfectly stellarator-symmetric, and the first of its two coil sets carries the whole deviation, 7e-5 against 1e-16. The bound is 2e-4. The stored field past the half period is a symmetry image of the first half rather than an independent evaluation, disagreeing with a direct ABSCAB evaluation there by 1e-6 to 1e-5 relative.

### DIFF
--- a/src/vmecpp/cpp/vmecpp_large_cpp_tests/free_boundary/mgrid_provider/mgrid_provider_test.cc
+++ b/src/vmecpp/cpp/vmecpp_large_cpp_tests/free_boundary/mgrid_provider/mgrid_provider_test.cc
@@ -130,24 +130,11 @@ TEST_P(LoadMGridTest, CheckLoadMGrid) {
 
   ASSERT_EQ(nc_close(ncid), NC_NOERR);
 
-  // TODO(jons): A flag if stellarator symmetry was used in computing a given
-  // mgrid file is not stored in the mgrid file. For now, hard-code this to
-  // `true`, since all our test cases assume stellarator symmetry. To be revised
-  // when a) we use non-stellarator-symmetric coil sets _and_ b) we have
-  // transitioned to only using our own `makegrid`, in which we can define new
-  // output variables and have the MakegridParameters at hand anyways.
-  bool assume_stellarator_symmetry = true;
-
-  // NOTE: The coil geometry in `coils.cth_like` was found to not be perfectly
-  // stellarator-symmetric. Therefore, the resulting magnetic field is also not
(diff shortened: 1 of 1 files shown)


jurasic-pf:
Why is the field not perfectly stellarator symmetric? Is the discretization of the coils asymmetric? If so this needs fixing

CharlesCNorton:
The coil geometry is symmetric. Under the 2π/5 rotation vertex k maps to k+120 (the filament makes two toroidal turns), and with that map the 200 vertices are invariant under the full group to 5.9e-6, the rounding of six-digit coordinates. The grid comes within a few centimetres of the inner leg, so that rounding reaches 6.9e-5 in the field on the mid-period plane, the only plane that is not a mirror copy. Averaging each vertex over its ten images removes it (residual 1.8e-11, vertices moved by at most 3.6e-6) but shifts the cth free-boundary equilibrium by 6e-7 in iota, past the 2e-9 tolerances of the educational_VMEC references, so the corrected coils file comes with regenerated references: #781. The comment here now states the cause.

jurasic-pf:
Okay, I forgot that the coils file carried the fully symmetrized coils. Sorry for that.

Can we lower the test tolerances now that, the fix to the reference files is merged?

jurasic-pf (approved):

CharlesCNorton:
Lowered from 2e-4 to 1e-9. With the symmetrized coils and regenerated mgrid from #781 the measured residual is 4.7e-11 of the largest field component, down from 6.9e-5, so the bound now sits about 20x above it instead of 3x.

CharlesCNorton:
The sanitizer failures here are AxisymmetricRunIsIndependentOfNzeta in vmec_test, a different test binary from the one this PR touches: lmns_full reads 1.06e-08 under asan and 1.46e-08 under ubsan against a 1e-8 bound. #815 raises that lambda bound to 1e-6 with the measured spread behind it.

jurasic-pf (approved):

PR #763, opened 2026-09-01
Fold the per-thread contributions into shared accumulators in thread order

CharlesCNorton:
Eleven per-iteration reductions folded each thread's contribution into a shared accumulator inside `#pragma omp critical`, which admits threads in whichever order they arrive. The floating-point sum therefore depended on thread timing, and the run varied.

On `cma` at 8 threads, 224 of the 438 datasets in the output file differed between three runs of the same binary. `fsqr`, `fsqz` and `fsql` were among them, so the force residuals themselves moved; `jdotb` moved by 9.0e-8 relative.

## What replaces them

The reductions in `ideal_mhd_model` and `radial_profiles` are now folded per surface: each thread writes its own surfaces' rows into shared scratch, and one thread sums those rows in surface order (`SumSurfaceRows` in `util.h`). The grouping of the sum follows the surface index alone, so the total is identical for any number of threads, not merely reproducible at a fixed one.

| file | accumulator |
|---|---|
| `ideal_mhd_model.cc` | `fResInvar`, `fResPrecd`, `voli`, `plasmaVolume`, `thermalEnergy` and `magneticEnergy`, `fNormRZ` and `fNormL` and `fNorm1` |
| `radial_profiles.cc` | the volume-averaged spectral width |

The four free-boundary sites take turns by thread id instead (`AddInThreadOrder`): `matrixShare` is `mnpd_dim` squared, so a tree or a per-surface fold would need one copy of it per thread. They run once per vacuum solve, where the extra barriers do not matter.

| file | accumulator |
|---|---|
| `laplace_solver.cc` | `matrixShare`, `bvecShare` |
| `nestor.cc`, `only_coils.cc` | `bSubUVac`, `bSubVVac` |

An OpenMP `reduction` clause does not give either property: the standard leaves the order in which the partial results are combined unspecified.

## Result

Three runs of `cma` at 8 threads agree on all 438 datasets, as do three runs at 16 threads and three runs of `cth_like_free_bdy`, which covers the NESTOR and Laplace solver sites. With the per-surface fold the `ideal_mhd_model` totals also agree between 1, 8 and 16 threads.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/util/util.h
+++ b/src/vmecpp/cpp/vmecpp/common/util/util.h
@@ -35,6 +35,39 @@ namespace vmecpp {
 using RowMatrixXd =
     Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
 
+// Folds one contribution per thread into a shared total: every thread writes
+// its own row of `m_slots`, then a single thread sums the rows in thread order.
+//
+// A critical section admits the threads in whichever order they arrive, so the
+// floating-point sum it produces varies from run to run. The order here is the
+// thread id, so the total is the same on every run at a given thread count.
+//
+// Every thread of the team must call this with the same arguments. The barrier
+// that ends the fold publishes `m_total`, so no further barrier is needed
+// before reading it.
+inline void SumOverThreads(const double *contribution, int width, int thread_id,
+                           int num_threads, double *m_slots, double *m_total) {
+  double *row = m_slots + thread_id * width;
+  for (int i = 0; i < width; ++i) {
+    row[i] = contribution[i];
+  }
+#ifdef _OPENMP
+#pragma omp barrier
+#pragma omp single
+#endif  // _OPENMP
+  {
+    for (int i = 0; i < width; ++i) {
+      m_total[i] = 0.0;
+    }
+    for (int t = 0; t < num_threads; ++t) {
+      const double *slot = m_slots + t * width;
+      for (int i = 0; i < width; ++i) {
+        m_total[i] += slot[i];
+      }
+    }
+  }
+}
+
 inline Eigen::VectorXd ToEigenVector(const std::vector<double> &v) {
   return Eigen::Map<const Eigen::VectorXd>(v.data(),
                                            static_cast<Eigen::Index>(v.size()));
--- a/src/vmecpp/cpp/vmecpp/free_boundary/laplace_solver/laplace_solver.cc
+++ b/src/vmecpp/cpp/vmecpp/free_boundary/laplace_solver/laplace_solver.cc
@@ -4,10 +4,10 @@
 // SPDX-License-Identifier: MIT
 #include "vmecpp/free_boundary/laplace_solver/laplace_solver.h"
 
+#include <algorithm>
 #include <cmath>
 #include <vector>
 
-#include "absl/algorithm/container.h"
 #include "absl/log/check.h"
 
 namespace vmecpp {
@@ -17,7 +17,7 @@ LaplaceSolver::LaplaceSolver(
     const TangentialPartitioning* tp, int nf, int mf,
     std::span<double> matrixShare,
     Eigen::PartialPivLU<Eigen::MatrixXd>* lu_decomposition,
-    std::span<double> bvecShare)
+    std::span<double> bvecShare, std::span<double> reduce_slots)
     : s_(*s),
       fb_(*fb),
       tp_(*tp),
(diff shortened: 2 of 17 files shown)


jurasic-pf (changes requested):
If you're already refactoring what are fundamentally reductions, I'd prefer if we could use a reduction clause with log2 N scaling instead of linear.

There are custom parallel reduction primitives implemented in new OpenMP versions https://www.openmp.org/spec-html/5.0/openmpsu107.html and https://theartofhpc.com/pcse/omp-reduction.html
which should enable all the patterns needed here. For the scalar reduction they may just be expressible as a native reduce call, if I'm not mistaken. I'm just not sure how it interacts with the parallel region we're keeping open the entire time.

CharlesCNorton:
Done: the per-iteration reductions now fold a fixed binary tree over a shared `[num_threads, 4]` scratch, so a reduction costs ceil(log2 N) barriers instead of N.

A native `reduction` clause is not deterministic: the spec leaves the combining order unspecified, and libgomp combines in arrival order. The tree pins the pairing to thread ids.

The four free-boundary sites keep the turn chain: `matrixShare` is too large for per-thread rows, and the rest run once per vacuum solve. Re-verified: three runs of `cma` and `cth_like_free_bdy` at 8 threads, all 438 datasets bit-identical.

CharlesCNorton:
fixing clang

jurasic-pf:
Imo it's fine if it's not deterministic, if it's just up to thread order. It will still be up to the number of threads though, won't it?

jurasic-pf:
Or not even that... Hmmm it might be nice actually

CharlesCNorton:
Done. Every per-iteration sum now folds one partial per radial surface in surface order, including rmsPhiP, the lambda normalization, which was summed per thread at initialization. The output file is bit-identical across 1, 8 and 16 threads for cma, cth_like_fixed_bdy, cth_like_fixed_bdy_nzeta_37 and solovev, all 438 datasets each. The free-boundary vacuum accumulations keep the thread-order chain, so a free-boundary run is reproducible at a fixed thread count only.

CharlesCNorton:
The tree landed in 61cd862, and 377e2e5 then replaced it: the `ideal_mhd_model` and `radial_profiles` reductions are now folded per surface, so the grouping follows the surface index and the total is identical for any thread count rather than only reproducible at a fixed one. Three runs of `cma` agree on all 438 datasets at 8 and at 16 threads, and the totals agree between 1, 8 and 16.

A `reduction` clause cannot give either property: the standard leaves the order in which partial results are combined unspecified.

The four free-boundary sites still take turns. Their accumulator is the `mnpd_dim` squared vacuum response matrix, so a tree needs one copy per thread; they run once per vacuum solve.

jurasic-pf:
Ok, sorry for the back and forth - if we can't use openmp native constructs, then I would prefer we go for the simplest communication avoiding solution, sketched like here:

// Barrier 1: Ensure all threads finished writing their local_sums
    #pragma omp barrier

    // 2. Deterministic reduction by a single thread
    #pragma omp master  // or omp single which needs synchronization
    {
        for (int t = 0; t < NUM_THREADS; t++) {
            global_sum += local_sums[t];
        }
    } // Implicit barrier here at the end of '#pragma omp single'

    // 3. SAFE: All threads can now read

This trades less barriers for more serial work.

Further, please remove the changes that are unrelated to this from the PR.

CharlesCNorton:
Rewritten to the sketched shape. Every thread writes its contribution into its own row of a shared slot buffer, one barrier follows, and a single thread sums the rows in thread order; the implicit barrier at the end of the `single` publishes the total. That is the one mechanism now, `SumOverThreads` in `util.h`, used at all eleven sites. The per-surface fold and the turn-taking chain are both gone.

Three runs of `cma` agree on all 103 wout fields at 8 threads and at 16, and three runs of `cth_like_free_bdy`, which covers the Nestor and Laplace solver sites, agree on all 106.

The summation order is the thread id, so the grouping follows the thread count: 50 of the 103 fields of `cma` differ between 1, 8 and 16 threads, as they did before this PR. The per-surface fold gave the stronger property of a total independent of the thread count; this trades that for the simpler code, as you asked.

The vacuum team needs one row per thread of its widest sum, the response matrix, so `Vmec` now holds `vac_num_threads_ * mnpd_dim^2` doubles: 4 MB at 8 threads for a case with `mpol = 12, ntor = 12`, 66 KB per thread for the cth-like cases. Say the word if that is too much and I will keep the two large vacuum sums on a different footing.

The diff is now confined to the reduction sites and the plumbing they need. The changes to `fourier_coefficients`, `fourier_forces`, the pybind layer and the fourier-geometry test are gone.

PR #766, opened 2026-09-01
Add a content-addressed equilibrium cache

CharlesCNorton:
`EquilibriumCache` stores every converged `VmecOutput` under a SHA-256 of the canonicalized input and answers repeated solves from the store. For a free-boundary case the `mgrid_file` path is replaced by a hash of the file content before keying, so the key follows the vacuum field rather than the filesystem.

A repeated input is answered from the store. With `verify=True`, the default, the stored state is passed to `vmecpp.run` as `restart_from` and the solver re-establishes force balance in a few iterations; with `verify=False` the stored output is returned directly. A new input hot-restarts from the nearest compatible stored neighbor, interpolated to the requested resolution with `interpolate_solution`, and falls back to a cold solve if the restart fails. On solovev this warm start beats a cold solve of the same grid for boundary changes below about 1e-4 relative, and `warm_start=False` turns it off.

`submit` runs a solve on a background thread and returns a `concurrent.futures.Future`; the solver releases the GIL, so up to `max_parallel_solves` solves run concurrently. The store is a directory of gzipped JSON entries with an input sidecar each, written atomically and shareable between processes and machines. The tests cover key stability, the mgrid content addressing, both hit paths, the warm start, reopening a store from disk, and the async path.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -21,6 +21,7 @@ import numpy as np
 import pydantic
 
 from vmecpp import _util
+from vmecpp._cache import EquilibriumCache, cache_key
 from vmecpp._continuation import _run_fourier_continuation, interpolate_solution
 from vmecpp._free_boundary import (
     MagneticFieldResponseTable,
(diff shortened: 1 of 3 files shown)


CharlesCNorton:
Same scope reasoning as #767: an orchestration layer, not the solver. Closing.

PR #767, opened 2026-09-01
Add bootstrap-consistent equilibria via the Redl formula

CharlesCNorton:
A reactor-relevant current profile is set by the equilibrium itself: the bootstrap current depends on the geometry and the geometry on the current. `bootstrap_consistent` closes that loop for `ncurr = 1`. Each pass evaluates the Redl `<J.B>` (Redl et al., Physics of Plasmas 28, 022502 (2021)) on the converged equilibrium, integrates it to an enclosed-current profile, under-relaxes the constraint toward it as a `cubic_spline_i` profile with matching `curtor`, and re-solves by hot restart from the previous solution. On `cth_like_fixed_bdy` with `ne = 1e20 (1 - 0.8 s)` and `Te = Ti = 1500 eV (1 - 0.8 s)`, three solves bring the residual from 0.67 through 0.37 to 0.045, driving 102 kA of bootstrap current, and the hot-restarted solves take ~50 iterations each against the cold count.

The Redl evaluation comes from `simsopt.mhd.bootstrap`, which is already a dependency; this module contributes the geometry extraction from a `VmecOutput` and the loop. Profiles are callables of `s` (`ne` in 1/m^3, `Te`/`Ti` in eV, `Zeff` constant); anything with a `dfds` method, such as a simsopt `Profile`, supplies its own derivative.

Convergence is judged against the equilibrium rather than the loop's own inputs: the flux-function identity `mu0 <J.B> = (G I' - I G') <B^2>/(G + iota I)`, with `G` and `I` read from `bvco` and `buco`, evaluates the current the equilibrium actually carries, and the loop ends when that matches the Redl profile to the requested tolerance. The tests pin each convention the module rests on: the reconstructed `<B^2>` against `wout.bdotb`, `signgs * 2 pi * buco_edge / mu0` against `wout.ctor`, the identity against `wout.jdotb` (median ratio 0.999), the `<J.B>`-to-current integration against its own inverse, and the loop on `cth_like_fixed_bdy` end to end.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -21,6 +21,11 @@ import numpy as np
 import pydantic
 
 from vmecpp import _util
+from vmecpp._bootstrap import (
+    BootstrapConsistentResult,
+    bootstrap_consistent,
+    redl_bootstrap_current,
+)
 from vmecpp._continuation import _run_fourier_continuation, interpolate_solution
 from vmecpp._free_boundary import (
     MagneticFieldResponseTable,
@@ -2712,6 +2717,9 @@ populate_raw_profile = set_profile
 __all__ = [  # noqa: RUF022
     "run",
     "interpolate_solution",
+    "bootstrap_consistent",
+    "redl_bootstrap_current",
+    "BootstrapConsistentResult",
     "rescale",
     "VmecInput",
     "VmecOutput",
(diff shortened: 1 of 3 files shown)


jurasic-pf (changes requested):
Unfortunately I don't think the place for this is in `vmecpp`. vmecpp is really just intended as a solver, the `simsopt` dependency is also gonna disappear in the medium term. A bootstrap self consistent profile would be a specific usage example of it, and might be better fit as an intermediate level example in `simsopt/examples`

CharlesCNorton:
Agreed, the loop belongs downstream of the solver. Closing.

PR #768, opened 2026-09-01
Warn when a resonant harmonic marks a suppressed island chain

CharlesCNorton:
VMEC's nested-surface ansatz cannot open an island, so where the real field would, the resonance shows up instead as a locally enhanced (m, n) harmonic of |B| pinned to its rational surface. `ComputeIslandDiagnostics` scans the rotational transform profile for crossings of the field-period-compatible rationals n/m with n a multiple of nfp, gcd(m, n/nfp) = 1 and m up to a bound of 6 by default, and evaluates the resonant harmonic at each crossing: its amplitude normalized by |B|_00, the median of the same harmonic over the off-resonance surfaces, and the quasi-linear island width estimate 4 sqrt(amplitude / (m |iota'|)) in s.

After the wout is computed, a crossing whose resonant amplitude is at least twice its off-resonance level prints a warning naming the rational, its position, the enhancement factor and the width estimate. The shipped equilibria do cross low-order rationals (cma crosses 4/7, cth_like_fixed_bdy crosses 5/5 and 5/4, li383 crosses 3/5, 3/6 and 3/7), and none of them triggers the warning, so the default suite output is unchanged.

The test runs cma, whose only low-order crossing is 4/7: the default bound of 6 finds nothing, admitting m = 7 finds exactly that crossing at the surface where iotaf crosses 4/7 to 1e-12, and pinning an artificial bump of the (7, 4) harmonic to that surface raises the amplitude past the warning threshold against an unchanged background.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -6,12 +6,18 @@
 
 #include <Eigen/Dense>  // VectorXd
 #include <algorithm>
+#include <cmath>
+#include <cstddef>
+#include <iostream>
+#include <limits>
 #include <memory>
+#include <numeric>  // gcd
 #include <string>
 #include <vector>
 
 #include "H5Cpp.h"
 #include "absl/log/check.h"
+#include "absl/strings/str_format.h"
 #include "util/hdf5_io/hdf5_io.h"
 #include "util/testing/numerical_comparison_lib.h"
 #include "vmecpp/common/util/util.h"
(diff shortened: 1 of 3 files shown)


jurasic-pf:
While this sort of diagnostic may be useful, I see this intentionally out of scope for the vmecpp repo and more on the user to implement.

Even more, I have my doubts that a, B based diagnostics can reliably detect formation of resonant features like islands.Instead I expect formation of near singular currents that create an opposing resonant field to satisfy flux surfaces. Do you have any evidence to prove otherwise? Is there poincare plots or verifications with other solvers that indicate this metric correlates with the actual island width observed for a rational crossing?

jurasic-pf:
If not I'd rather close the issue.
If there's evidence that this is a good metric for predicting emergence of islands we can merge it as a post processing example, but not as part of the core solver :)

CharlesCNorton:
No evidence beyond the artificial-bump test. Closing.

PR #769, opened 2026-09-01
Add opt-in Anderson acceleration of the descent iteration

CharlesCNorton:
**Change** - `VmecInput.anderson_acceleration`, off by default, switches on Anderson acceleration of the descent iteration (JSON and HDF5 round trip through `VmecINDATA`); `Vmec::SetAndersonAcceleration(window, start_iteration, frequency, zero_velocity_on_jump)` keeps the tuning parameters for C++ callers. One Garabedian time step maps the Fourier state x to g(x). `AndersonAcceleration` keeps a short per-thread history of map outputs and residuals r = g - x; the team solves the least-squares problem for the combination of recent residual differences that best cancels the current residual, each thread applies that combination to its map outputs, and the momentum restarts. A thread stores the history of its own slice over its full stored radial range, satellite points included, so the combination stays consistent at the partition seams. The k x k normal equations are the only cross-thread data: `SumInFixedTree` reduces them over a binary tree fixed by thread id, they are Tikhonov-regularized by 1e-12 of their mean diagonal, and a combination that is not finite or exceeds 100 in magnitude is skipped. The defaults are a window of 2, engagement 25 iterations after a stage starts or restarts, and disengagement once fsqr + fsqz + fsql falls below max(1e-4, 1e-3 x the residual at stage entry, 1e3 ftol). Free-boundary runs are included.

**Cause** - the extrapolation pays in the nonlinear phase at the start of each multigrid stage, where the descent is still finding the basin. In the tail the fixed-point map is linear and the momentum iteration already converges at its optimal rate; extrapolating there held the residual in a limit cycle on the QUASR fixed-boundary cases at ftol 1e-9, so the tail is handed back to the descent. An earlier study ran Anderson from fsq = 1e-6 down, where it loses.

**Evidence** - iterations to convergence on the QUASR configurations of `tests/test_free_boundary_quasr.py` (ns [8, 16, 31], mpol = ntor = 6, ftol 1e-9, 8 threads), summed over the multigrid stages. Fixed boundary:

| case | plain | accelerated |
|---|---|---|
| 954 | 414 | 394 |
| 9914 | 739 | 546 |
| 19493 | 2186 | 1816 |
| 19609 | 1719 | 1472 |
| 19940 | 1311 | 1122 |
| 29346 | 1079 | 959 |
| 50136 | 3866 | 3894 |
| 65579 | 3342 | 3279 |
| 112718 | 1789 | 1671 |
| 165868 | 1288 | 1152 |
| 167381 | 1277 | 1197 |
| 336902 | 1314 | 1280 |
| total | 20324 | 18782 |

Free boundary:

| case | plain | accelerated |
|---|---|---|
| 954 | 1512 | 1414 |
| 9914 | 1339 | 1021 |
| 19609 | JACOBIAN_75_TIMES_BAD | 1433 |
| 19940 | 1546 | 1285 |
| 29346 | 1391 | 1302 |
| 65579 | 3212 | 2672 |

The other six free-boundary cases do not converge either way. Accelerated and plain runs converge to the same equilibrium: over the twelve fixed-boundary cases they differ by less, in `wb`, axis, iota profile and every Fourier array, than the plain run at ftol 1e-9 differs from the plain run at ftol 1e-11. Repeated 8-thread accelerated runs agree in all 438 output datasets.

**Tests** - `AndersonAcceleratedCthMatchesPlainIteration` in `vmec_test` solves `cth_like_fixed_bdy` both ways, requires fewer iterations with the flag, and holds each equilibrium-defining quantity to at least five times its measured deviation between the two paths. `test_run_honors_anderson_acceleration_flag` in `tests/test_iteration.py` drives the input flag through `vmecpp.run`.

**Scope** - with the flag unset the window is 0, no history is allocated and the time step is the one that was there.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -425,6 +425,15 @@ class VmecInput(BaseModelWithNumpy):
     """Time-step / restart control scheme for the equilibrium iteration (``"vmec_8_52"``
     or ``"parvmec"``)."""
 
+    anderson_acceleration: bool = False
+    """Accelerate the equilibrium iteration by Anderson extrapolation of the Garabedian
+    time step from a short history of iterates.
+
+    The extrapolation acts in the nonlinear early phase of each multigrid stage and
+    hands the tail back to the plain descent. The accelerated and the plain iteration
+    converge to the same equilibrium within the force tolerance.
+    """
+
     nstep: int = 10
     """Printout interval at which convergence progress is logged."""
 
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
@@ -228,6 +228,7 @@ VmecINDATA::VmecINDATA() {
   tcon0 = 1.0;
   lforbal = false;
   iteration_style = IterationStyle::VMEC_8_52;
+  anderson_acceleration = false;
   return_outputs_even_if_not_converged = false;
 
   // zero-initialized magnetic axis
@@ -352,6 +353,7 @@ absl::Status VmecINDATA::WriteTo(H5::H5File& file) const {
   WriteH5Dataset(delt, "/indata/delt", file);
   WriteH5Dataset(tcon0, "/indata/tcon0", file);
   WriteH5Dataset(lforbal, "/indata/lforbal", file);
+  WriteH5Dataset(anderson_acceleration, "/indata/anderson_acceleration", file);
   WriteH5Dataset(return_outputs_even_if_not_converged,
                  "/indata/return_outputs_even_if_not_converged", file);
 
@@ -454,6 +456,15 @@ absl::Status VmecINDATA::LoadInto(VmecINDATA& m_indata, H5::H5File& from_file) {
   ReadH5Dataset(m_indata.delt, "/indata/delt", from_file);
   ReadH5Dataset(m_indata.tcon0, "/indata/tcon0", from_file);
   ReadH5Dataset(m_indata.lforbal, "/indata/lforbal", from_file);
+
+  // Added after the initial schema; files written before it load the
+  // default.
+  if (H5Lexists(from_file.getId(), "/indata/anderson_acceleration", 0) == 1) {
+    ReadH5Dataset(m_indata.anderson_acceleration,
+                  "/indata/anderson_acceleration", from_file);
+  } else {
+    m_indata.anderson_acceleration = false;
+  }
 
   // Legacy way of checking for dataset existence
   if (H5Lexists(from_file.getId(),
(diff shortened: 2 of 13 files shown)


jurasic-pf:
A few comments on architecture - please don't gate functionality behind env variables, the canonical way people should toggle features is via boolean settings in the `VmecInput` objects.
Other than that, Anderson acceleration is an interesting idea! I briefly looked into it a while ago and concluded that it doesn't capture the sort of non linear convergence behavior we require very well, so I closed the PRs again, curious to understand why your results are different in that regard. How does it perform on a wider range of cases, like the QUASR free boundary runs I added to the test suite? You can run them in fixed boundary too

CharlesCNorton:
The environment variable is gone. The feature is the boolean `anderson_acceleration` on `VmecInput` and `VmecINDATA`, off by default.

The difference from Finding 13 is the regime. That study ran Anderson on the tail, from fsq = 1e-6 down, where it loses; the gain here is in the nonlinear phase at the start of each multigrid stage. The earlier gate mixed the two (its floor max(1e3 ftol, 1e-8) is 1e-6 on QUASR at ftol 1e-9), which is why the first QUASR runs lost. The gate is now: window 2, disengage below max(1e-4, 1e-3 x the residual at stage entry, 1e3 ftol), free boundary included.

QUASR inputs of `tests/test_free_boundary_quasr.py` (ns [8, 16, 31], mpol = ntor = 6, ftol 1e-9), iterations summed over the stages. Fixed boundary:

| case | plain | accelerated |
|---|---|---|
| 954 | 414 | 394 |
| 9914 | 739 | 546 |
| 19493 | 2186 | 1816 |
| 19609 | 1719 | 1472 |
| 19940 | 1311 | 1122 |
| 29346 | 1079 | 959 |
| 50136 | 3866 | 3894 |
| 65579 | 3342 | 3279 |
| 112718 | 1789 | 1671 |
| 165868 | 1288 | 1152 |
| 167381 | 1277 | 1197 |
| 336902 | 1314 | 1280 |
| total | 20324 | 18782 |

Free boundary:

| case | plain | accelerated |
|---|---|---|
| 954 | 1512 | 1414 |
| 9914 | 1339 | 1021 |
| 19609 | JACOBIAN_75_TIMES_BAD | 1433 |
| 19940 | 1546 | 1285 |
| 29346 | 1391 | 1302 |
| 65579 | 3212 | 2672 |

The other six free-boundary cases do not converge either way. Accelerated and plain runs converge to the same equilibrium: over the twelve fixed-boundary cases they differ by less, in wb, axis, iota profile and every Fourier array, than the plain run at ftol 1e-9 differs from the plain run at ftol 1e-11.

jurasic-pf:
One thing I looked at was just an animation of how flux surfaces evolve during a solve starting from initialization. the descent has a characteristic dynamic when doing it where you can clearly see the cross coupling terms and long tail contributions in the eigenvalue spectrum taking much longer than roughly nesting the surfaces.

I wonder if side by side comparisons would be informative here too, unfortunately i cant find the animation script anymore

jurasic-pf:
also just plain fsqt over iteration plots for all the configs before and after might be helpful, just generally plots to better understand what's going on

CharlesCNorton:
Residual traces and surface animations, all from the PR branch with the flag off and on (8 threads, the QUASR inputs of the test suite).

Invariant residual per iteration, fixed boundary (plain blue, accelerated red; the spikes are the multigrid interpolations):

QUASR fixed boundary residuals

Free boundary. The cases that fail do so in the sawtooth plateau of the free-boundary descent (restart, decay to 1e-5, restart); 19609 leaves it once accelerated, and 112718 never engages because its plateau sits below the gate:

QUASR free boundary residuals

Flux surfaces during the first stage (ns = 8), plain left and accelerated right, at zeta = 0 and pi/nfp; each frame is the state after k iterations from a run truncated at k, which traces the same deterministic path. 954 fixed boundary:

954 surfaces

19609 free boundary, where the plain run cycles until JACOBIAN_75_TIMES_BAD and the accelerated one converges (frames spaced geometrically):

19609 surfaces

The residual histories are in `fsq_fixed.csv` and `fsq_free.csv` on the same branch, and the two scripts (`fsq_plots.py`, `surface_animation.py`) next to them.

PR #770, opened 2026-09-01
Add the Boozer transform of a converged equilibrium

CharlesCNorton:
`BoozerTransform` (C++, `vmecpp/vmec/boozer`) and `vmecpp.boozer_transform` (Python) compute the Boozer coordinates of the half-grid flux surfaces of a converged stellarator-symmetric equilibrium from its `wout`: the Boozer spectra of |B|, R, Z, the angle transformation function nu and the Boozer Jacobian, together with G, I and iota per surface and the mode numbers in the `wout` convention.

The Boozer angles are theta_B = theta + lambda + iota nu and zeta_B = zeta + nu, with nu = (w - I lambda) / (G + iota I) and w the periodic part of the magnetic scalar potential on the surface, assembled mode by mode from B_theta and B_zeta. Each spectrum is a direct quadrature over the VMEC angles carrying the Jacobian of the angle map, so the map is never inverted. Geometry is taken to the half grid as booz_xform does, odd-m coefficients interpolated as coefficient / sqrt(s) with the m = 1 quotient extrapolated to the axis on the first half-grid surface, and the Jacobian is reported as (G + iota I) / |B|^2 in the Boozer angles. `jacobian_spread` gives per surface the relative variation of sqrt(g_B) |B|^2, which is a flux function in exact Boozer coordinates and so measures the residual of the transformation.

Verification is `tests/test_boozer.py` and `vmecpp/vmec/boozer:boozer_test`. On `cth_like_fixed_bdy` with mboz = nboz = 8 every surface and mode of `bmnc_b`, `rmnc_b`, `zmns_b`, `numns_b` and `gmnc_b` agrees with booz_xform 0.1.0 to 1e-9 of the largest coefficient of the array (the test is skipped when booz_xform is not installed), and G and I agree with `bvco` and `buco` exactly. The C++ tests check that sqrt(g_B) |B|^2 is a flux function to the force tolerance of the case, that the Jacobian-weighted surface average of |B| is the same in the VMEC and Boozer angles, that an axisymmetric equilibrium has no toroidal Boozer harmonics, and the argument validation. Non-stellarator-symmetric equilibria return `Unimplemented`.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -2707,10 +2707,38 @@ def set_profile(
 populate_raw_profile = set_profile
 
 
+def boozer_transform(
+    wout: VmecWOut,
+    mboz: int,
+    nboz: int,
+    surfaces: typing.Sequence[int] | None = None,
+) -> _vmecpp.BoozerCoordinates:
+    """Boozer coordinates of the flux surfaces of a converged equilibrium.
+
+    Returns the Boozer spectra of |B| (``bmnc_b``), R (``rmnc_b``), Z
+    (``zmns_b``), the angle transformation function nu (``numns_b``, with
+    zeta_B = zeta + nu and theta_B = theta + lambda + iota nu) and the Boozer
+    Jacobian (``gmnc_b``, the expansion of (G + iota I) / |B|^2, whose radial
+    coordinate is the toroidal flux per radian, as booz_xform reports it), each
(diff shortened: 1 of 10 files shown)


jurasic-pf:
Hi Charles, unfortunately I don't this vmecpp is the place for this. Theres multiple tools in the stellarator community that solve this problem like booz_xform and interact perfectly fine with vmecpp

PR #772, opened 2026-09-02
Carry the sign of the Jacobian in chi

CharlesCNorton:
**Change** - `chi` carries the sign of the Jacobian: `ComputeIntermediateThreed1FirstTableQuantities` forms it as `2 pi * sign_of_jacobian * chi1`, and `wout.chi` follows.

**Cause** - `chi` was accumulated as `2 pi * chi1` with `chi1 = sum(phipH * iotaH) * deltaS`, while `phi`, `phipf` and `chipf` all carry `sign_of_jacobian`. With the usual negative Jacobian the wout therefore held `chi` = minus the integral of iota over phi. Fortran VMEC has the same omission (`eqfor.f`, `chi = twopi*chi1`) and PARVMEC 10.0 reproduces it, so the reference wout files hold the negated profile.

**Evidence** - on `cth_like_fixed_bdy`, `phi(ns)` is -0.035, the integral of `chipf` over s is -0.040, and `chi(ns)` was +0.040. `iota = dchi/dphi` read back from the file had the sign opposite to `iotaf`, and the threed1 `psi`, which does carry the sign, disagreed with `chi`.

**Tests** - the five comparisons against the reference files negate the reference for `chi`, with a note at each: `CheckWOutFileContents` in `output_quantities_test`, `CheckThreed1FirstTable` in the large `output_quantities_test`, `test_output_quantities` in `test_pybind_vmec.py`, and `test_vmecwout_io` and `test_against_reference_wout` in `tests/test_init.py`.

**Scope** - nothing else reads `chi1`, and the force balance never reads `chi`, so the converged equilibrium is unchanged.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -3520,8 +3520,11 @@ vmecpp::ComputeIntermediateThreed1FirstTableQuantities(
         (vmec_internal_results.phipH[jHi] * vmec_internal_results.iotaH[jHi]) *
             fc.deltaS;
 
+    // phi, phipf and chipf carry the sign of the Jacobian, so chi does too;
+    // iota = d(chi)/d(phi) then holds between the wout profiles.
     threed1_first_table_intermediate.chi[jF] =
-        2.0 * M_PI * threed1_first_table_intermediate.chi1[jF];
+        2.0 * M_PI * vmec_internal_results.sign_of_jacobian *
+        threed1_first_table_intermediate.chi1[jF];
   }  // jF
 
   // calc_fbal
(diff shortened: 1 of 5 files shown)


CharlesCNorton:
fixing

CharlesCNorton:
Since #584 landed, `MakeGeometry` accumulates the poloidal flux with `sign_of_jacobian`, so main now carries two poloidal fluxes of opposite sign: across solovev, cth_like_fixed_bdy, cth_like_fixed_bdy_asym, li383_low_res, cma, up_down_asym and circular_tokamak, `geometry.poloidal_flux / wout.chi` is exactly -1.000000000000.

Taking `d(chi)/d(phi)` from the file picks the same side: it matches `+iotas` to 3e-15 for the geometry profile and returns `-iotas` for `wout.chi`, which also equals minus the integral of the `chipf` stored beside it.

PR #773, opened 2026-09-02
Extrapolate chipF onto the axis like iotaF

CharlesCNorton:
`computeBContra` filled `chipF` on the interior full-grid points and, when the boundary is in range, at the edge, but never at the axis, so `wout.chipf[0]` stayed at its zero initialisation while `iotaf[0]` and `phipf[0]` are extrapolated there. `chipf/phipf` equals `iotaf` on every other surface and is 0 at the axis; `q = phipf/chipf` is infinite there, and any radial interpolation through `chipf` dips to zero at `s = 0`.

The axis now gets the rule `iotaF` uses, `1.5 * chipH[0] - 0.5 * chipH[1]`. This is the rule PARVMEC added to `add_fluxes.f90` in 2016 (`chipf(1) = c1p5*chips(2) - p5*chips(3)`, "SPH ADDED THIS 4-8-16"); the 8.52 lineage behind the reference files never received it, so the references carry a zero there and the comparisons skip the axis entry of `chipf`, with a note at each. The `ideal_mhd_model` test compares `chipF` from `nsMinFi` on and is unaffected. #736 handles the boundary entry of the same profile; the two changes are independent.

`bazel test //vmecpp/vmec/output_quantities:output_quantities_test`, `tests/test_init.py` and `tests/cpp/vmecpp/vmec/pybind11/test_pybind_vmec.py` pass.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/ideal_mhd_model/ideal_mhd_model.cc
@@ -1790,6 +1790,12 @@ void IdealMhdModel::computeBContra() {
   }
 
   // update full-grid chi'
+  if (r_.nsMinF1 == 0) {
+    // The axis value is extrapolated the same way as iotaF below. Fortran VMEC
+    // (add_fluxes.f90) leaves chipf(1) unset, so the reference carries a zero
+    // there; PARVMEC added this rule in 2016.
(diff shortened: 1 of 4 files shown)


jurasic-pf:
good idea, it should extrapolate nicely. assuming constant iota, we can say that chi is proportional to the area enclosed between the axis and rho, so proportional to rho (called r below)

s=r**2
dA/dr propto r
and dA/ds=dA/dr dr/ds with r=s^1/2 thats r'=s^-1/2
so dA/ds at 0 is constant

PR #781, opened 2026-09-02
Symmetrize the cth_like helical coil to the file's own precision and regenerate its mgrid and free-boundary references

CharlesCNorton:
The helical filament in `coils.cth_like` is written with six significant digits, so its 200 vertices are stellarator-symmetric only to 6e-6: under the field-period rotation vertex k maps to k+120 (the filament makes two toroidal turns with an l = 2 winding) and under the mirror to -k, and with those maps the vertex set is invariant to 5.9e-6. The mgrid box comes within a few centimetres of the inner leg, where that rounding reaches 6.9e-5 of the largest field component on the mid-period plane, the only stored plane that is not a mirror copy of another (measured in #761).

Each vertex is replaced by the average of its ten symmetry images, which moves no vertex by more than 3.6e-6 and makes the set invariant to 5e-16, and the file is written with twelve digits. `mgrid_cth_like.nc` is regenerated from it with the `makegrid` binary of this tree (the same binary reproduces the checked-in file from the old coils to zero relative difference). The mid-period residual of the response table drops from 6.9e-5 to 1.8e-11; the field moves by up to 8.9e-4 of its maximum next to the coil, and the `cth_like_free_bdy` equilibrium by 7e-10 in energy, 6e-7 in iota and 5e-7 in the boundary coefficients.

The `cth_like_free_bdy` references were regenerated with educational_VMEC on the new mgrid from `input.cth_like_free_bdy` as checked in (same dump flags and iterations, 490 iterations as before). The same binary run on the old mgrid reproduces the checked-in dumps to 5e-8 or better, most to 1e-11, so it is the reference generator. Of the 158 dumps, the 71 that changed by more than 1e-12 are replaced; the 87 that do not depend on the vacuum field (the fixaray basis tables, the initial profiles, the pre-vacuum iterations) keep their checked-in versions. `wout_cth_like_free_bdy.nc` under `vmecpp_large_cpp_tests` takes the fresh values into the checked-in file's layout, and the copy under `vmecpp/test_data`, which is a VMEC++ output used by `tests/test_init.py`, is regenerated by VMEC++ on the new mgrid. `jxbout`, `threed1` and `mercier` outputs come from the same run.

All targets under `//vmecpp/...` and `//vmecpp_large_cpp_tests/...` pass, as do `tests/test_free_boundary.py` and `tests/test_init.py`.


jurasic-pf (approved):
Thanks, annoying that it wasnt symmetrical in tge first place

PR #790, opened 2026-09-03
Write and read variable-length string datasets through character pointers

### LINKED ISSUE #780: OutputQuantities::Save segfaults after a free-boundary run
`OutputQuantities::Save` segfaults for a free-boundary run. The run itself completes; the crash is in `WOutFileContents::WriteTo` while writing a `std::vector<std::string>` dataset:

```
#0  ... in ?? ()
#1  H5T__vlen_mem_str_getlen ()
#2  H5T__conv_vlen ()
#3  H5T_convert ()
#4  H5D__scatgath_write ()
...
#11 H5::DataSet::write(...)
#12 hdf5_io::WriteH5Dataset(std::vector<std::string> const&, std::string const&, H5::H5File&)
#13 vmecpp::WOutFileContents::WriteTo(H5::H5File&) const
#14 vmecpp::OutputQuantities::Save(std::filesystem::path const&) const
```

Reproduction, from `src/vmecpp/cpp/vmecpp/test_data`:

```python
import numpy as np
import vmecpp
from vmecpp.cpp import _vmecpp as c
indata = vmecpp.VmecInput.from_file("input.cth_like_free_bdy")
indata.lasym = True
indata.rbs = np.zeros_like(indata.rbc)
indata.zbc = np.zeros_like(indata.zbs)
indata.raxis_s = np.zeros_like(indata.raxis_c)
indata.zaxis_c = np.zeros_like(indata.zaxis_s)
oq = c.run(indata._to_cpp_vmecindata(), max_threads=1, verbose=c.OutputMode.SILENT)
print(oq.wout.niter)  # 490, the run completed
oq.save("out.h5")     # SIGSEGV
```

The fixed-boundary `input.cth_like_fixed_bdy` saves in both symmetry modes, and `input.cth_like_free_bdy` crashes in both. `curlabel` is the only `std::vector<std::string>` member of `WOutFileContents`, and it is empty for a fixed-boundary run, which has no mgrid to take coil group names from. Present on `main` at 91ab6b3.

The `lasym = True` lines above are not needed to reproduce it; they are kept so the trace matches the one quoted.

CharlesCNorton:
Fixes #780.

`WriteH5Dataset(const std::vector<std::string>&, ...)` passed `vs.data()` to a dataset of variable-length strings. That is a `const std::string*` where HDF5 expects an array of `char*`, so it read the leading bytes of each `std::string` as a pointer and took its length there. An empty vector dereferences nothing, which is why the only crash seen so far was on a free-boundary run: a fixed-boundary run has no mgrid, so no coil group names, so `curlabel` is empty. It now writes through an array of `c_str()` pointers.

`ReadH5Dataset` had the same defect on the way back, and dropped the buffers HDF5 allocates for each element. It now reads into `char*`, copies the characters into the strings, and reclaims.

The test covers the round trip for a non-empty vector, including a string past the small-string buffer so the characters live on the heap, and for an empty one. It segfaults against the previous write on `main` and passes with this change, in 0.2 s and without a VMEC run. `bazel test //util/... //vmecpp/vmec/output_quantities/...` is green.

### DIFF
--- a/src/vmecpp/cpp/util/hdf5_io/BUILD.bazel
+++ b/src/vmecpp/cpp/util/hdf5_io/BUILD.bazel
@@ -11,3 +11,13 @@ cc_library(
         "//third_party/hdf5",
     ],
 )
+
+cc_test(
+    name = "hdf5_io_test",
+    srcs = ["hdf5_io_test.cc"],
+    deps = [
+        ":hdf5_io",
+        "@googletest//:gtest_main",
+    ],
+    size = "small",
+)
--- a/src/vmecpp/cpp/util/hdf5_io/hdf5_io.cc
+++ b/src/vmecpp/cpp/util/hdf5_io/hdf5_io.cc
@@ -32,7 +32,14 @@ void hdf5_io::WriteH5Dataset(const std::vector<std::string>& vs,
   H5::DataSpace dataspace(/*rank=*/1, /*dims=*/&size);
   H5::StrType str_type(H5::PredType::C_S1, H5T_VARIABLE);
   H5::DataSet dataset = file.createDataSet(name, str_type, dataspace);
-  dataset.write(vs.data(), str_type);
+  // A variable-length string dataset is written from an array of pointers to
+  // the characters, not from the std::string objects themselves.
+  std::vector<const char*> pointers;
+  pointers.reserve(vs.size());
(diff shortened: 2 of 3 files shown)


jurasic-pf:
Thanks for the fix!

PR #791, opened 2026-09-03
Fix the reason strings for wout status codes 0 and 5

philippbogdan:
`VmecWOut.reason` described two status codes wrongly.

Code 0: `output_quantities.cc` maps `SUCCESSFUL_TERMINATION` (11) to 0 when writing the wout, as educational_VMEC does, so a converged run reports 0. The string said "no fatal error but convergence was not reached". It now says normal termination, and names the one case in which 0 is not a converged run (`return_outputs_even_if_not_converged`).

Code 5: `VmecStatus::UNRECOVERABLE_ERROR`. The string said "input parsing error", the meaning of 5 in the Fortran code. It now follows the enum's comment.

Seen on a run of 312 QUASR free-boundary cases where every converged wout carried `reason == "no fatal error but convergence was not reached"`.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -860,11 +860,11 @@ class VmecWOut(BaseModelWithNumpy):
     @property
     def reason(self) -> str:
         return {
-            0: "no fatal error but convergence was not reached",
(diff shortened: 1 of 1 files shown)


jurasic-pf (approved):
Thanks! Nicely spotted

PR #794, opened 2026-09-04
Return netCDF failures from WriteMakegridNetCDFFile instead of aborting

CharlesCNorton:
`WriteMakegridNetCDFFile` returns `absl::Status` and reports its two argument checks that way since #722, while every netCDF call in the same function is a `CHECK_EQ` against `NC_NOERR`. An output path whose directory does not exist ends the process at the first of them:

```
F0000 00:00:1788529526.383789 1532117 makegrid_lib.cc:762] Check failed:
nc_create(makegrid_filename.c_str(), NC_CLOBBER, &ncid) == NC_NOERR (2 vs. 0)
*** Check failure stack trace: ***
```

The same holds for the definition and write calls below it, which is where a full disk lands.

The 47 checks become status returns carrying `nc_strerror` and the file name. `nc_create` and `nc_close` are written out, one having no open file to close yet and the other being the close itself; the rest go through a `VMECPP_RETURN_IF_NETCDF_ERROR` macro that closes the file and is undefined again after the function.

`CheckWriteMakegridNetCDFFileReportsUnwritablePath` writes into a directory that does not exist and requires `kInternal` and the file name in the message. It aborts against the previous code and passes with this change.

`//vmecpp/common/makegrid_lib:makegrid_lib_test` passes 19 tests, including the write round trip and the argument rejections that #722 added.

Independent of #793, which changes a different function in this file.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib.cc
+++ b/src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib.cc
@@ -735,6 +735,19 @@ absl::StatusOr<MakegridCachedVectorPotential> ComputeVectorPotentialCache(
   return response_table_a;
 }  // ComputeVectorPotentialCache
 
+// Report a netCDF failure through the return value of the enclosing function,
+// carrying the library's own message, and close the file being written.
+#define VMECPP_RETURN_IF_NETCDF_ERROR(nc_call)                               \
+  do {                                                                       \
+    const int nc_status = (nc_call);                                         \
+    if (nc_status != NC_NOERR) {                                             \
+      nc_close(ncid);                                                        \
+      return absl::InternalError(absl::StrFormat("could not write '%s': %s", \
+                                                 makegrid_filename,          \
+                                                 nc_strerror(nc_status)));   \
+    }                                                                        \
+  } while (0)
+
 absl::Status WriteMakegridNetCDFFile(
     const std::string& makegrid_filename,
     const MakegridParameters& makegrid_parameters,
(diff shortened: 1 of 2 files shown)


src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib.cc
```diff
@@ -733,6 +733,19 @@ absl::StatusOr<MakegridCachedVectorPotential> ComputeVectorPotentialCache(
...
+#define VMECPP_RETURN_IF_NETCDF_ERROR(nc_call)                               \
+  do {                                                                       \
+    const int nc_status = (nc_call);                                         \
+    if (nc_status != NC_NOERR) {                                             \
+      nc_close(ncid);                                                        \
+      return absl::InternalError(absl::StrFormat("could not write '%s': %s", \
+                                                 makegrid_filename,          \
+                                                 nc_strerror(nc_status)));   \
+    }                                                                        \
+  } while (0)
```
jurasic-pf, line 749: cant you just do local scoping with {} ? why do we need the do while?

jurasic-pf (approved):
good pr

src/vmecpp/cpp/vmecpp/common/makegrid_lib/makegrid_lib.cc
CharlesCNorton: Bare braces would make `MACRO(x);` expand to a compound statement plus a null statement, which breaks an `if (...) MACRO(x); else ...`; the do/while(0) is what lets it be used as an ordinary statement anywhere. Every call site here is standalone today, so braces would compile, but the macro would stop being safe to move.

PR #764, opened 2026-09-01
Set the comparison tolerances from the measured deviation

CharlesCNorton:
Several comparison tolerances in the C++ tests sit within round-off of the deviation they bound, so they fail intermittently on the CI runners while passing locally. They take down pull requests that cannot reach the code they test.

Two have done so this week. `Threed1ShafranovIntegralsTest/cma` reads 5.114e-11 and 5.462e-11 against its 5.0e-11 bound on the asan and opt runners, and measures 4.08e-11 locally; it failed #740 and #816, neither of which touches the Shafranov integrals. `ExtrapolateBSubsSTest/cth_like_fixed_bdy_nzeta_37` carried 5.0e-15 against a deviation measuring 3.3e-15 locally and 5.2e-15 on the CI compiler.

Every tolerance in the suite reaches `testing::IsCloseRelAbs`, so recording the worst `|rel_abs_error|` per call site there measures all of them at once. Over the full suite that is 2121 distinct (file, line, tolerance) sites, measured four times: the `opt`, `asan` and `ubsan` builds this repository tests in CI, and one built with `-march=native`.

The three CI builds agree bit-for-bit on all 2121 sites. `-march=native` moves 1461 of them, by up to a factor of four, which is the size of the shift a different instruction set or compiler produces. Twenty-one round-off-bound sites sit below a factor of 1.5 over those four builds, which is less than that shift.

Those are raised to at least five times the worst observed deviation, rounded to a round number:

| test | from | to | worst observed |
|---|---|---|---|
| `RecomputeMagneticAxisToFixJacobianSignTest` | 1.0e-15 | 5.0e-15 | 9.66e-16 |
| `CovariantMagneticFieldTest` | 2.0e-15 | 1.0e-14 | 1.96e-15 |
| `VolumeTest` | 1.0e-15 | 5.0e-15 | 9.39e-16 |
| `EvolveTest` | 1.0e-15 | 5.0e-15 | 8.65e-16 |
| `EvolveTest` | 2.0e-14 | 1.0e-13 | 1.36e-14 |
| `InverseFourierTransformGeometryTest` | 2.0e-15 | 1.0e-14 | 1.41e-15 |
| `JacobianTest` | 5.0e-15 | 2.0e-14 | 3.44e-15 |
| `JacobianTest` | 2.0e-14 | 1.0e-13 | 1.37e-14 |
| `FourierGeometryToStartWithTest` | 1.0e-14 | 5.0e-14 | 7.10e-15 |
[...]

PR #795, opened 2026-09-04
Carry the Jacobian sign in the Mercier volume derivative

CharlesCNorton:
**Change** - `ComputeMercierStability` multiplies the volume derivative of the first Mercier table by `sign_of_jacobian`, as it already does for WELL at the same site.

**Cause** - the volume derivative was the plain half-grid average of `vp_real`. `mercier.f90:191` forms it as

```fortran
sqs = p5*(vp_real(i) + vp_real(i+1))*sign_jac
```

and then writes `shear(i)/sqs` as SHEAR, `sqs` itself as VP, `ip(i)/sqs` as ITOR' and `presp(i)/sqs` as PRES'. The Jacobian sign is therefore carried by all four columns, and with the usual negative Jacobian, which is every bundled case, they came out negated.

**Evidence** - ratio to the `mercier.<case>` file educational_VMEC writes, before the change, to the five significant figures the file carries, on every case:

| column | ratio before |
| --- | --- |
| SHEAR | -0.999987 |
| VP | -1.000005 |
| ITOR' | -0.999999 |
| PRES' | -0.999996 |
| WELL | 0.999978 |
| IOTA | 0.999974 |

S, PHI, PRES and ITOR agree as well. Comparing every column of both Mercier tables against that file across the bundled cases, nine go from four differing columns to all fifteen agreeing: `solovev`, `solovev_no_axis`, `solovev_analytical`, `cth_like_fixed_bdy`, `cth_like_fixed_bdy_iota`, `cth_like_fixed_bdy_nzeta_37`, `cth_like_fixed_bdy_spline_pressure`, `cth_like_free_bdy` and `li383_low_res`. What still differs elsewhere does not come from this quantity: `cma` carries no current, so its ITOR and ITOR' are round-off around 1e-10; the asymmetric cases differ by the `tmult` normalisation; and `up_down_asymmetric_tokamak` sits in a different poloidal gauge.

**Tests** - `CheckMercierStability` reproduced the same unsigned average and asserted the first table against it, so the test and the implementation agreed with each other and both differed from the reference. The test now applies `sign_of_jacobian` where it already reads it for the WELL comparison.

**Scope** - the second table is unaffected, since `Dshear`, `Dcurr`, `Dwell` and `Dgeod` are built from the intermediates and not from this quantity. Nothing outside the output stage reads these fields.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -3187,9 +3187,12 @@ vmecpp::MercierFileContents vmecpp::ComputeMercierStability(
     // S
     mercier.s[jF] = mercier_intermediate.s[jF];
 
+    // mercier.f90 forms this as sqs = 0.5*(vp_real(i) + vp_real(i+1))*sign_jac
+    // and divides SHEAR, ITOR' and PRES' by it, so the Jacobian sign belongs
+    // here rather than only on WELL.
     const double vp_full = (mercier_intermediate.vp_real[jHo] +
                             mercier_intermediate.vp_real[jHi]) /
-                           2.0;
+                           2.0 * vmec_internal_results.sign_of_jacobian;
     if (vp_full == 0.0) {
       // skip this surface
       continue;
--- a/src/vmecpp/cpp/vmecpp_large_cpp_tests/vmec/output_quantities/output_quantities_test.cc
+++ b/src/vmecpp/cpp/vmecpp_large_cpp_tests/vmec/output_quantities/output_quantities_test.cc
@@ -671,7 +671,7 @@ TEST_P(MercierStabilityTest, CheckMercierStability) {
 
     const double vp_full = (static_cast<double>(mercier["vp_real"][jHo]) +
                             static_cast<double>(mercier["vp_real"][jHi])) /
-                           2.0;
+                           2.0 * sign_of_jacobian;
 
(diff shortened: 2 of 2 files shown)


CharlesCNorton:
The one job that failed here is `Install and run tests (macos-14, true, 3.14)`; the other fifteen were cancelled by fail-fast. Its failure is `test_pybind_vmec.py::test_output_quantities` on `wout.jcuru[50]`, the extrapolated edge value `2*jcuru(ns-1) - jcuru(ns-2)` of `cma`, at 1.03e-6 against the 1e-6 tolerance (`4.814665522566e+01` against the reference `4.814670572801e+01`). The Ubuntu 3.12 job in the same run passed the test, and the same macOS job passed on #792, #793 and #794 at this base.

`jcuru` comes from `bsubu` and `bsubv` through the radial force balance and reads none of the four Mercier columns this change touches; the only Mercier quantities that reach `wout` are the second-table `DMerc`, `DShear`, `DCurr`, `DWell` and `DGeod`, which are built from the intermediates and are unchanged. A Linux build of this branch puts that edge value 2.1e-7 from the reference. The margin on that assertion is the kind #764 measures.

I cannot rerun the workflow from a fork.

CharlesCNorton:
Checked independently against the `mercier.<case>` text file educational_VMEC writes, rather than the dumps: on main the SHEAR, VP, ITOR' and PRES' columns come out with the opposite sign for `cth_like_fixed_bdy` and `solovev`, and on this branch all ten columns agree to the five digits the file prints.

Issue #775, opened 2026-09-02
Antisymmetric part of a free-boundary solution differs from PARVMEC by 10%


CharlesCNorton:
The antisymmetric part of a free-boundary solution differs from PARVMEC by about 10% of its own size, while the symmetric part agrees.

Measured on `test_data/cth_like_free_bdy_asym.json` (stellarator-symmetric boundary and profiles, `mgrid_cth_like_asym.nc` as the external field), run through VMEC++ at `main` with #771 applied, through educational_VMEC, and through PARVMEC (Stellarator-Tools, serial, `version_ = 10.0`) from the same namelist. Max abs difference over the max abs of the PARVMEC array:

| quantity | VMEC++ vs PARVMEC | educational_VMEC vs PARVMEC |
|---|---|---|
| rmnc | 5.0e-4 | 5.2e-4 |
| zmns | 1.7e-3 | 1.7e-3 |
| bmnc | 5.5e-4 | 5.7e-4 |
| wb | 2.8e-5 | 3.1e-5 |
| rmns | 4.0e-2 | 3.7e-2 |
| zmnc | 1.5e-1 | 1.3e-1 |
| lmnc | 6.0e-2 | 5.4e-2 |
| gmns | 1.0e-1 | 9.5e-2 |
| bmns | 1.1e-1 | 1.0e-1 |

VMEC++ and educational_VMEC take 248 iterations, PARVMEC 256. The two 8.52-lineage codes agree with each other to the level of their symmetric parts, so the difference sits in what the lineages do with the antisymmetric external field in Nestor. With the symmetric `mgrid_cth_like.nc` and `lasym = true`, all three agree to 1e-7, so the symmetric path of Nestor is not involved.

The `LasymFreeBoundaryMatchesEducationalVmec` test checks seven scalars (aspect, volume, Rmajor, Aminor, beta, rbtor, volavgB) to 1e-4 against educational_VMEC and only requires the antisymmetric arrays to be non-zero, so the suite cannot see this. Which side is right is open; a comparison of `bsqvac` from the two Nestor implementations on a fixed asymmetric boundary would settle it. PARVMEC's `Sources/NESTOR_vacuum` received a series of recurrence and indexing fixes between 2026-05 and 2026-07.

CharlesCNorton:
Two attempts at isolating the vacuum solve, for whoever picks this up:

[...]

PR #796, opened 2026-09-04
Fold the sin-source vacuum kernel about theta -> -theta for lasym

CharlesCNorton:
`LaplaceSolver::PerformPoloidalFourierTransforms` projects the source-side kernel onto the reduced poloidal range `[0, pi]`. For `lasym` that kernel is evaluated on the full range, so each block has to fold it about `theta -> -theta` with the parity its projection needs. The two cos-source blocks do that. The two sin-source blocks did not: `amat_sin_sin` took `actemp` and `astemp` on the reduced range as they stand, and `amat_sin_cos` took the same arrays with cos weights. `fouri.f90:127-141` integrates all four blocks over the full range with plain weights.

On a stellarator-symmetric surface `actemp` is odd and `astemp` even in the source angle, so the full-range integral behind `amat_sin_cos` vanishes while its half-range sum does not. That block maps the cos potential into the sin equation, and the antisymmetric part of the vacuum field therefore leaked into the symmetric potential at first order. `amat_sin_sin` was exact only while the surface stayed symmetric. Nothing in the tree could see either: the degenerate `lasym` tests carry no cos potential, and no asymmetric free-boundary case has a reference.

Measured on `cth_like_free_bdy_asym`, whose mgrid is asymmetric at 2.7e-4 of the field, against its mirror image (plane `k -> (kp - k) % kp`, `Z -> -Z`, `B_R -> -B_R`, the relation the symmetric `mgrid_cth_like.nc` satisfies to 2e-15). At the first vacuum solve the surface is symmetric to 1e-10, so the sin half of `potvac` must be identical for the symmetric, asymmetric and mirrored fields and the cos half must change sign:

| | before | after |
| --- | --- | --- |
| `\|sin(asym) - sin(sym)\|` | 8.1e-3 | 1.1e-15 |
| `\|sin(asym) - sin(mirror)\|` | 1.6e-2 | 2.4e-15 |
| `\|cos(asym) + cos(mirror)\|` | 9.9e-14 | 9.9e-14 |

At convergence (`ftol` 1e-8, 248 iterations both ways), the original and mirrored equilibria:

| | before | after |
| --- | --- | --- |
| `volume`, `aspect`, `wb` | 8.6e-6, 2.1e-5, 6.1e-6 | 1.0e-13, 7.5e-15, 1.2e-15 |
| `rmax_surf`, `rmin_surf`, `zmax_surf` | 1.2e-4, 1.3e-4, 4.8e-4 | 1.7e-13, 4.5e-14, 8.4e-13 |
| `rmns`, `zmnc`, `lmnc`, `bmns` (`\|a+b\|/max\|a\|`) | 5.4e-3, 2.2e-2, 1.2e-2, 1.7e-2 | 1.6e-11, 6.4e-11, 3.4e-11, 5.0e-11 |

educational_VMEC and PARVMEC sit at 1e-11 on those arrays and 1e-13 on those scalars.

Against the other two codes at the same tolerance, as `max|a-b|/max|reference|`:

| | VMEC++ vs edu before | VMEC++ vs edu after | VMEC++ vs PARVMEC after | edu vs PARVMEC |
| --- | --- | --- | --- | --- |
| `rmns` | 2.7e-3 | 4.3e-6 | 3.72e-2 | 3.72e-2 |
| `zmnc` | 1.0e-2 | 1.4e-5 | 1.196e-1 | 1.196e-1 |
| `lmnc` | 5.8e-3 | 8.4e-6 | 5.11e-2 | 5.11e-2 |

So the antisymmetric part now agrees with educational_VMEC to 1e-5, and what remains of the difference #775 measured against PARVMEC is shared by both 8.52-lineage codes to four digits.

The `lasym = false` path is untouched: `cth_like_free_bdy` against the checked-in Fortran wout stays at 1.9e-12. The fixed-boundary `lasym` mirror pair (`cth_like_fixed_bdy_asym` with `rbs`, `zbc`, `raxis_s`, `zaxis_c` negated) is exact to 1e-14 before and after, which confines the change to the vacuum solver.

`LasymFreeBoundaryIsMirrorCovariant` reads `mgrid_cth_like_asym.nc` into a response table, mirrors it in memory, runs both and requires the gauge-invariant scalars to agree to 1e-10, the symmetric arrays to agree and the antisymmetric arrays to change sign to 1e-9 of their scale. Against `main` every one of its fourteen expectations fails.

`laplace_solver_test` (9), `singular_integrals_test` (4), `vmec_in_memory_mgrid_test` (3 plus the new case) and `vmec_test` (14, including both `lasym` free-boundary cases) pass.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/free_boundary/laplace_solver/laplace_solver.cc
+++ b/src/vmecpp/cpp/vmecpp/free_boundary/laplace_solver/laplace_solver.cc
@@ -422,19 +422,40 @@ void LaplaceSolver::PerformPoloidalFourierTransforms() {
         astemp_l[l] = astemp[base_idx + l];
       }
 
-      Eigen::VectorXd result_ss = sinmui_scaled.transpose() * actemp_l -
-                                  cosmui_scaled.transpose() * astemp_l;
-
-      for (int m = 0; m < mf + 1; ++m) {
-        const int idx_amat = (all_n * (mf + 1) + m) * mnpd + mn;
-        amat_sin_sin[idx_amat] = result_ss[m];
-      }
-
-      if (s_.lasym) {
-        Eigen::VectorXd result_sc = cosmui_scaled.transpose() * actemp_l +
-                                    sinmui_scaled.transpose() * astemp_l;
+      if (!s_.lasym) {
+        Eigen::VectorXd result_ss = sinmui_scaled.transpose() * actemp_l -
+                                    cosmui_scaled.transpose() * astemp_l;
         for (int m = 0; m < mf + 1; ++m) {
           const int idx_amat = (all_n * (mf + 1) + m) * mnpd + mn;
+          amat_sin_sin[idx_amat] = result_ss[m];
+        }
+      } else {
+        // The sin-source kernel grpmn_sin is evaluated on the full theta range
+        // as well, so its poloidal projection folds about theta -> -theta the
+        // same way as the cos-source kernel below: the sin-projection takes
+        // the part of actemp that is odd under the reflection and the part of
+        // astemp that is even, the cos-projection the other two parts.
+        Eigen::VectorXd ac_odd(s_.nThetaReduced), as_even(s_.nThetaReduced);
+        Eigen::VectorXd ac_even(s_.nThetaReduced), as_odd(s_.nThetaReduced);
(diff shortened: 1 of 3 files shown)


jurasic-pf:
Well done this is a tricky bug

PR #798, opened 2026-09-05
Reject the unimplemented sum_cossq profile names at input validation

CharlesCNorton:
`sum_cossq_s`, `sum_cossq_sqrts` and `sum_cossq_s_free` are registered in `AllProfileParameterizations()` with `allowedForCurr = true`, so `CheckProfile` accepts them as current profiles. `evalProfileFunction` has no case for any of them: they fall through to `default:`, which writes one line to stderr and returns `0.0` for every `s`.

That is the failure the validation added in #719 was meant to prevent, per its own comment: an unrecognized name "reaches the solver as a zero profile, which converges to a silently wrong equilibrium."

With `ncurr = 1` the consequence is that the prescribed net toroidal current is dropped. `evalCurrProfile(1.0)` returns zero, so the `|edgeCurrent| > eps * curtor` guard in `evalRadialProfiles` fails, `Itor` stays zero, and `currH` is zero on every surface. Running `cth_like_fixed_bdy.json` (`ncurr = 1`, `curtor = 4.322908e+04`) unchanged and then with `pcurr_type = "sum_cossq_s"`:

| | `two_power` (as shipped) | `sum_cossq_s` |
|---|---|---|
| `ier_flag` | 0 | 0 |
| `fsqr` | 9.348e-07 | 9.918e-07 |
| `ctor` | +4.322913e+04 A | -2.554651e-11 A |
| max abs `jcurv` | 1.201671e+04 | 2.284859e-10 |

The run terminates normally and writes a `wout` that reports convergence to a zero-current equilibrium while the input asked for 43.2 kA. The only signal is the stderr line, which carries no error flag and is easily lost through the Python API.

The parameterization table now records whether a name has an evaluator, and `CheckProfile` rejects one that does not, so these inputs fail up front instead of solving the wrong problem. The three names stay in the table and keep their enum values, so implementing them later is a local change. The README already listed them as not implemented; it now also says they are rejected rather than evaluated as zero.

The new test fails on all three names without the validation change and passes with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,6 +352,8 @@ VMEC++:
    * `sum_cossq_s`
    * `sum_cossq_sqrts`
    * `sum_cossq_s_free`
+
+   The `sum_cossq_*` family has no evaluator at all, so those three names are rejected at input validation instead of evaluating to a zero profile.
 - some (rarely used) free-boundary-related output quantities are not implemented yet:
    * `curlabel` - declared but not populated yet
    * `potvac` - declared but not populated yet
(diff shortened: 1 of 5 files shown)


CharlesCNorton:
Superseded by #832, which implements the three profiles instead of rejecting them.

PR #799, opened 2026-09-06
Recover cold initial-Jacobian failures through the coarse bootstrap

philippbogdan:
A cold run can reject its first radial grid with a bad Jacobian even though a three-surface bootstrap reaches the requested equilibrium. Both the inner solver and grid driver returned before the existing fallback could execute. Allow that initial failure to reach the driver, restore the input axis and radial state, and retry once at three surfaces before the original grid sequence. Correct the logger's stage index for the inserted stage. The public single-grid solver keeps its existing interface and error handling.

On a frozen set of 53 historical QUASR initial-Jacobian failures, the baseline converges on 0 cases and this change on the same 15 cases in each of three repetitions: 5 of 12 discovery cases and 10 of 41 held-out cases. Both use identical boundary coefficients, profiles, field tables, final resolution and `1e-9` tolerance. This is an initialization-robustness result on a selected stress set, not a population failure-rate or physical-accuracy claim.

The PR includes the frozen source hashes, portable paired benchmark, measured ledgers, and regression tests. The numerical tests compare the complete recovered geometry and force history with an explicit cold-three solve. A regenerated educational_VMEC reference checks geometry, lambda, iota, pressure and flux. The corresponding Fortran fix is [educational_VMEC #29](https://github.com/jonathanschilling/educational_VMEC/pull/29).

Validation details and physical-diagnostic limits are recorded in `benchmarks/initial_jacobian_recovery/README.md`. Native VMEC tests pass on macOS arm64; the paired Python regression run found the same six failures on baseline and the recovery implementation, with 201 passes, two skips and 36 deselections. That comparison preceded the final interface-preserving wrapper refactor; native tests, Python reference checks and independent API probes were repeated on the final build. Independent API checks cover fixed/free boundaries, finite pressure/current, one/four threads, all output modes, hot starts, later-stage failures and insufficient budgets.


philippbogdan:
The conda job has no git-lfs, so it read the reference wout as a pointer file; the tsan run of vmec_test exceeded 900 s.
Follow-up commit deselects the new Python test there, moves the LFS rule into src/vmecpp/cpp/.gitattributes, and trims the C++ test sweep.

PR #800, opened 2026-09-06
Scale the sum_atan arctangent terms by 2/pi

CharlesCNorton:
`evalSumAtan` summed the arctangent terms without the `2/pi` factor that Fortran VMEC applies in `piota` and `pcurr` for `sum_atan` (profile_functions.f), so each term reached `pi/2` times its amplitude instead of the amplitude, and the profile jumped at `s = 1`, where the closed-form branch sums the bare amplitudes. An iota or current profile of this type therefore converged to a different equilibrium than the same input in Fortran VMEC.

The factor is applied to the sum of the terms. The new test pins a unit-exponent term at half its amplitude at `s = 1/2`, checks a two-term profile against the closed form, and requires continuity with the `s >= 1` branch; it fails on the previous evaluator.

`bazel test --config=opt //vmecpp/...` passes.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
@@ -567,6 +567,8 @@ double RadialProfiles::evalGaussTrunc(const Eigen::VectorXd& coeffs, double x,
 }
 
 double RadialProfiles::evalSumAtan(const Eigen::VectorXd& coeffs, double x) {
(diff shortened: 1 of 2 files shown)


jurasic-pf (approved):

CharlesCNorton:
The same factor is missing from the rotational-transform profile. With `PIOTA_TYPE = 'sum_atan'` and `AI = 0.0 1.0 1.0 1.0 1.0` on a cth-like fixed-boundary case, main converges in 559 iterations against educational_VMEC's 469 and lands 1.6 relative away in `jcuru` and `currumnc`; on this branch the iteration count matches at 469 and every wout field agrees below 1e-09.

PR #801, opened 2026-09-06
Reject an mgrid whose number of field periods differs from the input

CharlesCNorton:
`MGridProvider` reads `nfp` from the mgrid file, or takes it from the in-memory response table, and `Vmec::run` never compared it with `VmecINDATA::nfp`. The vacuum field was then laid over the input's field period whatever the table was computed for: `cth_like_free_bdy` run against `mgrid_cth_like.nc` with its `nfp` header changed from 5 to 10 converges to the same equilibrium as with the file as shipped. LIBSTELL's `read_mgrid` stops on that mismatch.

`Vmec::run` now returns `kInvalidArgument` naming both values, beside the existing `nzeta` check. The new test builds the cth_like response table with one field period more than the input and requires the rejection; on the previous code that run converges.

`bazel test --config=opt //vmecpp/vmec/vmec/...` and `tests/test_free_boundary.py` pass.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec.cc
@@ -236,6 +236,12 @@ absl::StatusOr<bool> Vmec::run(const VmecCheckpoint& checkpoint,
           "are consistent.",
           mgrid_.numPhi, indata_.nzeta));
     }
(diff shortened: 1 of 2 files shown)


jurasic-pf (approved):

jurasic-pf:
Thanks! Somebody actually noticed this internally a few weeks ago too, good to have it fixed

PR #805, opened 2026-09-06
Keep the full vacuum update every iteration until the free-boundary residuals settle

CharlesCNorton:
VMEC 8.52 forces a full Nestor update in every iteration while `ivac <= 2`, and `ivac` only grows past 2 in an iteration whose R and Z force residuals are below `1e-3` (`funct3d.f90`). `VacuumPressureState` capped at `kActive`, so the increment stopped as soon as the vacuum pressure was switched on and the `nvacskip` cadence started one iteration later, while the free-boundary transient was still one to three orders of magnitude above that threshold. Every Nestor call in that window reused the Green's function matrices instead of recomputing them.

`kSettled` marks the state after the residuals have settled, and the full update is forced while the state is at most `kActive`. The remaining comparisons against `kInitialized` and `kActive` become `>=`, so the new state keeps the vacuum pressure in the force balance, in the free-boundary edge outputs and in `DELBSQ`. The multigrid seeding of the vacuum state is dropped: the state carries over into the continuation stage by itself, as in Fortran, and the first iteration of the new stage still runs the free-boundary block through the widened gate. `cth_like_free_bdy_multigrid` converges in the same 321 iterations.

Largest relative deviation in `rmnc` from educational_VMEC on the single-grid free-boundary Solovev case (`ns = 16`, `ftol = 1e-12`), both codes reading the same INDATA file, with the iteration counts in brackets:

| nvacskip | before | after | Fortran |
|---|---|---|---|
| 1  | 1.3e-12 [514] | 1.3e-12 [514] | [514] |
| 3  | 9.1e-08 [519] | 5.8e-12 [521] | [521] |
| 6  | 1.1e-06 [525] | 4.8e-12 [529] | [529] |
| 12 | 2.2e-06 [530] | 8.9e-13 [526] | [526] |
| 24 | 4.2e-06 [552] | 1.6e-12 [538] | [538] |

At `nvacskip = 24` the deviation before the change reaches `2.0e-05` in `lmns` and `1.7e-05` in `jcurv`. `nvacskip = 1` is the control, where the two cadences coincide. `cth_like_free_bdy` is unchanged at every `nvacskip` tested (1, 9, 24), since its vacuum pressure switches on when the residuals are already below the threshold; it stays at `1.8e-13` in `rmnc` and reproduces the Fortran iteration count.

The new test solves the case twice, with `nvacskip` 1 and 24, and requires the force-residual history to agree exactly up to the evaluation at which the residuals settle. It fails on the previous cadence.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/util/util.h
+++ b/src/vmecpp/cpp/vmecpp/common/util/util.h
@@ -204,9 +204,13 @@ enum class VacuumPressureState : std::int8_t {
   // process of reducing rCon0,zCon0 *= 0.9;
   kInitialized = 1,
 
-  // vacuum pressure turned on
-  // in the process of reducing rCon0,zCon0 *= 0.9;
-  kActive = 2
+  // vacuum pressure turned on, R and Z force residuals still above 1e-3
+  // full vacuum update in every iteration; ivac == 2 in VMEC 8.52
+  kActive = 2,
+
+  // vacuum pressure turned on, R and Z force residuals below 1e-3
+  // vacuum update every nvacskip iterations; ivac > 2 in VMEC 8.52
+  kSettled = 3
 };
 
 int VmecStatusCode(const VmecStatus vmec_status);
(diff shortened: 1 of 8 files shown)


jurasic-pf:
wow! very nice

jurasic-pf:
@copilot resolve the merge conflicts in this pull request

jurasic-pf (approved):

jurasic-pf:
resolve conflicts please, then we can merge

jurasic-pf (approved):

CharlesCNorton:

CharlesCNorton:
fixing clang

CharlesCNorton:
Rebased onto current main. The clang-tidy warnings were the five THREED1_* constants of VmecCheckpoint, which predate this branch and only fell inside the reviewed region because #763 moved them to line 174; my hunk now sits at 204.

PR #806, opened 2026-09-06
Recover the full-grid lambda from the half-grid array when a wout file lacks it

CharlesCNorton:
`VmecWOut.from_wout_file` filled `lmns_full`, and `lmnc_full` for an asymmetric run, with zeros whenever the file did not carry them. Fortran VMEC stores lambda on the half grid only, so every wout written outside VMEC++ took that branch, and `run(..., restart_from=...)` then seeded the restart with zero lambda. The restart re-converged the lambda sector from scratch: `cth_like_fixed_bdy` took 122 iterations instead of 2, `solovev` 287 instead of 2, and the two restarts settled on states 4.6e-4 apart in `rmnc`.

The full-grid array is now recovered by inverting the interpolation VMEC applies when it writes the half-grid one: the average over the two adjacent full-grid surfaces for even m, weighted by sqrt(s) for odd m, with the innermost value of the m <= 1 modes taken from the surface outside. The recursion runs outward from the axis, where the m > 0 modes vanish. The axis value of the m = 0 modes is not represented on the half grid; it is extrapolated from the two innermost surfaces for the stellarator-symmetric half, matching the writer, and is left at zero for the antisymmetric half, which the writer does not extrapolate. A file that already carries the full-grid arrays keeps them, and one missing `lmns`, `xm` or `phipf` still falls back to zeros.

Measured against Fortran wout files for the same equilibrium:

| reference | recovered `lmns_full` vs the VMEC++ solve | hot restart |
|---|---|---|
| educational_VMEC `cth_like_fixed_bdy` | 1.2e-13 | 2 iterations (122 before) |
| educational_VMEC `solovev` | 8.1e-12 | 2 iterations (287 before) |
| PARVMEC 10.0 `cth_like_fixed_bdy` | 1.4e-06 | |

The PARVMEC figure is that run's own deviation from VMEC++: its half-grid `lmns` differs by the same 1.4e-06, so the recovery adds nothing.

Three tests cover it, all failing on the zero fallback: the array recovered from the shipped `wout_cth_like_fixed_bdy_spline_pressure.nc`, which stores lambda on the half grid only, against a fresh solve; both halves of an asymmetric equilibrium round-tripped through a wout with the full-grid arrays removed; and a hot restart from such a file, which must converge in under ten iterations.

`pytest` passes.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -803,42 +803,34 @@ def _lambda_on_full_grid(
     phipf: np.ndarray,
     extrapolate_axis: bool = True,
 ) -> np.ndarray:
-    """Invert the radial interpolation that writes lambda onto the half grid."""
-    n_surfaces = lambda_half.shape[1]
-    if n_surfaces < 2:
+    """Invert the radial interpolation that writes lambda onto the half grid.
+
+    Half-grid point j holds (w_out * lambda[j] + w_in * lambda[j - 1]) / 2, the
+    weights being 1 for even m and the sqrt(s) ratios of the odd-m scaling
+    otherwise; on the innermost interval the m <= 1 modes take both values from
+    surface 1 and the higher modes vanish on the axis.
+    """
+    ns = lambda_half.shape[1]
+    if ns < 2:
         return np.zeros_like(lambda_half)
 
-    sqrt_s_full = np.sqrt(np.arange(n_surfaces, dtype=float) / (n_surfaces - 1.0))
-    sqrt_s_full[-1] = 1.0
-    sqrt_s_half = np.sqrt(
-        (np.arange(n_surfaces - 1, dtype=float) + 0.5) / (n_surfaces - 1.0)
-    )
-    # odd-m interpolation weights of the outer and inner surface
-    outer = sqrt_s_half / sqrt_s_full[1:]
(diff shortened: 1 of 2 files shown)


jurasic-pf:
I don't think we want to do this in general. Restarting from a wout that is not produced by `vmecpp` has caveats, especially since I don't think the full lambda profile is exactly recoverable in general, there should be one node missing afaik?

CharlesCNorton:
One node is missing: ns-1 half-grid values against ns unknowns. The recursion closes it by regularity, since the m > 0 modes vanish at the axis, leaving the m = 0 axis value, which is extrapolated as the writer does; the recovered array matches a fresh solve to 1.2e-13. The current behaviour on a foreign wout is a zero fill rather than a refusal, which restarts in 122 iterations against 2 and lands 4.6e-4 away in rmnc.

jurasic-pf:
True, I didn't think of that. Sure, then let's merge :)

jurasic-pf (approved):
I wonder if we can simplify the code a bit, but functionally it looks good

CharlesCNorton:
Simplified: the even and odd parities share one recursion with per-mode weights, and both lambda halves go through one loop. The result is bit-identical to the previous version on the symmetric and asymmetric cases and within 1e-16 of the solver's own full-grid lambda.

PR #808, opened 2026-09-06
Normalize jsups by the same averaged volume derivative as the rest of jxbout

CharlesCNorton:
`jxbout.jsups3` is normalized by `1 / dVdsH[jH]`, the reciprocal of a single half-grid `V'`. Every other quantity in `ComputeJxBOutFileContents`, and `jxbforce.f90` for all of them including this one, uses `ovp = 2 / (V'(js) + V'(js+1)) / (4 pi^2)`, the reciprocal of the average over the two half-grid points adjacent to the full-grid surface the row belongs to. The radial current density therefore carries a factor `(V'(js) + V'(js+1)) / (2 V'(js))`, which is 1 only where `V'` happens to be locally flat.

The error is a smooth radial modulation, positive inside and negative outside where `dV/ds` turns over. Largest relative deviation of `jsups` from educational_VMEC, both codes reading the same INDATA file:

| case | before | after |
|---|---|---|
| `cth_like_fixed_bdy` | 1.15e-03 | 3.2e-10 |
| `solovev` | 5.17e-04 | 1.8e-06 |
| `cma` | 8.50e-04 | 5.5e-08 |

What remains is the cancellation in `bsubuv - bsubvu`, which is about `4e-3` of either term on this grid, so the 1e-12 agreement of the two derivatives themselves lands in the 1e-10 to 1e-6 range depending on how small `jsups` is for the case.

The reference comparison in `vmecpp_large_cpp_tests` already covered `jsups3`, but at a hard-coded `5.0e-3` attributed to that cancellation, which is seven orders looser than the deviation the cancellation actually produces and wide enough to admit the normalization error. It now uses the same per-case tolerance as the rest of the jxbout contents, `1e-5` for the cth-like cases and `2e-5` to `1e-4` for the others, and fails on the previous normalization for every case in the suite.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/output_quantities/output_quantities.cc
@@ -2984,7 +2984,11 @@ vmecpp::JxBOutFileContents vmecpp::ComputeJxBOutputFileContents(
     // The loop in jxbforce.f90:594 goes over js=2,ns1,
     // which means that the last half-grid point is not touched.
     for (int jH = 0; jH < vmec_internal_results.num_half - 1; ++jH) {
-      const double ovp = 1.0 / vmec_internal_results.dVdsH[jH] / dnorm1;
+      // row jH holds the full-grid surface jF = jH + 1
+      const double ovp = 2.0 /
+                         (vmec_internal_results.dVdsH[jH + 1] +
+                          vmec_internal_results.dVdsH[jH]) /
+                         dnorm1;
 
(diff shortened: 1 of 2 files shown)


CharlesCNorton:
The `cth_like_fixed_bdy_nzeta_37` case of `ExtrapolateBSubsSTest` carried a 5.0e-15 tolerance against a deviation that measures 3.3e-15 on main and 2.3e-15 here, and 5.2e-15 on the CI compiler, so any change to the translation unit flips it. It now uses the 2.0e-14 of the `cth_like_fixed_bdy` case beside it.

PR #809, opened 2026-09-06
Scale the enclosed toroidal flux by the bloating factor

CharlesCNorton:
`bloat` expands the profiles into a larger enclosed toroidal flux. Fortran VMEC applies it in two places: `readin` scales `phiedge` by it, and `pcurr`/`pmass` evaluate the profile at `min(|s * bloat|, 1)`. VMEC++ has the second and not the first, so the flux stays at `phiedge` and the equilibrium is not the one either Fortran implementation computes.

Against PARVMEC 10.0 on `cth_like_fixed_bdy`, which has `ncurr = 1` as `bloat` requires, reading the same INDATA file:

| bloat | edge phi, PARVMEC | edge phi, VMEC++ before | largest deviation before | after |
|---|---|---|---|---|
| 1.0 | -3.500e-02 | -3.500e-02 | 3.9e-08 | 3.9e-08 |
| 1.5 | -5.250e-02 | -3.500e-02 | 1.2e+00 in `betatotal` | 2.7e-07 |
| 0.5 | -1.750e-02 | -3.500e-02 | 2.9e+00 in `wb` | 3.7e-07 |

The remaining 1e-7 is PARVMEC's own deviation from VMEC++ on this case, the same at `bloat = 1` where nothing changes. The iteration counts also match exactly after the change: 649, 682 and 717 for the three values.

educational_VMEC cannot referee this input. It scales `phiedge` in `readin.f90` but its `profile_functions.f90` carries no `bloat` at all, so it implements the other half of the pair; a `bloat != 1` run there differs from PARVMEC as well.

The factor enters at the single place `phiedge` is consumed, the maximum toroidal flux in `RadialProfiles::computeMagneticFluxes`, rather than mutating the input as Fortran does, so a `VmecInput` still round-trips to the values the user supplied. The test runs the case at `ns = 9` for `bloat` 1, 1.5 and 0.5 and requires the edge value of `phi` to be `phiedge * bloat`; it fails at 1.5 and 0.5 before the change.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
@@ -335,7 +335,9 @@ std::string RadialProfiles::profileTypeToString(ProfileType profileType) {
 
 /** Compute the maximum toroidal and poloidal magnetic fluxes. */
 void RadialProfiles::computeMagneticFluxes() {
-  maxToroidalFlux = signOfJacobian * id_.phiedge / (2.0 * M_PI);
+  // the bloating factor scales the enclosed toroidal flux
+  const double phiedge = id_.phiedge * id_.bloat;
+  maxToroidalFlux = signOfJacobian * phiedge / (2.0 * M_PI);
   double edgeToroidalFluxFromProfile = torflux(1.0);
   if (edgeToroidalFluxFromProfile != 0.0) {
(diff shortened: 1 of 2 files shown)


jurasic-pf (approved):

jurasic-pf:
resolve conflicts please, then we can merge

CharlesCNorton:

CharlesCNorton:
fixing conflict

CharlesCNorton:

PR #810, opened 2026-09-06
Integrate the toroidal flux with the half-grid flux derivative

CharlesCNorton:
`RecomputeToroidalFlux` accumulates the full-grid `dphi/ds` at the inner endpoint of each interval, a left-endpoint rectangle rule. Fortran VMEC accumulates the half-grid `phip` between the two surfaces (`fileout.f90`), the midpoint rule, which is exact for a linear `dphi/ds`. The two agree only when `dphi/ds` is constant, which is the default `aphi = [1]` that every shipped case uses.

With any other `aphi` the reported flux is wrong at first order in the radial step. On `cth_like_fixed_bdy` at `ns = 15` with `aphi = [1, 0.5]`, whose flux profile is `phiedge (s + s^2/2) / 1.5`:

| quantity | before | after |
|---|---|---|
| edge value of `phi` | -3.4167e-02 against phiedge -3.5000e-02 | -3.5000e-02 |
| `phi` | 2.4e-02 | below 1e-09 |
| `currumnc` | 7.1e-01 | below 1e-09 |
| `bsubsmns` | 3.2e-01 | below 1e-09 |
| `currvmnc` | 2.5e-01 | below 1e-09 |
| `bdotb` | 2.5e-02 | below 1e-09 |

Relative to educational_VMEC reading the same INDATA file.

Agreement across `aphi` depends on the degree of `dphi/ds`, and only the linear cases reach round-off. With `aphi` of `[1]` and `[1, 0.5]` every wout field agrees at the 1e-07 residual level with matching iteration counts, apart from the axis `chipf` that the reference leaves at zero. `[1, -0.3, 0.3]` leaves 2.4e-04 in the worst field and 1.5e-05 in `phi`, and `[1, 0, 0, 0.5]` 2.4e-04.

That remainder is the reference, not this change. Fortran's `torflux` (`magnetic_fluxes.f90`) integrates the `aphi` polynomial with a fixed 100-panel trapezoid, which is exact only up to a linear integrand, and the result normalizes its enclosed flux. The error is independent of the radial grid: the code-to-code deviation in `phi` for `[1, -0.3, 0.3]` is 1.500e-05 at ns = 15, 29 and 57 alike, which is the trapezoid error `a_2 / 20000` for that coefficient. VMEC++ integrates the polynomial in closed form, so its own edge flux converges on `phiedge` as the radial grid refines, at 3.8e-04, 9.6e-05 and 2.4e-05 for those three resolutions, second order in the step, where the reference stalls against its quadrature floor.

The equilibrium itself is unaffected: `phi` is post-processing, and `rmnc`, `zmns` and the geometry were already within 1e-09. The same accumulation in the geometry adapter, used when `phiF` has not been filled, is switched over as well.

Two tests, each failing on the rectangle rule. `ToroidalFluxFollowsTheAphiPolynomial` solves the case at `ns = 9` with `aphi = [1, 0.5]` and requires `phi` to equal the analytic integral at every surface, which the rectangle rule misses by 2 percent. `DerivesToroidalFluxFromTheHalfGridDerivative` drives `MakeGeometry` with no `phiF` to copy, so the adapter's own accumulation runs, over a linear `dphi/ds` whose integral is known in closed form.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/geometry/vmec_geometry.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/geometry/vmec_geometry.cc
@@ -91,7 +91,7 @@ Geometry MakeGeometry(const VmecINDATA& indata,
       result.toroidal_flux[j] =
           result.toroidal_flux[j - 1] + internal.sign_of_jacobian * 2.0 *
                                             std::numbers::pi * delta_s *
-                                            internal.phipF[j - 1];
+                                            internal.phipH[j - 1];
     }
   }
   for (int j = 1; j < internal.num_full; ++j) {
--- a/src/vmecpp/cpp/vmecpp/vmec/geometry/vmec_geometry_test.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/geometry/vmec_geometry_test.cc
@@ -45,6 +45,46 @@ TEST(VmecGeometryTest, ConvertsInternalScalingWithoutWout) {
   EXPECT_NEAR(point.poloidal_flux[1], -1.6 * M_PI, 1e-14);
 }
 
+// With no phiF to copy, the adapter integrates the enclosed toroidal flux from
+// the half-grid dphi/ds between the two surfaces. That is exact for a linear
+// dphi/ds, where the value at the inner endpoint of each interval is not.
+TEST(VmecGeometryTest, DerivesToroidalFluxFromTheHalfGridDerivative) {
+  VmecINDATA indata;
+  indata.nfp = 1;
+  indata.mpol = 2;
+  indata.ntor = 0;
+
+  const int num_full = 4;
+  const double delta_s = 1.0 / (num_full - 1);
+  VmecInternalResults internal;
+  internal.sign_of_jacobian = 1;
+  internal.lamscale = 1.0;
+  internal.num_full = num_full;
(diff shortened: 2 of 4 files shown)


jurasic-pf:
Just a nit The new ToroidalFluxFollowsTheAphiPolynomial test only exercises RecomputeToroidalFlux but not the vmec_geometry.cc fallback branch (has_toroidal_flux == false), which got the identical fix. geometry_test.cc exists and already tests Geometry/EvaluateProfile, so a small addition there covering the non-fixed-phiF path would close the gap cheaply.

CharlesCNorton:
Added DerivesToroidalFluxFromTheHalfGridDerivative in vmec_geometry_test.cc, which is where MakeGeometry is reachable from; geometry_test.cc links only the geometry target. It drives the adapter with no phiF to copy over a linear dphi/ds whose integral is known in closed form, and fails on the rectangle rule at every surface.

CharlesCNorton:
Correcting a measurement in the description: I had written that every wout field agrees below 1e-09 across four aphi values. That holds for the two whose dphi/ds is linear, not for the higher-degree ones, which leave 2.4e-04. The remainder is the reference's own 100-panel trapezoid in torflux: the code-to-code deviation in phi is 1.500e-05 at ns 15, 29 and 57 alike, while VMEC++'s edge flux converges on phiedge second order in the radial step. Description updated.

PR #811, opened 2026-09-06
Integrate the line-segment profile to the evaluation point, not the next knot

CharlesCNorton:
`evalLineSegmentIntegrated` builds the last, truncated piece of the integral from the value at the *next knot* rather than the value at the truncation point, so the straight line it integrates has the wrong slope whenever the evaluation point falls between two knots. The result is exact on the knots and wrong in between.

For knots at `0, 0.1, ..., 1.0` and values `1.0, 0.95, 0.86, 0.74, 0.60, 0.46, 0.33, 0.21, 0.12, 0.06, 0.03`:

| s | before | exact |
|---|---|---|
| 0.05 | 4.875000e-02 | 4.937500e-02 |
| 0.25 | 2.280000e-01 | 2.295000e-01 |
| 0.50 | 3.880000e-01 | 3.880000e-01 |
| 0.75 | 4.627500e-01 | 4.638750e-01 |
| 1.00 | 4.845000e-01 | 4.845000e-01 |

`pcurr_type = 'line_segment_ip'` is the only profile that reaches this path. On `cth_like_fixed_bdy` with `ncurr = 1` and the current derivative above, the converged equilibrium differed from educational_VMEC by 1.8e-02 in `iotaf`, 1.6e-02 in `jcurv`, 8.6e-04 in `ctor` and 3.7e-05 in `rmnc`, and by 7.2e-01 in `DCurr`. Every field now agrees below 1e-09 with the iteration count unchanged at 413. The other nine current parameterizations, including `akima_spline_ip` and `cubic_spline_ip`, already agreed and still do.

The walk over the knots also read one past the end of both arrays when the evaluation point lay beyond the last knot, and added that piece twice.

The rewritten loop integrates the interpolant from 0 to x piece by piece, taking each endpoint value from the interpolant itself, which also covers the extrapolation below the first knot and above the last. The test compares it against a fine trapezoidal quadrature of the same interpolant on and between the knots and outside both ends.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
@@ -982,37 +982,32 @@ double RadialProfiles::evalLineSegmentIntegrated(
   }
 
   auto integrate_segment = [](double x0, double x1, double y0, double y1) {
+    if (x1 == x0) {
+      return 0.0;
+    }
     const double m = (y1 - y0) / (x1 - x0);
     const double b = y0 - m * x0;
     return m * 0.5 * (x1 * x1 - x0 * x0) + b * (x1 - x0);
   };
+  auto value_at = [&](double s) {
+    return evalLineSegment(splineKnots, splineValues, s);
+  };
 
-  double xi = x;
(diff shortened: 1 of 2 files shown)


CharlesCNorton:
Folded into #812, which now fixes both line-segment evaluators together as requested there. Closing.

PR #812, opened 2026-09-06
Evaluate and integrate a line-segment profile over the same interpolant

CharlesCNorton:
The two `line_segment` evaluators disagreed about what profile they represent.

`evalLineSegment` returned the first or last knot value outside the knots, where Fortran VMEC's `line_seg` continues the first or last segment, taking `ilow, ihigh = 1, 2` below the knots and `n-1, n` above them.

`evalLineSegmentIntegrated` integrated something else again. It started the sum at the first knot, so the piece from `s = 0` up to that knot was dropped; below the first knot it drew a line from `(0, values[0])` to the evaluation point, which is the clamped value rather than the profile's own value at zero; and the truncated last piece took its upper value from the next knot instead of from the evaluation point, so that piece carried the wrong slope. The walk also read one past the end of both arrays beyond the last knot.

Both now describe one interpolant: linear on each knot interval and continued along the first and last segments outside them. `evalLineSegmentIntegrated` integrates that interpolant from 0 to x, splitting at the knots that fall inside the interval and taking every endpoint value from `evalLineSegment`, which is exact because each piece is linear.

Measured against educational_VMEC reading the same INDATA file, largest relative deviation over the wout, with iteration counts in brackets.

`PMASS_TYPE = 'line_segment'` with nine knots on `[0.1, 0.9]`, a cth-like fixed-boundary case:

| quantity | before | after |
|---|---|---|
| `DWell` | 8.98e-01 | 2.46e-12 |
| `DCurr` | 5.93e-01 | 1.10e-10 |
| `betaxis` | 8.40e-02 | 1.75e-14 |
| `presf` | 8.34e-02 | 3.79e-16 |
| `iotaf` | 4.47e-04 | 4.47e-14 |
| iterations | [414] | [415], as Fortran |

`PCURR_TYPE = 'line_segment_ip'` on the same case, which is the only profile reaching the integrated path:

| quantity | before | after |
|---|---|---|
| `DCurr` | 7.17e-01 | 9.75e-12 |
| `iotaf` | 1.80e-02 | 4.37e-14 |
| `jcurv` | 1.57e-02 | 4.95e-14 |
| `ctor` | 8.60e-04 | 2.02e-15 |
| `rmnc` | 3.72e-05 | 2.46e-14 |
| iterations | [413] | [413] |

Every wout field of both cases now agrees below 1e-10, and so do `line_segment_i` and the same pressure profile with knots reaching `[0.0, 1.0]`, which is what every shipped input uses and which is unchanged. The nine other current parameterizations, `akima_spline_ip` and `cubic_spline_ip` among them, already agreed and still do.

The test compares the integrated form against a fine quadrature of `evalLineSegment` itself, on knots, between them, below the first knot and beyond the last, with knots that reach neither end of the domain; it fails on the previous integral at every one of those points. The direct evaluator keeps its existing test, with the two points outside the knots pinned to the continued values instead of the clamped ones.

This supersedes #811, which fixed the truncated last piece alone.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/radial_profiles/radial_profiles.cc
@@ -954,8 +954,8 @@ double RadialProfiles::evalRational(const Eigen::VectorXd& coeffs, double x) {
                               : std::numeric_limits<double>::max();
 }
 
-// Linear interpolation between the two knots that bracket x.
-// Clamp the profile if x is outside the range of knots.
+// Linear interpolation between the two knots that bracket x, continued along
+// the first or last segment for an x outside the knots.
 double RadialProfiles::evalLineSegment(const Eigen::VectorXd& splineKnots,
                                        const Eigen::VectorXd& splineValues,
                                        double x) {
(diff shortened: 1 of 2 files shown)


jurasic-pf:
evalLineSegmentIntegrated is now inconsistent with evalLineSegment i think, can we apply the fix to both?

jurasic-pf:
in particular i think the constant coeff at 0 is missing

CharlesCNorton:
Both evaluators now describe one interpolant: the integral runs from 0 to x over exactly what evalLineSegment returns, which restores the piece below the first knot you spotted and fixes the truncated last piece. line_segment_ip on a cth-like case goes from 7.2e-01 in DCurr and 1.8e-02 in iotaf against educational_VMEC to below 1e-10 in every wout field; this supersedes #811, which I have closed.

PR #815, opened 2026-09-07
Bound the nzeta-independence lambda comparison by its measured spread

CharlesCNorton:
`AxisymmetricRunIsIndependentOfNzeta` holds every compared quantity to one `1e-8`. Lambda does not fit under it. The test compares an `nzeta = 1` run against an `nzeta > 1` run of the same axisymmetric equilibrium, and the sum over identical planes changes the round-off that seeds the descent, so the two runs follow different paths into the same minimum. Lambda is the softest direction and carries the largest part of that difference.

Measured on the three builds this repository tests, at 1, 4, 8 and 16 threads, `lmns_full` ranges from 2.2e-13 to 1.6e-11 locally, while `rmnc` stays at 2.8e-13, `zmns` at 4.6e-14 and `jcurv` at 6.3e-12. On the CI runners the same comparison reads 1.06e-08 under asan and 1.46e-08 under ubsan, both above the bound, on runs whose only difference from a passing run is the machine. The failures are intermittent rather than systematic, so the site sits at the edge rather than past it, and it fails PRs that do not touch this code: it took down `Build and run asan tests` on #761, which changes one file in a different test binary, and then `Build and run ubsan tests` on the retry.

The lambda comparisons move to their own `1e-6`, about 70 times the largest value observed anywhere. Everything else keeps `1e-8`, which is three orders above what those quantities measure. What the test discriminates is unaffected: the defect it was written for, a toroidal loop that filled one plane, moves these arrays by order one.

`bazel test //vmecpp/vmec/vmec:vmec_test` passes under `opt`, `asan` and `ubsan`, and `bazel test --config=opt //vmecpp/...` passes.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec_test.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/vmec/vmec_test.cc
@@ -387,9 +387,12 @@ TEST(TestVmec, AxisymmetricRunIsIndependentOfNzeta) {
     const auto& a = single_plane->wout;
     const auto& b = planes->wout;
 
-    // The sum over identical planes changes the round-off, which the descent
-    // carries into lambda at the 1e-9 level.
+    // The sum over identical planes changes the round-off, and the descent
+    // carries that into lambda, the softest direction: it measures between
+    // 2e-13 and 1.5e-08 over the opt, asan and ubsan builds at 1 to 16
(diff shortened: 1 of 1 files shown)


jurasic-pf:
I think the required diff is just relaxing the tolerance minimally.

jurasic-pf (approved):

PR #816, opened 2026-09-07
Reject a profile that its evaluator cannot evaluate

CharlesCNorton:
**Change** - `IsConsistent` rejects a profile its evaluator cannot evaluate. The parameterization table records, per name, the knots a spline evaluator needs (4 for the Akima and cubic families, 2 for the line segments) and the coefficients a closed form needs (`gauss_trunc` 2, `two_power` and `two_power_gs` 3, `two_lorentz` 8; the zero-padded power series accept any count), and `CheckProfile` rejects an input that carries fewer. Spline knots must increase strictly, and the Akima and cubic families, which are evaluated only inside their knots, must run from 0 to 1; the line segments continue their end segments and may stop short. `NeedsSplineData()` becomes the knot count being positive, so its one caller is unchanged.

**Cause** - `CheckProfile` required a spline profile to carry knots and the two arrays to have the same length, but not to be long enough to evaluate. Below their counts the evaluators return `0.0` for every `s` after one warning: `evalAkima` and `evalCubic` need four knots, `evalLineSegment` two, and `gauss_trunc`, `two_power`, `two_power_gs` and `two_lorentz` need 2, 3, 3 and 8 coefficients. The Akima and cubic evaluators also return `0.0` outside their knots. That is the failure the validation added in #719 exists to prevent, in its own words: an unusable parameterization "reaches the solver as a zero profile, which converges to a silently wrong equilibrium." Fortran VMEC halts instead. `profile_functions.f` counts the knots and stops before the evaluator: `IF (i < 4) STOP 'pcurr: check s-grid for curr-grid values!'`, with the same guard on the pressure and iota branches.

**Evidence** - `cth_like_fixed_bdy.json` unchanged, and then with `pmass_type = "cubic_spline"` over three and four knots, on main:

| | `two_power` (as shipped) | `cubic_spline`, 3 knots | `cubic_spline`, 4 knots |
|---|---|---|---|
| exit | converged | converged | converged |
| `presf(0)` | 4.322928e+02 | 0.000000e+00 | 4.322908e+02 |
| `betatotal` | 1.926384e-03 | 0.000000e+00 | 1.681736e-03 |
| iterations | 121 | 119 | 121 |

The three-knot run writes a `wout` reporting a converged zero-pressure equilibrium while the input asked for 432 Pa. With `ncurr = 1` the same holds for the prescribed current, as in #798.

**Tests** - in `vmec_indata_test`: `CheckSplineProfilesNeedEnoughKnots` drives the three spline families one knot short of and at their minimum for the pressure profile and holds the current and iota profiles to the same counts; `CheckClosedFormProfilesNeedEnoughCoefficients` does the same for the four closed forms; `CheckSplineKnotsIncreaseAndSpanTheRadius` rejects unordered and repeated knots and Akima or cubic knots that stop short of 0 or 1, and accepts line segments that do.

**Scope** - validation only; the evaluators are untouched. An input rejected now is one whose profile evaluated to zero over part or all of the radius.

### DIFF
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
@@ -44,17 +44,19 @@ std::string ProfileTypeName(vmecpp::ProfileType profile_type) {
 }
 
 // Checks that `type_name` names a profile parameterization that may be used for
-// `profile_type`, and that a spline parameterization was given its knots. An
-// unrecognized name otherwise reaches the solver as a zero profile, which
-// converges to a silently wrong equilibrium.
+// `profile_type`, that a spline parameterization was given its knots, and that
+// a closed-form parameterization was given the coefficients it needs. An
+// unrecognized name or a profile short of its data otherwise reaches the
+// solver as a zero profile, which converges to a silently wrong equilibrium.
 //
-// The polynomial coefficient arrays are deliberately not required to be
-// non-empty: they are zero-padded on read, so an empty array is a valid way to
-// specify a zero profile.
+// The polynomial coefficient arrays are zero-padded on read, so an empty
+// array is a valid way to specify a zero power series; the closed forms that
+// raise to a coefficient or divide by one need their full set.
 absl::Status CheckProfile(const std::string& type_key,
                           const std::string& type_name,
                           vmecpp::ProfileType profile_type,
-                          const std::string& aux_key,
+                          const std::string& coefficient_key,
+                          const Eigen::VectorXd& coefficients,
                           const Eigen::VectorXd& aux_s,
                           const Eigen::VectorXd& aux_f) {
   const vmecpp::ProfileParameterizationData* const parameterization =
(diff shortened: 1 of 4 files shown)


CharlesCNorton:
The opt failure is Threed1ShafranovIntegralsTest/cma at 5.462e-11 against its 5.0e-11 bound, in a test this PR cannot reach: cma uses power_series, whose minimum knot count is zero, so the new check never fires. It passes locally at opt on this branch. #764, now unstacked and standalone, raises that bound to 5.0e-10 from the measured 4.08e-11.

CharlesCNorton:
Extended to the other ways a profile reaches the solver as zero. The closed-form parameterizations have the same gap as the splines: below their coefficient count `gauss_trunc`, `two_power`, `two_power_gs` and `two_lorentz` log a warning and return zero for every s, so `pmass_type = two_power` with `am = [1000]` converges to a zero-pressure equilibrium; the table now carries the count each needs (2, 3, 3, 8) beside the knot count. The spline evaluators walk their knots in order and the Akima and cubic ones return zero outside them, where Fortran VMEC stops with `iflag = -2` and `-1`; knots out of order gave a different equilibrium without a message (beta 1.80e-3 against 1.68e-3 for the same five points on `cth_like_fixed_bdy`), and knots ending at 0.9 gave zero pressure beyond them. `IsConsistent` now requires strictly increasing knots for every spline and knots from 0 to 1 for the Akima and cubic families; the line segments continue their end segments and may stop short. `CheckClosedFormProfilesNeedEnoughCoefficients` and `CheckSplineKnotsIncreaseAndSpanTheRadius` cover it.

PR #820, opened 2026-09-07
Add an opt-in undamped lambda preconditioner

CharlesCNorton:
**Change** - `VmecInput.undamped_lambda_preconditioner`, off by default (JSON and HDF5 round trip through `VmecINDATA`). Set, the lambda preconditioner scales its assembled diagonal by `1 / lamscale^2` alone, through `kLambdaPreconditionerUndampedFactor`; unset, the scale stays `kLambdaPreconditionerDampingFactor / (4 lamscale^2)`.

**Cause** - the `1 / lamscale^2` converts the internally rescaled lambda coefficients. The remaining factor of 0.5 comes from VMEC, where `lamcal.f90` sets `damping_fac = 2.0` beside a commented-out `pfactor0/4` marked "sometimes helps convergence"; neither the Fortran nor the VMEC++ comment gives a derivation for it.

**Evidence** - iterations summed over the multigrid schedule, single-threaded, each case at its own ftol. The twelve QUASR configurations of `tests/test_free_boundary_quasr.py`, run in fixed-boundary mode at mpol = ntor = 6 and ns_array = [8, 16, 31]:

| case | off | on |
| --- | --- | --- |
| 954 | 141 | 133 |
| 9914 | 328 | 261 |
| 19493 | 819 | 689 |
| 19609 | 587 | 447 |
| 19940 | 492 | 416 |
| 29346 | 431 | 432 |
| 50136 | 1325 | 1331 |
| 65579 | 905 | 772 |
| 112718 | 636 | 509 |
| 165868 | 441 | 377 |
| 167381 | 464 | 386 |
| 336902 | 461 | 367 |
| total | 7030 | 6120 |

and the fixed-boundary cases of the test suite:

| case | off | on |
| --- | --- | --- |
| solovev | 204 | 217 |
| solovev_analytical | 400 | 397 |
| circular_tokamak | 749 | 589 |
| cma | 207 | 148 |
| cth_like_fixed_bdy | 121 | 113 |
| cth_like_fixed_bdy_iota | 118 | 111 |
| li383_low_res | 124 | 133 |
| up_down_asym | 594 | 635 |
| cth_like_fixed_bdy_asym | 137 | 132 |
| total | 2654 | 2475 |

That is 13 percent fewer iterations over the QUASR set and 7 percent over the test suite, at no cost per iteration, since the change is the value of a scalar computed once per preconditioner update. The gain is not uniform. Three of the nine test-suite cases take more iterations, and on the W7-X example the count is not a smooth function of the factor: scanning 2.0 through 6.0 gives 2954, 2699, 3604, 3616, 3520, 2580, 2609 and 3017, a twenty percent swing in either direction with no trend.

The converged state moves with the descent path. At a case's own ftol the two settings stop about 1e-3 apart in `iotaf`; held to 1e-12 they are 4.1e-6 apart on `cth_like_fixed_bdy` and 5.1e-6 on `cth_like_fixed_bdy_asym`, and the enclosed volume is identical at every tolerance. What remains at 1e-12 is the spectral-condensation angle gauge, which moves the poloidal spectrum without moving the flux surfaces: `zmns` differs by 2.9e-4 where `wb` agrees to 2.3e-8. Changing only the thread count leaves all of these below 3e-14.

**Tests** - in `vmec_test`, on `solovev`, `cth_like_fixed_bdy` and `cth_like_fixed_bdy_asym`: `UndampedLambdaPreconditionerLeavesTheFirstForceResidual` requires `fsqt(0)` to be bit-identical between the two settings, since the scale multiplies an already assembled force, and both runs to reach the tolerance; `UndampedLambdaPreconditionerConvergesToTheSameEquilibrium` compares `rmnc`, `zmns`, `iotaf` and `wb` at ftol 1e-12 against five times the measured deviation.

**Scope** - unset, the assembled preconditioner is bit-identical to the current one, which every case above reproduces exactly.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -442,6 +442,9 @@ class VmecInput(BaseModelWithNumpy):
     lforbal: bool = False
     """Hack: directly compute innermost flux surface geometry from radial force balance"""
 
+    undamped_lambda_preconditioner: bool = False
+    """Scale the lambda preconditioner by 1 / lamscale^2 alone."""
+
     return_outputs_even_if_not_converged: bool = False
     """If true, return a wout even if VMEC++ did not converge, instead of raising a
     RuntimeError.
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
@@ -227,6 +227,7 @@ VmecINDATA::VmecINDATA() {
   delt = 1.0;
   tcon0 = 1.0;
   lforbal = false;
+  undamped_lambda_preconditioner = false;
   iteration_style = IterationStyle::VMEC_8_52;
   return_outputs_even_if_not_converged = false;
 
@@ -352,6 +353,8 @@ absl::Status VmecINDATA::WriteTo(H5::H5File& file) const {
   WriteH5Dataset(delt, "/indata/delt", file);
   WriteH5Dataset(tcon0, "/indata/tcon0", file);
   WriteH5Dataset(lforbal, "/indata/lforbal", file);
+  WriteH5Dataset(undamped_lambda_preconditioner,
+                 "/indata/undamped_lambda_preconditioner", file);
   WriteH5Dataset(return_outputs_even_if_not_converged,
                  "/indata/return_outputs_even_if_not_converged", file);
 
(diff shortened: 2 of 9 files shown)


jurasic-pf:
Oh, that's very cool!

jurasic-pf:
I'll try to review properly in the upcoming week

Issue #818, opened 2026-09-07
Further improve the preconditioning strategy


jurasic-pf:
Many terms like tcon0 are empirically chosen. There's various ways the default VMEC preconditioner could be improved:
- How do the solver dynamics change when solving exactly for lambda at each step, e.g. by preconditioning it via the inverse? Does it negatively affect robustness?
- hyperparameters could be chosen via a data driven approach over a large dataset (prioritizing robust convergence on all cases, then convergence speed)
- Can we invert a better "proxy" system for the radial preconditioner?

Motivated by this investigation into one of the hard-coded parameters by @CharlesCNorton https://github.com/proximafusion/vmecpp/pull/731#discussion_r3918870603_

CharlesCNorton:
Measurements on the three bullets, over the 9 fixed-boundary test-suite cases and the 12 QUASR configurations run fixed-boundary.

**Exact lambda solve.** lambda enters the energy through its half-grid average, so the exact row is `(L_in, L_in + L_out, L_out) / 2`, an averaging operator whose null direction is the radial sawtooth. Written as `(-w L_in, L_in + L_out, -w L_out) / 2` the condition number is bounded by `(1+|w|)/(1-|w|)`, with `w = 0` the present diagonal form, reproduced bit for bit. Toward the exact operator: cth_like_fixed_bdy goes 121 -> 163 iterations at `w = -0.5`, and no case converges at `w = -0.9`. In conditioning the exact lambda block is worth 1.06 to 1.32, against 4.2 to 10.2 for the R/Z block.

**A better proxy.** The full second variation over R, Z and lambda jointly, banded to `|dj| <= 1`, assembled by differencing the force and factorised as one block-tridiagonal system. Conditioning is 30x better than VMEC++'s at ns=9 and 37x at ns=15 on cth_like_fixed_bdy. Iteration counts, direction from the exact solve and length from the per-mode one, blended `a` between them so `a = 0` is today:

| case | base | a=0 | a=0.1 | a=0.25 | a=0.5 | a=1 |
| --- | --- | --- | --- | --- | --- | --- |
| solovev ns=9 | 118 | 118 | 127 | 166 | 606 | fail |
| cth_like ns=9 | 223 | 223 | 237 | 264 | 434 | fail |
| cth_like_asym ns=9 | 230 | 230 | 234 | 246 | 461 | |
| li383 ns=15 | 189 | 189 | 199 | 257 | 363 | |
[...]

PR #822, opened 2026-09-07
Default tcon0 to 0.5

CharlesCNorton:
**Change** - changes the `tcon0` default in `VmecINDATA` and `VmecInput` from 1.0 to 0.5.

**Cause** - 1.0 is inherited from VMEC and underived; #818 names it as the example of an empirically chosen term. It also sits beside a cliff: at 1.5 and 2.0 none of the 21 configurations below converge.

**Evidence** - iterations summed over the multigrid schedule, single-threaded, each case at its own ftol, over the 9 fixed-boundary test-suite cases and the 12 QUASR configurations of `tests/test_free_boundary_quasr.py` run fixed-boundary. Every row converges 21/21:

| tcon0 | lambda scale | iterations |
| --- | --- | --- |
| 1.0 | inherited | 9684 |
| 0.5 | inherited | 8244 |
| 1.0 | undamped (#820) | 8595 |
| 0.5 | undamped (#820) | 7459 |

15 percent on its own, and independent of #820. A coordinate sweep over `kNDamp`, the preconditioner update interval, `delt` and the two high-m lambda damping terms found those already at or beside their best values, so `tcon0` and the lambda scale are the two that move.

**Tests** - every input under `test_data` sets `tcon0` explicitly, so no reference comparison moves; `TestVmecINDATA.CheckDefaults` is updated to 0.5. Both C++ suites pass, 44/44. `examples/data/solovev.json` omits `tcon0`, so the `run` doctest sees the new default and `b0` moves from 0.2033313711 to 0.2033305335; the example now rounds to six digits, 0.203331, which both values share and which also covers the C library variation its comment already notes.

**Scope** - affects only inputs that omit `tcon0`. Fortran clamps the value to `<= 1`, so 0.5 stays in range.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -436,7 +436,7 @@ class VmecInput(BaseModelWithNumpy):
     delt: float = 1.0
     """Initial value for artificial time step in iterative solver."""
 
-    tcon0: float = 1.0
+    tcon0: float = 0.5
     """Constraint force scaling factor for ns --> 0."""
 
     lforbal: bool = False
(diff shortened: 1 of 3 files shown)


CharlesCNorton:
think the ci failure was an outage

CharlesCNorton:
Closed and reopened to rerun the checks. Every failure of the previous run was a 504 from GitHub while the jobs downloaded their dependencies.

PR #824, opened 2026-09-07
Scale extcur with the magnetic field when rescaling a free-boundary equilibrium

CharlesCNorton:
`rescale` scales the boundary, the axis, phiedge, curtor and pres_scale, but leaves `extcur` untouched. For a free-boundary run the vacuum field is set by the coil currents, so a state scaled to B -> b_scale * B sits against an unchanged external field and is no longer in force balance. `rescale` returns it after a single iteration against `ftol = 1.0`, so anything below that residual is reported as converged.

On `cth_like_free_bdy` (converged at fsqr = 9.3e-13):

| b_scale | fsqr returned | |
|---|---|---|
| 1.5 | 1.6e+00 | raises |
| 1.02 | 7.8e-03 | returned as converged |
| 1.5, extcur scaled | 4.5e-12 | |

`solovev_free_bdy` behaves the same way: 3.6e+00 at b_scale 1.5, 1.8e-02 returned silently at 1.02, and 2.7e-12 once extcur carries the factor.

`extcur` now scales with `b_scale`. Fixed-boundary results are unchanged, and the scaling was already exact there: on solovev, cth_like_fixed_bdy and cth_like_fixed_bdy_asym every scaling law holds to 1e-15 and the rescaled residual equals the base one.

`r_scale` is rejected for a free-boundary input. The mgrid file fixes both the grid extent and the coil geometry, and scaling the currents does not compensate: at r_scale = 1.02 the residual stays at 1.1e-02 either way, and larger factors move the boundary outside the vacuum grid, which `MGridProvider::interpolate` reports as clamped and not physically meaningful.

Both new tests fail on main and pass here. `pytest tests/` is 210 passed, 2 skipped.

https://claude.ai/code/session_01QG1y5xASoAYZtrdSLwTrPN

### DIFF
--- a/src/vmecpp/_rescale.py
+++ b/src/vmecpp/_rescale.py
@@ -5,7 +5,7 @@ import typing
 import numpy as np
 
 if typing.TYPE_CHECKING:
-    from vmecpp import VmecOutput
+    from vmecpp import MagneticFieldResponseTable, VmecOutput
 
 
 def rescale(
@@ -13,6 +13,7 @@ def rescale(
     b_scale: float,
     r_scale: float,
     scale_pressure: bool = True,
+    magnetic_field: "MagneticFieldResponseTable | None" = None,
 ) -> "VmecOutput":
     from . import run  # noqa: PLC0415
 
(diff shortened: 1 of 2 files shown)


jurasic-pf:
Fair enough

jurasic-pf (approved):

CharlesCNorton:
fixing conda

PR #828, opened 2026-09-08
Carry signgs in the m = 1 polar constraint

### LINKED ISSUE #827: Make sure flipping `signgs` doesn't change physics
VMEC uses left handed coordinates (to the dismay of many). In principle, there's a `signgs` input variable, also called `signOfJacobian` in the code, that can flip between the two. Unfortunately we know that the `signgs` term isn't carried in all the computations that need it (approximately all the uneven terms that follow sin symmetry), so it results in wrong physics. Cleaning this up would simplify interop with other codes without thinking about coordinate handedness at each step along the way.

A valuable cleanup task would be to add/remove signgs factors across the codebase where they're needed. The verification is that we can run vmecpp with the same input with `signgs=1` or `signgs=-1` (and flipping all the sin terms of the boundary by -1), and get the same wout. First in the stellarator symmetric case, then also in asym and free-boundary cases.

Ideally the PRs related to this should be all of the same simple form: A test case, or occasionally unit/component tests for parts of the code with only signgs changing, and some hardcoded + or - replaced by signgs, that's it.

CharlesCNorton:
**Change** - `signgs` becomes an input (`VmecInput.signgs`, default -1, JSON and HDF5 round trip), and every copy of the m = 1 polar constraint carries it: the coupling `rss <- rss + sigma zcs`, `zcs <- sigma rss - zcs` with `sigma = -signgs` in the coefficient map, the boundary, the hot-restart geometry, the force transpose, the wout recombination and the undo inside the axis guess, and the poloidal shift that puts an asymmetric boundary into the gauge `rbs(1,0) = sigma zbc(1,0)`.

**Cause** - the constraint was written for the left-handed system, where the frozen m = 1 combination is `rss = zcs`; in the right-handed one it is `rss = -zcs`. With `signgs = +1` the first update left the mirror image in every odd-m, n != 0 mode, an asymmetric boundary was shifted into the wrong gauge, and the axis guess decoded the boundary with the wrong sign and failed to start.

**Evidence** - each case run as is and as its reflection through the midplane in right-handed coordinates (`signgs = +1`, sine coefficients of the boundary and axis negated, `curtor` reversed, the response table reflected with `B_Z` negated) takes the same number of iterations and agrees to 1e-9 in every wout field, identically or with the sign of Z, the Jacobian and the current diagnostics reversed: `cth_like_fixed_bdy`, `solovev`, `li383_low_res` (axis recompute), `cth_like_fixed_bdy_asym`, `up_down_asym`, `cth_like_free_bdy`, and `cth_like_free_bdy_asym` on the cth coils raised by 5 mm, where the plasma settles 5.8 mm off the midplane. Flipping only `signgs` on an unchanged input reproduces the equilibrium under the theta -> pi - theta relabelling the code applies; a hot restart from a `signgs = +1` state converges in two iterations.

**Tests** - `tests/test_signgs.py`: the seven mirror comparisons, the relabelling with `signgs` alone, the hot restart, and the rejection of a `signgs` outside {-1, +1}.

**Scope** - `signgs = -1` is bit-identical to main. Fixes #827.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -410,6 +410,13 @@ class VmecInput(BaseModelWithNumpy):
     nvacskip: int = 1
     """Number of iterations between full vacuum calculations."""
 
+    signgs: int = -1
+    """Sign of the Jacobian of the (s, theta, zeta) coordinates: -1 for the left-handed
+    system of Fortran VMEC, +1 for a right-handed one.
+
+    The input boundary is flipped in theta to match it.
+    """
+
     free_boundary_method: typing.Annotated[
         FreeBoundaryMethod,
         pydantic.BeforeValidator(_validate_free_boundary_method),
--- a/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
+++ b/src/vmecpp/cpp/vmecpp/common/vmec_indata/vmec_indata.cc
@@ -218,6 +218,7 @@ VmecINDATA::VmecINDATA() {
   mgrid_file = "NONE";  // default from Fortran VMEC via indata2json
   // extcur is left empty
   nvacskip = 1;
+  signgs = -1;
   free_boundary_method = FreeBoundaryMethod::NESTOR;
 
   // tweaking parameters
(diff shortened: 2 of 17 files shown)


CharlesCNorton:
@jurasic-pf @jons-pf The PR now covers the asymmetric and free-boundary cases too: the poloidal shift of an asymmetric boundary and the constraint undo in the axis guess were the two remaining left-handed copies of the m = 1 constraint. Every bundled case, fixed and free boundary, reproduces its midplane reflection with signgs = +1 to 1e-9 in the same number of iterations; NESTOR needed no change.

CharlesCNorton:
Added one more path: an asymmetric boundary handed in with only `signgs` flipped goes through the theta flip, and the poloidal shift had already put it into the gauge of the other direction, so the run converged to a slightly different equilibrium (iota off by 1.6e-3 on `cth_like_fixed_bdy_asym`). The shift now reads the poloidal direction from the m = 1 determinant first. `test_flipping_signgs_alone_relabels_theta` now covers the asymmetric cases, which reproduce the base to 1e-14 under the theta -> pi - theta relabelling.

jurasic-pf (approved):
Wohooo! left 👈 and right 👉 handed coordinates incoming

PR #833, opened 2026-09-08
Let the vacuum potential's Fourier cutoffs exceed the plasma's

CharlesCNorton:
**Change** - `VmecInput.vacuum_mpol` and `VmecInput.vacuum_ntor` set the Fourier cutoffs of NESTOR's scalar potential independently of the plasma's `mpol` and `ntor`, which they default to. The potential and the vacuum field keep the plasma's tangential grid; the boundary is zero-padded into the larger layout, `nzeta` has to carry the toroidal cutoff, and `potvac` is written at the potential's own size.

**Cause** - Fortran VMEC expands the vacuum potential to `mf = mpol + 1`, `nf = ntor`, so the vacuum solution is truncated at the plasma's resolution. On the cth boundary that truncation, not the singular quadrature, is the floor #628 discusses: an exact exterior Neumann problem (a line of dipoles inside the surface, whose total field on it is zero) gives a residual of 5.3e-3 relative to the external field with `ntor = 4` for the potential at any `mpol` up to 32 and any grid up to `ntheta = 160`, `nzeta = 288`, and 4.8e-4 with the potential at `ntor = 8`. Against BIEST at twelve digits on the real cth vacuum field, NESTOR's rms error at `mpol = 16` is 1.42e-3 with `ntor = 4`, 3.5e-4 at 6, 9.9e-5 at 8 and 7.3e-5 at 12.

**Evidence** - in the loop on `cth_like_free_bdy` at `ftol = 1e-11` the change moves the boundary by tens of micrometres at its 0.19 percent beta and by 0.27 mm (`mpol = 5`) to 1.3 mm (`mpol = 8`) in `rmax_surf` at 2.1 percent beta on the wide mgrid; the axis moves by up to 76 micrometres. The numbers above were measured with the cross-term sign of #825; with the historical sign the vacuum error stays at 6.9e-3 for every cutoff, so the two changes only pay together.

**Tests** - `VacuumCutoffsStayAboveThePlasmaResolution` in `vmec_indata_test` (defaults, rejection below the plasma's cutoffs and of an `nzeta` that cannot carry `vacuum_ntor`, JSON round trip), `VacuumPotentialCutoffsAboveThePlasmaResolution` in `vmec_in_memory_mgrid_test` (the expanded run converges on the same grid, reports `potvac` at its size and stays within the vacuum resolution change of the base run), `test_vacuum_potential_cutoffs` in `tests/test_free_boundary.py`.

**Scope** - defaults are bit-identical to main: the vacuum `Sizes` equals the plasma's and the hand-off layout is unchanged. Raising the cutoffs costs only the vacuum solve, which grows with the potential's mode count.

### DIFF
--- a/src/vmecpp/__init__.py
+++ b/src/vmecpp/__init__.py
@@ -260,6 +260,20 @@ class VmecInput(BaseModelWithNumpy):
     the full ntor. < 0 (default) means geometry uses ntor.
     """
 
+    vacuum_mpol: int = 0
+    """Poloidal Fourier cutoff of the vacuum potential in a free-boundary run.
+
+    NESTOR expands the vacuum potential to mpol like the plasma; a value above mpol
+    raises the potential's cutoff alone. 0 (default) means the potential uses mpol.
+    """
+
+    vacuum_ntor: int = 0
+    """Toroidal Fourier cutoff of the vacuum potential in a free-boundary run.
+
+    A value above ntor raises the potential's cutoff alone; nzeta must then be at least
+    2 * vacuum_ntor + 4. 0 (default) means the potential uses ntor.
+    """
+
     ntheta: int = 0
     """Number of poloidal grid points (ntheta >= 0).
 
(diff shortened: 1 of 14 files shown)


CharlesCNorton:
fixing

PR #835, opened 2026-09-08
Size the parallel regions with num_threads clauses instead of the process thread count

CharlesCNorton:
**Change** - every parallel region of a run states its team size with a `num_threads` clause, and nothing in a run calls `omp_set_num_threads` any more: the main region takes `vmec_adjust_num_threads`'s count, the vacuum region already took its own, and the output stage takes the run's `max_threads`. The thread budget a run reads from the runtime is therefore the process's, whatever ran before. The macOS wheel tests clear `OpenMP_ROOT` in their environment.

**Cause** - `vmec_adjust_num_threads` set the process-wide OpenMP thread count to `min(max_threads, ns / 2)` at every multigrid step, and an explicit `max_threads` did the same in `FlowControl`. The next run in the same process, and any other OpenMP user in it, then read that count as the maximum: after one cth run at `ns = 15` on a 20-thread machine, `omp_get_max_threads()` returns 7, a following `ns = 99` run solves on 7 radial threads instead of 20, every following free-boundary run solves its vacuum on 7 threads instead of 20, and a response table built afterwards is computed on 7. The smaller vacuum team also regroups the thread-ordered reductions of the vacuum solve, so the first free-boundary run of a process and all later ones disagreed at round-off: 1.4e-14 in `rmnc` at 20 threads, 9e-14 at 12, 1.6e-13 at 8, identical only when the radial and vacuum teams coincide.

With the process thread count left alone, the macOS CPython 3.14 wheel test runs on three threads, and it segfaulted in libomp. simsopt has no 3.14 wheel and builds from source in the test environment, where `OpenMP_ROOT` made it link Homebrew's libomp, a second copy next to the one bundled in the vmecpp wheel. The test imports simsopt first, so dyld binds the bundled copy's weak `__kmp_suspend_64` to the Homebrew copy, whose runtime is never initialized, and a vmecpp worker that goes to sleep faults there. On main the first `simsopt_compat.Vmec` run (`max_threads=1`) leaves the whole test process on one thread, so no worker exists.

**Evidence** - with the change `omp_get_max_threads()` is the same before and after a run, with and without an explicit `max_threads`, and consecutive runs of one input agree bit for bit. The cth cases take the same wall time either way on the 20-thread laptop used here. On macos-14 with CPython 3.14, a three-thread run with simsopt imported first segfaults on [main](https://github.com/CharlesCNorton/vmecpp/actions/runs/35506082563) and [here](https://github.com/CharlesCNorton/vmecpp/actions/runs/35506082587), and passes with vmecpp imported first or on one thread; with `OpenMP_ROOT` cleared simsoptpp links no OpenMP and the wheel tests [pass](https://github.com/CharlesCNorton/vmecpp/actions/runs/35506082744).

**Tests** - `RunLeavesTheProcessThreadCountUnchanged` in `vmec_test`, `ConsecutiveRunsAreBitIdentical` in `vmec_in_memory_mgrid_test` and `test_consecutive_runs_are_identical` in `tests/test_free_boundary.py`, all failing before the change on a machine with more threads than the coarse run uses.

**Scope** - the team sizes of the solve are unchanged; `omp_set_dynamic(0)` stays so the requested teams are granted in full. The output stage now runs on the run's `max_threads` rather than the last radial count. The `pyproject.toml` line affects only the macOS wheel-test environment.

### DIFF
--- a/tests/test_free_boundary.py
+++ b/tests/test_free_boundary.py
@@ -85,25 +85,6 @@ def test_raise_invalid_nzeta():
         vmecpp.run(vmec_input, response, verbose=False)
 
 
-def test_consecutive_runs_are_identical():
-    """A run does not change the thread budget the next run in the process sees."""
-    makegrid_params = vmecpp.MakegridParameters.from_file(
-        TEST_DATA_DIR / "makegrid_parameters_cth_like.json"
-    )
-    response = vmecpp.MagneticFieldResponseTable.from_coils_file(
-        TEST_DATA_DIR / "coils.cth_like", makegrid_params
-    )
-    vmec_input = vmecpp.VmecInput.from_file(TEST_DATA_DIR / "cth_like_free_bdy.json")
-    vmec_input.ns_array = np.array([7])
-    vmec_input.niter_array = np.array([60])
-    vmec_input.return_outputs_even_if_not_converged = True
-    first = vmecpp.run(vmec_input, response, verbose=False)
-    second = vmecpp.run(vmec_input, response, verbose=False)
-    np.testing.assert_array_equal(second.wout.rmnc, first.wout.rmnc)
-    np.testing.assert_array_equal(second.wout.lmns, first.wout.lmns)
-    assert second.wout.wb == first.wout.wb
-
-
 def test_makegrid_parameters_conversion(makegrid_params):
     # Convert resolution parameters to C++ and back to Python
     cpp_params = makegrid_params._to_cpp_makegrid_parameters()
--- a/tests/test_free_boundary.py
+++ b/tests/test_free_boundary.py
@@ -85,6 +85,25 @@ def test_raise_invalid_nzeta():
         vmecpp.run(vmec_input, response, verbose=False)
 
 
+def test_consecutive_runs_are_identical():
+    """A run does not change the thread budget the next run in the process sees."""
+    makegrid_params = vmecpp.MakegridParameters.from_file(
+        TEST_DATA_DIR / "makegrid_parameters_cth_like.json"
+    )
+    response = vmecpp.MagneticFieldResponseTable.from_coils_file(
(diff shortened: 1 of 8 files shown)


CharlesCNorton:
Closed and reopened to rerun the checks. The one failure was a segmentation fault of the macOS 3.14 wheel test inside the response-table computation, which this change does not touch.

tests/test_free_boundary.py
```diff
@@ -85,6 +85,25 @@ def test_raise_invalid_nzeta():
...
+    vmec_input.ns_array = np.array([7])
+    vmec_input.niter_array = np.array([60])
+    vmec_input.return_outputs_even_if_not_converged = True
+    first = vmecpp.run(vmec_input, response, verbose=False)
+    second = vmecpp.run(vmec_input, response, verbose=False)
+    np.testing.assert_array_equal(second.wout.rmnc, first.wout.rmnc)
+    np.testing.assert_array_equal(second.wout.lmns, first.wout.lmns)
+    assert second.wout.wb == first.wout.wb
+
+
```
jurasic-pf: This test assumes the exact floating point results depend on thread order or at least number of threads, which another PR is intending to remove, I don't think it's robustly testing what we want

jurasic-pf (approved):
Interesting! I didn't think of that edge case

tests/test_free_boundary.py
CharlesCNorton: Was just thinking about that. Looking now.

CharlesCNorton: Replaced both bit-identity checks with a direct assertion that omp_get_max_threads() is unchanged across a run, one fixed-boundary and one free-boundary for the vacuum team's nested region; both fail with the omp_set_num_threads call restored and pass without it.

CharlesCNorton:
fixing ubsan

PR #836, opened 2026-09-08
Size the T_l recurrence passes of the analytic add-back by their growth rate

CharlesCNorton:
**Change** - `SingularIntegrals` evaluates the T_l of Merkel's analytic add-back with a forward pass of the three-term recurrence only while the spurious solutions of that recurrence grow by less than a factor of ten over the kL = mf + nf steps, and otherwise with a backward pass seeded far enough above kL that the seed has decayed below 1e-17 of T_l at kL. The Miller normalization of the backward pass to T_0 is gone, since the recurrence is inhomogeneous. Both passes are one function, `ComputeTl`.

**Cause** - the homogeneous solutions of the recurrence are a complex pair of modulus sqrt(B/A), with B/A = ap/am for T^-, not B/A as the previous comments assumed. The backward pass therefore damped its zero seed by (A/B)^25 over its fixed 50-step tail instead of (A/B)^50, and the forward pass was kept up to a growth of (B/A)^(kL/2) = 1e5. On the `cth_like_free_bdy` boundary at the point of its largest metric cross term (0.89 of sqrt(guu gvv), ap/am = 2.2) that left 2e-9 of the seed in T^-_33 and 5e-12 in T^-_20, the values a 60-digit emulation of the same passes gives at every l, and independent of the working precision: a quad-precision build reproduces them. The alternating binomial sums of the add-back amplify an error in T_33 by up to 1e11 at m + |n| = 33, so the assembled coefficient of mode (17, 16) came out 3.2 times its value.

**Evidence** - at that point, the assembled `bvec` coefficients against direct quadrature of the subtracted tangent-plane kernel (mpol 16, ntor 16, kL 33, with the cross-term sign of #825), as the largest relative error per band m + |n|:

| m + \|n\| | before | after |
| --- | --- | --- |
| 21 | 3.1e-6 | 1.2e-9 |
| 24 | 9.2e-5 | 2.8e-8 |
| 27 | 2.7e-3 | 2.8e-7 |
| 30 | 7.7e-2 | 2.6e-6 |
| 33 | 2.2 | 2.0e-5 |

and `grpmn` at band 33 from 2.8e2 to 6.5e-3. On the exact exterior Neumann problem of an interior dipole line on that boundary (residual relative to the external field, zero being exact):

| mpol, ntor, nzeta | before | after |
| --- | --- | --- |
| 16, 14, 60 | 1.9e-2 | 2.3e-4 |
| 16, 16, 60 | 0.67 | 2.3e-4 |
| 28, 8, 36 | 0.24 | 1.3e-4 |

which is the floor the previous code reached at mpol 16, ntor 12 (2e-4). In the loop, `cth_like_free_bdy` at ftol 1e-9 with the boundary padded to higher cutoffs:

| mpol, ntor | before | after |
| --- | --- | --- |
| 12, 12 | 661 iterations | 661 iterations, same equilibrium |
| 16, 12 | 805 iterations, `rmax_surf` 1.00872, edge iota 0.86370 | 689 iterations, `rmax_surf` 1.00900, edge iota 0.86381 |
| 16, 14 | no convergence, fsqr 4.8 after 4000 iterations | 689 iterations, `rmax_surf` 1.00899, edge iota 0.86380 |
| 16, 16 | no convergence, fsqr 2.2 after 4000 iterations | 689 iterations, `rmax_surf` 1.00899, edge iota 0.86380 |

The three mpol 16 runs now agree with each other to the digits shown, where the previous mpol 16, ntor 12 result differed from them by 0.3 mm in `rmax_surf` and 1e-4 in the edge iota.

**Tests** - `TlpTlmAccuracyTest` gains the cth metric at kL = 33, where T^-_33 was 2e-9 off the quadrature reference against a 1e-11 bound; the three existing cases pass unchanged. `bazel test --config=opt //vmecpp/... //vmecpp_large_cpp_tests/...` and `pytest` (217 passed) pass.

**Scope** - only the T_l passes change. The forward pass is kept exactly where it was accurate, so the bundled cases move at round-off only: every reference comparison of both suites passes at its existing tolerance, and `cth_like_free_bdy` converges in the same 490 iterations to the same `rmax_surf` to nine digits. A backward pass costs at most 17 kL steps per surface point; `cth_like_free_bdy` takes the same wall time as before (0.52 s against 0.52 s at four threads, minimum of three runs).

### DIFF
--- a/src/vmecpp/cpp/vmecpp/free_boundary/singular_integrals/singular_integrals.cc
+++ b/src/vmecpp/cpp/vmecpp/free_boundary/singular_integrals/singular_integrals.cc
@@ -8,9 +8,65 @@
 #include <cmath>
 #include <cstdlib>
 #include <cstring>
+#include <numbers>
 #include <vector>
 
 namespace vmecpp {
+
+namespace {
+
+// Logarithm of the growth of the homogeneous solutions of the T_l recurrence
+// over a forward pass, sqrt(B/A)^kL, above which the recurrence is run
+// backward instead; a forward pass amplifies the rounding of its inputs by at
+// most this factor, ten.
+constexpr double kMaxForwardLogGrowth = std::numbers::ln10;
+
+// The zero seed of a backward pass contaminates T_l by (A/B)^{(top - l)/2} of
+// T_top; the pass starts far enough above kL to bring that below 1e-17 at kL.
+constexpr double kMinSeedLogDecay = 17.0 * std::numbers::ln10;
+
+// T_l = int_{-1}^{1} t^l / sqrt(A t^2 + 2 d t + B) dt for l = 0, ..., kL at
+// tangential grid point kl from the three-term recurrence
+//   (l + 1) A T_{l+1} + (2 l + 1) d T_l + l B T_{l-1} = sqrtc2 - (-1)^l sqrta2
+// with T_0 given. The characteristic roots of the homogeneous recurrence are a
+// complex pair of modulus sqrt(B/A), so a forward pass amplifies the rounding
+// of T_0 and of the right-hand sides by sqrt(B/A)^l and is used while that
+// stays below exp(kMaxForwardLogGrowth); otherwise the recurrence runs
+// backward from a zero seed, whose contamination decays by sqrt(A/B) per step.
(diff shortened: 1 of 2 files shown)


jurasic-pf (approved):
Amazing! I really like these recurrence relations, think they're cool

PR #851, opened 2026-09-12
Let Python attach attributes to VmecModel

jurasic-pf:
## Scope

Bug: `py::class_<VmecModel>` in `src/vmecpp/cpp/vmecpp/vmec/pybind11/pybind_vmec.cc` is
declared without `py::dynamic_attr()`. Python code that wants to attach state to a solved
model instance, for example a cached factorization or some metadata, cannot: `model.some_cache
= ...` raises `AttributeError`. Code that wraps such a setattr in a silent `try/except
AttributeError` hides the failure instead of surfacing it.

## Change

Add `py::dynamic_attr()` to the `VmecModel` class binding. This gives each `VmecModel`
instance a per-instance `__dict__`, the same mechanism pybind11 uses for any Python class:
one extra pointer in the instance layout, plus a dict allocated lazily the first time an
attribute is set. No other binding or method changes.

## Tests and measurements

Added `tests/test_vmec_model_dynamic_attr.py`:
- setting and reading back an arbitrary Python attribute on a `VmecModel` created from
  `examples/data/solovev.json`;
- setting an attribute, then calling `solve()`, and checking the attribute is unchanged and
  `get_state()` still returns a finite, correctly shaped array (dynamic attributes do not
  interfere with the C++-backed state).

Both tests pass locally. Full existing test suite unaffected (no other binding changed).

### DIFF
--- a/src/vmecpp/cpp/vmecpp/vmec/pybind11/pybind_vmec.cc
+++ b/src/vmecpp/cpp/vmecpp/vmec/pybind11/pybind_vmec.cc
@@ -1520,7 +1520,10 @@ PYBIND11_MODULE(_vmecpp, m) {
   // Single-resolution iteration model: exposes the forward model and the
   // time-step / restart primitives so the equilibrium iteration can be driven
   // from Python (see vmecpp._iteration).
-  py::class_<VmecModel>(m, "VmecModel")
+  // py::dynamic_attr() adds a per-instance __dict__ (extra pointer plus a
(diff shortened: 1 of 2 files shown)


jurasic-pf:
Closing: caching state alongside a solved model is better handled on the Python side (a cache keyed by the boundary in the autodiff wrapper, coming in a separate PR) than by attaching dynamic attributes to the C++ object.

PR #871, opened 2026-09-21
Reject a lasym input whose asymmetric arrays are not set

CharlesCNorton:
**Change** - `IsConsistent` returns `kInvalidArgument` for an input that sets `lasym` while `raxis_s`, `zaxis_c`, `rbs` or `zbc` is not set, naming the array: `input variable 'raxis_s' has to be set when 'lasym' is true.` It mirrors the existing rejection of those arrays when `lasym` is false.

**Cause** - the four arrays are `std::optional`, and under `lasym` the size checks reach them as `raxis_s->size()`, `rbs->rows()` and so on without asking whether they hold a value. `FromJson`, `LoadInto` and `SetMpolNtor` all emplace them, so the inputs that reach `IsConsistent` from a file or through `VmecInput` carry them; a `VmecINDATA` filled in by hand in C++ carries them only if the caller emplaced them, and the default constructor leaves them empty.

**Evidence** - on main, `VmecINDATA indata; indata.lasym = true;` is rejected with a size read from the empty optional: `input variable 'raxis_s' has wrong size: should be 1, but it is 140726129447008.` An input parsed from `cth_like_fixed_bdy_asym.json` with any one of the four arrays `reset()` afterwards passes `IsConsistent` with `OK`, the stale size of the destroyed array being the expected one, and goes on to the solver, which dereferences the same optional.

**Tests** - `CheckLasymNeedsTheAsymmetricArrays` in `vmec_indata_test` requires the rejection, by array name, of a default input with `lasym` set and of the parsed asymmetric input with each of the four arrays reset in turn, and requires the default input to pass once `SetMpolNtor` has sized them. On main all five rejections fail as described above.

**Scope** - validation only. An input that carries all four arrays is checked as before.


CharlesCNorton:
looking

